PR #7, opened 2021-02-18
New placement APIs

elgnay:
Signed-off-by: Yang Le <yangle@redhat.com>

### DIFF
--- a/enhancements/sig-architecture/6-placements/README.md
+++ b/enhancements/sig-architecture/6-placements/README.md
@@ -1,0 +1,554 @@
+# New Placement APIs
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+The proposed work would define new placement APIs to select a set of ManagedClusters from multiple ManagedClusterSets.
+
+## Motivation
+
+Since placementrule is used in multiple components (GRC/App/Observability). It is more appropriate to define it in API repo and have a generic component to consume it based on the current placementrule APIs in https://github.com/open-cluster-management/multicloud-operators-placementrule. 
+
+The placement APIs will provide a definition on how to select a set of ManagedClusters in multiple ManagedClusterSets. And placement decisions will be generated according to the spec defined in the placement APIs.
+
+Placement APIs can be used by a higher primitive and controller to decide which ManagedClusters the manifests should be placed.
+
+### User Stories
+
+#### Story 1: A cluster-admin on hub can place a policy or workload to all ManagedClusters in all ManagedClusterSets; The cluster-admin on hub can also place a policy or workload to any selected ManagedClusters in ManagedClusterSets.
+
+#### Story 2: A normal user on hub can place a policy or a workload to a slice of ManagedClusters in one or multiple ManagedClusterSets that the user is authorized to access.
+
+#### Story 3: A user can use predicate to select ManagedClusters
+  - Select ManagedClusters based on a specified flavor - e.g. OpenShift
+  - Select ManagedClusters in aws and gcp
+  - Select ManagedClusters not in vsphere
+  - Select ManagedClusters based on a specified flavor, like OpenShift, with a specific version
+  - Select ManagedClusters labeled as prod (which is set by users).
+
+#### Story 4: A user can define the number of ManagedClusters to be selected
+  - Select 3 ManagedClusters based on a specified flavor
+  - Select 5 ManagedClusters on aws
+
+#### Story 5: The placement API should be scalable to support selecting hundreds of ManagedClusters from thousands of ManagedClusters.
+
+#### Story 6: A user can select ManagedClusters across different regions.
+
+#### Story 7: A controller is able to consume placements to select desired ManagedClusters.
+
+#### Story 8: A user can specify tolerations and select ManagedClusters with particular taints.
+
+#### Story 9: A user can fetch metrics data and figure out how the decision is made for a placement.
+
+#### Story 10: A user can specify a time window for a placement rule.
+
+### Goals
+
+- Define a new placement APIs in API repo to select a set of ManagedClusters in multiple ManagedClusterSets;
+- Have a component created with controllers to consume placements and generate placement decisions;
+- Limit the access to ManagedClusters by leveraging ManagedClusterSets;
+- Replace and deprecate the current placementrule APIs in https://github.com/open-cluster-management/multicloud-operators-placementrule
+- The user stories 8-10 need further investigation and will be covered in separate enhancements when API version moves to beta;
+
+## Proposal
+
+### API Spec
+
+#### Placement API
+`Placement` is a namespace scoped API and represents a selection of ManagedClusters in one or multiple ManagedClusterSets. When a placement is created, a controller consuming the placement should generate one or multiple placement decisions based on the spec defined in placement.
+```go
+// Placement defines a rule to select a set of ManagedClusters from the ManagedClusterSets bound
+// to the placement namespace.
+//
+// Here is how the placement policy combines with other selection methods to determine a matching
+// list of ManagedClusters:
+// 1) Kubernetes clusters are registered with hub as cluster-scoped ManagedClusters;
+// 2) ManagedClusters are organized into cluster-scoped ManagedClusterSets;
+// 3) ManagedClusterSets are bound to workload namespaces;
+// 4) Namespace-scoped Placements specify a slice of ManagedClusterSets which select a working set
+//    of potential ManagedClusters;
+// 5) Then Placements subselect from that working set using label/claim selection.
+//
+// No ManagedCluster will be selected if there is no ManagedClusterSet bound to the placement
+// namespace. User is able to bind a ManagedClusterSet to a namespace by creating a 
+// ManagedClusterSetBinding in the namespace if they have an RBAC rule to CREATE on the virtual
+// subresource of `managedclustersets/bind`.
+//
+// A slice of PlacementDecisions with label placement.open-cluster-management.io={placement name}
+// will be created to represent the ManagedCluster selected by this placement.
+// 
+// If a ManagedCluster is selected and added into the PlacementDecisions, other components may apply
+// workload on it; once it is removed from the PlacementDecisions, the workload applied on this
+// ManagedCluster should be evicted accordingly.
+type Placement struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the attributes of Placement.
+	Spec PlacementSpec `json:"spec"`
+
+	// Status represents the current status of the Placement
+	// +optional
+	Status PlacementStatus `json:"status,omitempty"`
+}
+
+// PlacementSpec defines the attributes of Placement.
+// An empty PlacementSpec selects all ManagedClusters from the bound ManagedClusterSets to placement namespace.
+type PlacementSpec struct {
+	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
+	// +optional
+	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// Affinity represents the scheduling constraints
+	// +optional
+	Affinity *Affinity `json:"affinity,omitempty"`
+}
+
+// ClusterPredicate represents a predicate to select ManagedClusters.
+type ClusterPredicate struct {
+	// ClusterSets represent the ManagedClusterSets from which the ManagedClusters are selected.
+	// If the slice is empty, ManagedClusters will be selected from the ManagedClusterSets bound to the placement
+	// namespace, otherwise ManagedClusters will be selected from the intersection of this slice and the
+	// ManagedClusterSets bound to the placement namespace.
+	// +optional
+	ClusterSets []string `json:"clusterSets,omitempty"`
+
+	// NumberOfClusters represents the number of ManagedClusters to be selected which meet the
+	// requirements in this predicate.
+	// 1) If not specified, it indicates all matched ManagedClusters will be selected;
+	// 2) Otherwise if the nubmer of matched ManagedClusters is larger than NumberOfClusters, a random subset
+	//    with expected number of ManagedClusters will be selected;
+	// 3) If the nubmer of matched ManagedClusters is equal to NumberOfClusters, all matched ManagedClusters
+	//    will be selected;
+	// 4) If the nubmer of matched ManagedClusters is less than NumberOfClusters, all matched ManagedClusters
+	//    will be selected, and the status of condition `PlacementConditionSatisfied` will be set to false;
+	// +optional
+	NumberOfClusters *int32 `json:"numberOfClusters,omitempty"`
+
+	// RequiredDuringSchedulingRequiredDuringExecution represents a scheduling strategy for a
+	// cluster selector. If specified, the cluster selector should be matched at both scheduling
+	// time (when the placement decisions are made for the first time) and execution time (After
+	// one or multiple placement decisions are made). It means,
+	// 1) Any ManagedCluster, which does not match the selector, should not be selected;
+	// 2) If a selected ManagedCluster ceases to match the selector (e.g. due to an update) after the
+	//    placement decisions are made, it will be eventually removed from the placement decisions;
+	// 3) If a ManagedCluster starts to match the selector after the placement decisions are made, it
+	//    will either be selected or at least has chance to be selected (when NumberOfClusters is
+	//    specified);
+	//
+	// Other scheduling strategies not yet implemented includes
+	// RequiredDuringSchedulingIgnoredDuringExecution and PreferredDuringSchedulingIgnoredDuringExecution
+	// +optional
+	RequiredDuringSchedulingRequiredDuringExecution *ClusterSelector `json:"requiredDuringSchedulingRequiredDuringExecution,omitempty"`
+}
+
+// ClusterSelector represents the AND of the containing selectors
+type ClusterSelector struct {
+	// LabelSelector represents a selector of ManagedClusters by label
+	// +optional
+	LabelSelector metav1.LabelSelector `json:"labelSelector,omitempty"`
+
+	// ClaimSelector represents a selector of ManagedClusters by clusterClaims in status
+	// +optional
+	ClaimSelector ClusterClaimSelector `json:"claimSelector,omitempty"`
+}
+
+// A clusterclaim selector is a claim query over a set of ManagedClusters. An empty cluster claim
+// selector matches all objects. A null cluster claim selector matches no objects.
+type ClusterClaimSelector struct {
+	// matchExpressions is a list of cluster claim selector requirements. The requirements are ANDed.
+	// +optional
+	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
+}
+
+// Affinity is a group of affinity scheduling rules.
+type Affinity struct {
+	// ClusterAntiAffinity describes cluster affinity scheduling rules for the placement.
+	// +optional
+	ClusterAntiAffinity *ClusterAntiAffinity `json:"clusterAntiAffinity,omitempty"`
+}
+
+// ClusterAntiAffinity is a group of cluster affinity scheduling rules.
+type ClusterAntiAffinity struct {
+	// RequiredDuringSchedulingRequiredDuringExecution is a scheduling strategies for
+	// ClusterAntiAffinity. The anti-affinity requirements specified by this field should be met
+	// at both scheduling time (when the placement decisions are made for the first time) and
+	// execution time (After one or multiple placement decisions are made). It means,
+	// 1) Any ManagedCluster, which does not meet the requirements, should not be selected;
+	// 2) If a selected ManagedCluster ceases to meet the requirements (e.g. due to an update) after the
+	//    placement decisions are made, it will be eventually removed from the placement decisions;
+	// 3) If a ManagedCluster starts to meet the requirements after the placement decisions are made, it
+	//    will either be selected or at least has chance to be selected;
+	// When there are multiple elements, the lists of ManagedClusters corresponding to each
+	// ClusterAffinityTerm are intersected, i.e. all terms must be satisfied.
+	//
+	// Other scheduling strategies not yet implemented includes
+  // RequiredDuringSchedulingIgnoredDuringExecution and PreferredDuringSchedulingIgnoredDuringExecution
+	// +optional
+	RequiredDuringSchedulingRequiredDuringExecution []ClusterAffinityTerm `json:"requiredDuringSchedulingRequiredDuringExecution,omitempty"`
+}
+
+// ClusterAffinityTerm defines a constraint on a set of ManagedClusters. Any ManagedCluster in this set
+// should be co-located (affinity) or not co-located (anti-affinity) with each other, where co-located
+// is defined as the value of the label/claim with name <topologyKey> on ManagedCluster is the same.
+type ClusterAffinityTerm struct {
+	// TopologyKey is either a label key or a cluster claim name of ManagedClusters
+	// +required
+	TopologyKey string `json:"topologyKey"`
+
+	// TopologyKeyType indicates the type of TopologyKey. It could be Label or Claim.
+	// +required
+	TopologyKeyType string `json:"topologyKeyType"`
+}
+
+const (
+	TopologyKeyTypeLabel string = "Label"
+	TopologyKeyTypeClaim string = "Claim"
+)
+
+type PlacementStatus struct {
+	// NumberOfSelectedClusters represents the number of selected ManagedClusters
+	NumberOfSelectedClusters int32 `json:"numberOfSelectedClusters"`
+
+	// Conditions contains the different condition statuses for this Placement.
+	Conditions []metav1.Condition `json:"conditions"`
+}
+
+const (
+	// PlacementConditionSatisfied means Placement is satisfied.
+	// A placement is satisfied only if all predicates are satisfied. And a predicate
+	// is not satisfied if
+	// 1) NumberOfClusters is specified;
+	// 2) And the number of matched ManagedClusters for this predicate is less than NumberOfClusters;
+	PlacementConditionSatisfied string = "PlacementSatisfied"
+)
+```
+
+#### PlacementDecision API
+`PlacementDecision` represents a slice of clusters selected by a placement. A `Placement` can link to multiple `PlacementDecisions`, and uses a label selector for reference.
+```go
+// PlacementDecision indicates a decision from a placement
+// PlacementDecision should has a label placement.open-cluster-management.io={placement name}
+// to reference a certain placement.
+type PlacementDecision struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Decisions is a slice of decisions according to a placement
+	// The number of decisions should not be larger than 100
+	Decisions []ClusterDecision `json:"decisions"`
+}
+
+// ClusterDecision represents a decision from a placement
+type ClusterDecision struct {
+	// ClusterName is the name of the ManagedCluster
+	// +required
+	ClusterName string `json:"clusterName"`
+
+	// Reason represents the reason why the ManagedCluster is selected.
+	// +required
+	Reason string `json:"reason"`
+}
+```
+
+### Examples
+
+#### 1. Select ManagedClusters with cluster names
+
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: openshift-clusters-with-specific-names
+  namespace: ns1
+spec:
+  predicates:
+    - requiredDuringSchedulingRequiredDuringExecution:
+        claimSelector:
+          matchExpressions:
+            - key: id.k8s.io
+              operator: In
+              values:
+                - cluster1
+                - cluster2
+```
+
+#### 2. Select 3 ManagedClusters from a specific ManagedClusterSet.
+
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: clusters-from-specific-clusterset
+  namespace: ns1
+spec:
+  predicates:
+    - clusterSets:
+      - clusterset1
+      numberOfClusters: 3
+```
+**Note:** No ManagedCluster will be selected if the ManagedClusterSet is not bound to the placement namespace.
+
+#### 3. Select ManagedClusters based on a specific flavor with specific versions
+
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: openshift-clusters-with-specific-version
+  namespace: ns1
+spec:
+  predicates:
+    - requiredDuringSchedulingRequiredDuringExecution:
+        labelSelector:
+          matchLabels:
+            vendor: OpenShift
+        claimSelector:
+          matchExpressions:
+            - key: version.openshift.io
+              operator: In
+              values:
+                - 4.5.8
+                - 4.5.9
+```
+
+#### 4. Select 2 AWS ManagedClusters and 1 GCE ManagedCluster
+
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: clusters-from-aws-and-gce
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+### Non-Goals
+
+- Support tolerations/taints. ManagedCluster has no taint defined in spec yet;
+- Create metrics data when creating/updating placement decisions. It needs further investigation and should be covered in a separate enhancement; 
+
+## Proposal
+
+### User Stories
+
+#### Story 1: A cluster-admin on hub can place a policy or workload to all clusters in all clustersets; The cluster-admin on hub can also place a policy or workload to any selected clusters in clustersets.
```
qiujian16, line 40: I think we should use  ManagedClusterSets to replace clusterset all over the doc.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+
+#### Story 4: A user can define the number of clusters to be selected
+  - Select 3 openshift cluster
+  - Select 5 cluster on aws
+
+#### Story 5: The placement API should be scalable to support selecting hundreds of clusters from thousands of clusters.
+
+#### Story 6: A user can select clusters across different regions.
+
+#### Story 7: A controller is able to consume placements to select desired clusters.
```
qiujian16, line 59: should we also put taint/toleration as a story here also?

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+	TopologyKeyType string `json:"topologyKeyType"`
+}
+
+const (
+	TopologyKeyTypeLabel string = "Label"
+	TopologyKeyTypeClaim string = "Claim"
+)
+
+type PlacementStatus struct {
+	// Conditions contains the different condition statuses for this Placement.
```
qiujian16, line 219: I think we decide to put also number of selected clusters in the status

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+- A clustersetbinding binds only one clusterset to a namespace. It becomes worse if a user, like admin, wants to bind all clustersets to a namespace, because,
+  - The number of clustersets might be very large;
+  - The new added clustersets cannot be bound to the namespace automatically;
+
+### Workflows
+
+#### Workflow 1: Placement Reconciliation
+1. If the placement has old placementdicisions, fetch them all and get a slice of old matched clusters;
+2. Evaluate each cluster predicates (see Workflow 2 below for more details) and get the union of all predicates result;
+3. Enforce the afinity constrains on the union and remove clusters if necessary;
```
qiujian16, line 413: s/afinity /affinity/g

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+9. if `NumberOfClusters` is less than the size of the intersection, select random clusters from the intersection and return;
+10. Otherwise, return clusters in the intersection with some random additional clusters from the remaining of the new matched clusters;
+
+**Note:** The old matched clusters are preferred during the predicate evaluation, which makes the placement decisions as stable as possible.
+
+### Graduation Criteria
+
+1. The new placement APIs starts with version v1alpha1;
+2. The version name will become v1beta1 once it gets stable and is able to cover all the capabilities of the current placementrule APIs; 
+3. Components depends on the current placementrule APIs can migrated to the new placement APIs (v1beta1) then. And a conversion webhook will be provided to migrated existing placement objects to the new placement APIs (v1beta1);
```
qiujian16, line 435: s/migrated/migrate/g

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+4. The current placementrule APIs becomes deprecated and still exists in the next 3 releases.
```
qiujian16, line 436: i think only when we move this api to higher then v1beta1, we can start the deprecation process.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,466 @@
...
+              vendor: OpenShift
+    ```
+2. A validating webhook rejects the create/update/patch requests of placements if the user has no permission on the referenced clustersets;
+3. PlacementController generates placementdecisions in the same namespace according to the spec of the placement.
+
+#### Pro
+- No need to bind clustersets to the placement namespace;
+
+#### Cons
+- The webhook can only handle create/update/patch requests of placements. If we want to use clustersets to limit the access scope of other resource, new webhook is required.
```
qiujian16, line 466: it will not be easy to get the clustersets for this working ns either

enhancements/sig-architecture/6-placements/README.md
elgnay: Since tolerations/taints is under non-goals section, this proposal will not cover it. So we may not be able to add a story for it.

enhancements/sig-architecture/6-placements/README.md
elgnay: Done.

enhancements/sig-architecture/6-placements/README.md
elgnay: Added.

enhancements/sig-architecture/6-placements/README.md
elgnay: Updated.

enhancements/sig-architecture/6-placements/README.md
elgnay: Updated.

enhancements/sig-architecture/6-placements/README.md
elgnay: Updated.

enhancements/sig-architecture/6-placements/README.md
elgnay: Added.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+5. Enforce the affinity constrains on clusters if exist;
+6. If `NumberOfClusters` is not specified, return all matched clusters;
+7. Otherwise, get the intersection of old matched clusters and new matched clusters;
+8. If `NumberOfClusters` is equal to the size of the intersection, return clusters in the intersection;
+9. if `NumberOfClusters` is less than the size of the intersection, select random clusters from the intersection and return;
+10. Otherwise, return clusters in the intersection with some random additional clusters from the remaining of the new matched clusters;
+
+**Note:** The old matched clusters are preferred during the predicate evaluation, which makes the placement decisions as stable as possible.
+
+### Graduation Criteria
```
qiujian16, line 434: This would need explicitly indicate criteria of:

Alpha -> Beta
Beta  -> GA

qiujian16 (commented):
I think Test Plan, Upgrade Strategy and Version Skew section is needed.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+	// +optional
+	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// Affinity represents the scheduling constraints
+	// +optional
+	Affinity *Affinity `json:"affinity,omitempty"`
+}
+
+// ClusterPredicate represents a predicate to select clusters.
+type ClusterPredicate struct {
```
jnpacker, line 113: I am interested in adding a time window to the predicate.
I'm not ready to say we DON'T show the placement target, but at least in the status, mark the predicate list or grouping as active/inactive(blocked)

I could also be convinced this is a top level object, and just applies to the entire placement rule.

https://github.com/open-cluster-management/multicloud-operators-subscription/blob/84dfe430767e453872ccdb19218b3ccb1ab8e913/deploy/crds/apps.open-cluster-management.io_subscriptions.yaml#L321

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+The proposed work would define new placement APIs to select a set of managed clusters from multiple ManagedClusterSets.
+
+## Motivation
+
+Since placementrule is used in multiple components (GRC/App/Observability). It is more appropriate to  define it in API repo and have a generic component to consume it based on the current placementrule APIs in https://github.com/open-cluster-management/multicloud-operators-placementrule
```
deads2k, line 17: nit: breaking lines on sentences makes it a little easier to comment.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+`Placement` is a namespace scoped API and represents a selection of clusters in one or multiple ManagedClusterSets. When a placement is created, a controller consuming the placement should generate one or multiple placement decisions based on the spec defined in placement.
+```go
+// Placement defines a rule to select a set of ManagedClusters from the ManagedClusterSets bound
+// to the placement namespace.
+//
+// Here is how the placement policy combines with other selection methods to determine a matching
+// list of clusters:
+// 1) Clusters are registered with hub as cluster-scoped ManagedClusters;
+// 2) ManagedClusters are organized into cluster-scoped ManagedClusterSets;
+// 3) ManagedClusterSets are bound to workload namespaces;
```
deads2k, line 73: how is this bit done?  Link to existing API?

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+// 4) Namespace-scoped Placements specify a slice of ManagedClusterSets which select a working set
+//    of potential ManagedClusters;
+// 5) Then Placements subselect from that working set using name/label/claim selection.
+//
+// No ManagedClusters will be selected if there is no ManagedClusterSet bound to the placement
+// namespace. User is able to bind a ManagedClusterSet to a namespace in two ways:
+// 1) Create a ManagedClusterSetBinding in the namespace if they have an RBAC rule to CREATE
+//    on the virtual subresource of `managedclustersets/bind`;
+// 2) Create a service account with name 'placement', and create a ClusterRoleBinding to grant
```
deads2k, line 82: On generic kubernetes, how will a user be able to tell which managedclustersets a namespace has access to when it has been bound using this permission's based method.

deads2k, line 82: Since permissions could be created by edits to roles, clusterroles, clusterolebindings, rolebindings, external webhooks, how will a controller know that it has to refresh the placementdecisions?

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+	ClusterName string `json:"clusterName,omitempty"`
+
+	// NumberOfClusters represents the number of clusters to be selected which meet the
+	// requirements in this predicate.
+	// Nil indicates that NumberOfClusters is not set and there is no bound number.
+	// +optional
+	NumberOfClusters *int32 `json:"numberOfClusters,omitempty"`
+
+	// RequiredDuringSchedulingRequiredDuringExecution represents a scheduling strategy for a
+	// cluster selector. If specified, the cluster selector should be mached at both scheduling
```
deads2k, line 133: typoe: matched

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+
+// A clusterclaim selector is a claim query over a set of managed cluster. The result of matchClaims and
+// matchExpressions are ANDed. An empty cluster claim selector matches all objects. A null
+// cluster claim selector matches no objects.
+type ClusterClaimSelector struct {
+	// matchClaims is a map of {key,value} pairs. A single {key,value} in the matchClaims
+	// map is equivalent to an element of matchExpressions, whose key field is "key", the
+	// operator is "In", and the values array contains only "value". The requirements are ANDed.
+	// +optional
+	MatchClaims map[string]string `json:"matchClaims,omitempty"`
```
deads2k, line 168: You often see this in APIs when selecting nodes, but why have both options in `ClusterClaimSelector`?

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,470 @@
...
+	// +optional
+	NumberOfClusters *int32 `json:"numberOfClusters,omitempty"`
+
+	// RequiredDuringSchedulingRequiredDuringExecution represents a scheduling strategy for a
+	// cluster selector. If specified, the cluster selector should be mached at both scheduling
+	// time (when the placement decisions are made for the first time) and execution time (After
+	// one or multiple placement decisions are made). It means,
+	// 1) Any cluster, which does not match the selector, should not be selected;
+	// 2) If a selected cluster ceases to match the selector (e.g. due to an update) after the
+	//    placement decisions are made, it will be eventually removed from the placement decisions;
```
deads2k, line 138: Any guidance about what should happen when a managecluster is removed from the placementdecisions?  Should whatever was placed there be actively removed?

enhancements/sig-architecture/6-placements/README.md
elgnay: The original thought is to create a controller which generates ManagedClusterSetBindings automatically according to the RBAC rules of service account `placement` in any namespace.

elgnay: The ManagedClusterSetBindings will be created automatically according to RBAC rules of service account `placement`. And any changes on ManagedClusterSetBindings in the placement namespace will trigger the refresh of placementdecisions. I'll remove way 2 from the api doc because it is a separate logic and has nothing to do with the placement API itself. And we may not need it at all.

enhancements/sig-architecture/6-placements/README.md
qiujian16: I think it is an interesting use case. But I would limit the scope of this proposal as-is for now, and treated this case as a following proposal. WDYT?

@elgnay do you think we could put it as a use case in this doc and set it as API beta graduation criteria.

enhancements/sig-architecture/6-placements/README.md
elgnay: Anyway, I added support of tolerations/taints as a user story.

enhancements/sig-architecture/6-placements/README.md
elgnay: Updated as suggested.

enhancements/sig-architecture/6-placements/README.md
elgnay: I've added a user story for time window support and set it as API beta graduation criteria. We may need an additional enhancement to cover it.

enhancements/sig-architecture/6-placements/README.md
elgnay: Updated.

enhancements/sig-architecture/6-placements/README.md
elgnay: Updated.

enhancements/sig-architecture/6-placements/README.md
elgnay: MatchClaims is removed.

enhancements/sig-architecture/6-placements/README.md
elgnay: Added a guidance in the comments of `Placement` about what should happen when a managecluster is removed from the placementdecisions.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,517 @@
...
+	// +optional
+	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// Affinity represents the scheduling constraints
+	// +optional
+	Affinity *Affinity `json:"affinity,omitempty"`
+}
+
+// ClusterPredicate represents a predicate to select clusters.
+type ClusterPredicate struct {
```
yue9944882, line 120: if the type is an union, better having a discriminator field to fit k8s union api structure https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1027-api-unions

yue9944882 (commented):
https://github.com/open-cluster-management/governance-policy-propagator/blob/54f7d55262d6d7958a16da068981122715dbb047/pkg/apis/policies/v1/placementbinding_types.go
 
 the placement binding api looks pretty generic, and i wonder if it's a good idea to promote the type as well

enhancements/sig-architecture/6-placements/README.md
elgnay: Thanks for the suggestion. I removed `ClusterName` field from `ClusterPredicate` because user can get cluster name from both cluster labels and cluster claims. Now this struct is no longer a union type.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+	// namespace, otherwise clusters will be selected from the intersection of this slice and the
+	// ManagedClusterSets bound to the placement namespace.
+	// +optional
+	ClusterSets []string `json:"clusterSets,omitempty"`
+
+	// NumberOfClusters represents the number of clusters to be selected which meet the
+	// requirements in this predicate.
+	// Nil indicates that NumberOfClusters is not set and there is no bound number.
+	// +optional
+	NumberOfClusters *int32 `json:"numberOfClusters,omitempty"`
```
yue9944882, line 132: precisely, this should be `maxNumberOfClusters` considering a case that the matching clusters doesnt reach the number

qiujian16: the design currently is that the number must be met, otherwise decision will not be created. maxNumberOfClusters would be a different semantics, since we would also  need minNumberOfClusters together.

enhancements/sig-architecture/6-placements/README.md
elgnay: You may have already noticed that a placement will select no cluster if there is no clusterset bound to the placement namespace. So a cluster belongs to no clusterset will never been selected by any placement. This is one of the limitations of the current design. We'll figure out a way to fix it in beta phase.

> ### Beta
>- The new APIs is able to cover all the capabilities of the current placementrule APIs;
>- Conversion webhook is provided to migrate existing placement objects to the new placement APIs;-
>- At lease one component uses the new APIs;
>- **Support selecting clusters belong to no ManagedClusterSet;**
>- Support tolerations/taints (user story 8);
>- Create metrics data when creating/updating placement decisions (user story 9);
>- Support time window (user story 10);

In addition, without ClusterName field, it is still possible to select clusters with names from bound clustersets by using either label (see example below) or claim (see example 1 in proposal).
```
kind: Placement
metadata:
  name: clusters-with-specific-names
  namespace: ns1
spec:
  predicates:
    - requiredDuringSchedulingRequiredDuringExecution:
        labelSelector:
          matchExpressions:
            - key: name
              operator: In
              values:
                - cluster1
                - cluster2
```

qiujian16:
> https://github.com/open-cluster-management/governance-policy-propagator/blob/54f7d55262d6d7958a16da068981122715dbb047/pkg/apis/policies/v1/placementbinding_types.go
> 
> the placement binding api looks pretty generic, and i wonder if it's a good idea to promote the type as well

@yue9944882 we haven't thought about this yet. I think the main purpose for binding is to remove the reference of placement in the workload API? The reason we haven't introduce this is because it is not clear how a generic controller could control the binding api given that binding could ref to different primitives. It seems to me the the primitives controller itself has to understand the binding API.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+	Spec PlacementSpec `json:"spec"`
+
+	// Status represents the current status of the Placement
+	// +optional
+	Status PlacementStatus `json:"status,omitempty"`
+}
+
+// PlacementSpec defines the attributes of Placement.
+// An empty PlacementSpec selects all clusters from the bound ManagedClusterSets to placement namespace.
+type PlacementSpec struct {
```
deads2k, line 109: If I want to have my Placement select only "healthy" or "reporting" managedclusters, is there a way to do that?

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+
+Limitations:
+- A ManagedClusterSetBinding binds only one ManagedClusterSet to a namespace. It becomes worse if a user, like admin, wants to bind all ManagedClusterSets to a namespace, because,
+  - The number of ManagedClusterSets might be very large;
+  - The new added ManagedClusterSets cannot be bound to the namespace automatically;
+
+### Workflows
+
+#### Workflow 1: Placement Reconciliation
+1. If the placement has old placementdicisions, fetch them all and get a slice of old matched clusters;
```
deads2k, line 420: typo dicisions -> decisions

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+)
+```
+
+#### PlacementDecision API
+`PlacementDecision` represents a slice of clusters selected by a placement. A `Placement` can link to multiple `PlacementDecisions`, and uses a label selector for reference.
+```go
+// PlacementDecision indicates a decision from a placement
+// PlacementDecision should has a label placement.open-cluster-management.io={placement name}
+// to reference a certain placement.
+type PlacementDecision struct {
```
deads2k, line 238: I think you'll want an example of how a controller could handle a cluster being removed from a PlacementDecision that happens while the consuming controller is off.  I think the API is fine, but an explanation and/or sample code for handling the inverted match will help avoid some mistakes.

enhancements/sig-architecture/6-placements/README.md
deads2k: > If I want to have my Placement select only "healthy" or "reporting" managedclusters, is there a way to do that?

If we don't know how to do this now, making it a requirement to consider for beta is fine.

deads2k:
I like this design.  A nit and some things to solve for a beta, but I think exploring it as an alpha to find other limitations at this point makes sense.

/approve

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+1. The API group of the new placement APIs could be `cluster.open-cluster-management.io` or `scheduling.open-cluster-management.io`. Which one is better?
+
+### Test Plan
+
+- Unit tests will cover the functionality of the controllers.
+- Unit tests will cover the new APIs.
+- Integration tests will cover all user stories.
+- And e2e tests will cover the following cases.
+  - Create/update/delete a placement;
+  - ManagedCluster labels/claims changes;
```
pmorie, line 451: One thing I realized while reviewing this is that it's not clear how ClusterClaims on a ManagedCluster are reflected on the hub if you just look at the ClusterClaim type doc. We should probably add a little information to https://github.com/open-cluster-management/api/blob/master/cluster/v1alpha1/types.go#L118

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+### Graduation Criteria
+
+#### Alpha
+1. The new APIs is reviewed and accepted;
+2. Implementation is completed to support the functionalities;
+3. Develop test cases to demonstrate that the above user stories work correctly;
+
+#### Beta
+1. The new APIs is able to cover all the capabilities of the current placementrule APIs;
+2. Conversion webhook is provided to migrate existing placement objects to the new placement APIs;
```
pmorie, line 464: cc @deads2k is this more appropriate for an offline migration? I don't see this as the job of a conversion webhook

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+3. At lease one component uses the new APIs;
+4. Support selecting clusters belong to no ManagedClusterSet;
+5. Support tolerations/taints (user story 8);
```
pmorie, line 467: Is this a goal or a non-goal? In the beginning of this document it's nak'd by being in the list of non-goals

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+2. Change on `ClusterManager` CRD is required to include information for `placement-manager`;
+3. Change on `cluster-manager` is required to
+    - Install the new placement CRD on hub cluster;
+    - Deploy `placement-manager`;
+
+#### API Migration
+For those components who use the current placementrule API and want to migrate to the new placement APIs, the strategy below should be followed:
+1. The new placement APIs will co-exists with the current placementrule API in at least 3 releases;
+2. The objects of the current placementrule API will be converted into objects of the new placement APIs with the same name by conversion webhook;
+3. Mannual operation is required to make the converted new placement objects work by binding necessary clustersets to the placement namespace.
```
pmorie, line 490: This bullet is also sounding to me like conversion into the new APIs will need to be an offline process

pmorie (commented):
Looks pretty good, I mostly left comments about what could be clearer; the 'conversion webhook' stuff needs an ack from @deads2k in my mind as operable.

enhancements/sig-architecture/6-placements/README.md
deads2k: > cc @deads2k is this more appropriate for an offline migration? I don't see this as the job of a conversion webhook

I think this may be slightly difficult to parse, but a valid statement.  When going to beta, the existing placement objects will be v1alpha1.  Assuming the API doesn't change dramatically, it should be possible to build a conversion webhook to cleanly migrate from v1alpha1 to v1beta1.

I think they called out the requirement here because kube's general promotion rules do not *require* this style of conversion. The default rules allow for an alpha stage that must be removed before installing a beta.  Though we don't usually do that, it has (rarely) happened that features cannot be auto-promoted.

enhancements/sig-architecture/6-placements/README.md
elgnay: We plan to support selecting clusters with toleration/taints when api version moves to beta.  Unhealthy clusters will be marked with some taints which prevent the clusters from being selected by a placement without any toleration specified.

enhancements/sig-architecture/6-placements/README.md
elgnay: Corrected.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+#### Story 8: A user can specify tolerations and select clusters with particular taints.
+
+#### Story 9: A user can fetch metrics data and figure out how the decision is made for a placement.
+
+#### Story 10: A user can specify a time window for a placement rule.
+
+### Goals
+
+- Define a new placement APIs in API repo to select a set of managed clusters in multiple ManagedClusterSets;
+- Have a generic component to consume placements and generate placement decisions;
```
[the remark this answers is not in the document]
elgnay: Updated the sentence to make it more descriptive.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+	// +optional
+	Affinity *Affinity `json:"affinity,omitempty"`
+}
+
+// ClusterPredicate represents a predicate to select clusters.
+type ClusterPredicate struct {
+	// ClusterSets represent the ManagedClusterSets from which the clusters are selected.
+	// If the slice is empty, clusters will be selected from the ManagedClusterSets bound to the placement
+	// namespace, otherwise clusters will be selected from the intersection of this slice and the
+	// ManagedClusterSets bound to the placement namespace.
```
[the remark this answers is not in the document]
elgnay: Yes, it's true.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+	//
+	// Other scheduling strategies not yet implemented includes
+	// RequiredDuringSchedulingIgnoredDuringExecution and PreferredDuringSchedulingIgnoredDuringExecution
+	// +optional
+	RequiredDuringSchedulingRequiredDuringExecution *ClusterSelector `json:"requiredDuringSchedulingRequiredDuringExecution,omitempty"`
+}
+
+// ClusterSelector represents the AND of the containing selectors
+type ClusterSelector struct {
+	// LabelSelector represents a selector of clusters by label
```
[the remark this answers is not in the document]
elgnay: > I might suggest a more brief form of this name `requiredDuringSchedulingAndExecution`

Regarding `RequiredDuringSchedulingRequiredDuringExecution`, we borrowed this field name from NodeAffinity. They have the similar meanings. So I think it might be better to keep this filed name, for users may have already been familiar with it in NodeAffinity.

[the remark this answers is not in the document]
elgnay: > Label on which resource exactly? I think it will pay to be specific here.

Updated the comments with the resource type name.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,523 @@
...
+	// When there are multiple elements, the lists of clusters corresponding to each
+	// ClusterAffinityTerm are intersected, i.e. all terms must be satisfied.
+	//
+	// Other scheduling strategies not yet implemented includes
+  // RequiredDuringSchedulingIgnoredDuringExecution and PreferredDuringSchedulingIgnoredDuringExecution
+	// +optional
+	RequiredDuringSchedulingRequiredDuringExecution []ClusterAffinityTerm `json:"requiredDuringSchedulingRequiredDuringExecution,omitempty"`
+}
+
+// The selected clusters could not co-located with each other,
```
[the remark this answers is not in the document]
elgnay: The doc for type `ClusterAffinityTerm` is updated.

enhancements/sig-architecture/6-placements/README.md
elgnay: The following will be added to the doc of ClusterClaim when the placement api is added to api repo.

```go
// ClusterClaims created on a managed cluster will be collected and saved into
// the status of the corresponding ManagedCluster on hub.
type ClusterClaim struct {
```

enhancements/sig-architecture/6-placements/README.md
elgnay: > 2. Conversion webhook is provided to migrate existing placement objects to the new placement APIs;

I removed this item because
- Conversion webhook does not work across API groups;
- For the new placement APIs, upgrading from alpha to beta may not need a webhook to do the conversion;

I agree with @pmorie that an offline migration is needed between the old placementrule APIs and the new placement APIs

enhancements/sig-architecture/6-placements/README.md
elgnay: I moved them to Goals section. And we will consider to support them in beta.

enhancements/sig-architecture/6-placements/README.md
elgnay: An example is added as suggested.

enhancements/sig-architecture/6-placements/README.md
elgnay: Yes, you are right. It should be an offline process. And a conversion tool will be provided instead.

enhancements/sig-architecture/6-placements/README.md
elgnay: > This field seems like it expresses 'min' semantics. If this field is valued at 5, and one of the clusters is deprovisioned and removed from the set, what happens? I would expect that a new cluster will be selected automatically.

Yes, a new cluster will be selected if possible.

> What happens if this field is valued at 5 but there are only 3 clusters available?

The original thought is that placementdecision will still be created with 3 clusters, and the condition `PlacementSatisfied` in the status of placement will be set to false, while @qiujian16 suggested that no placementdecision will be created in this situation.

enhancements/sig-architecture/6-placements/README.md
```diff
@@ -0,0 +1,550 @@
...
+
+#### Story 1: A cluster-admin on hub can place a policy or workload to all clusters in all ManagedClusterSets; The cluster-admin on hub can also place a policy or workload to any selected clusters in ManagedClusterSets.
+
+#### Story 2: A normal user on hub can place a policy or a workload to a slice of clusters in one or multiple ManagedClusterSets that the user is authorized to access.
+
+#### Story 3: A user can use predicate to select clusters
+  - Select openshift clusters only
+  - Select clusters in aws and gcp
+  - Select cluster not in vsphere
+  - Select openshift clusters with a specific version
```
pdettori, line 33: Should the spec be broader and target Kubernetes clusters in general (e.g. "Select Kubernetes clusters with a specific version" and "Select Kubernetes clusters based on a specified flavor - e.g. OpenShift"

elgnay: Updated as suggested.

enhancements/sig-architecture/6-placements/README.md
elgnay: I updated the doc of this field to make it more clear.

PR #11, opened 2021-03-18
Support orphan delete in manifestwork

qiujian16:


enhancements/sig-architecture/10-deletepropagationstrategy/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+We will introduce a `DeletePropagations` property as following:
+
+```go
+// ManifestWorkSpec represents a desired configuration of manifests to be deployed on the managed cluster.
+type ManifestWorkSpec struct {
+	// Workload represents the manifest workload to be deployed on a managed cluster.
+	Workload ManifestsTemplate `json:"workload,omitempty"`
+
+	// DeletePropagationOrphan represents resources to be orphaned when the manifestwork is deleted.
```
deads2k, line 59: You want to selectively orphan instead of simply orphaning everything related to it?  Given that a ManifestWork is a unit of related sources, can you elaborate under what conditions you would want to be selective?  As I recall it was primarily geared for cases were a particular resource was shared in some way.

It might make sense to have specifying an orphan policy cover orphaning everything by default and selective orphaning as a secondary option.

enhancements/sig-architecture/10-deletepropagationstrategy/README.md
```diff
@@ -0,0 +1,205 @@
...
+	// Foreground deletion strategy is applied to all the resource in this manifestwork when it is nil.
+  // Orphan deletion strategy is applied to all the resource in this manifestwork when it is empty.
+	// +optional
+	DeletePropagationOrphan *DeletePropagationStrategy `json:"deletePropagationsOrphan,omitempty"`
+}
+
+type DeletePropagationStrategy struct {
```
deads2k, line 66: rules like this work best you can specify a slice of them.   Is this API close to what you want to describe?

```go
type DeletePropagationStrategy struct{
    // type can be Orphan or SelectivelyOrphan
    // SelectivelyOrphan should be rarely used.  It is provided for cases where a particular resource is transfering
    // ownership from one ManifestWork to another or another management unit.
    // Setting this value will allow a flow like
    // 1. create manifestwork/2 to manage foo
    // 2. update manifestwork/1 to selectively orphan foo
    // 3. remove foo from manifestwork/1 without impacting continuity because manifestwork/2 adopts it.
    Type string

    SelectivelyOrphan *SelectivelyOrphan
}

type SelectivelyOrphan struct{
    // each orphaningrule identifies a single resource included in this manifestwork
    OrphaningRules []OrphaningRule
}

type OrphaningRule struct{
  Group string
  Resource string
  ResourceNamespace string
  ResourceName string
}
```

deads2k:
Some minor suggestions on the API shape and some clarifying comments requested in the API doc itself, but otherwise I like it.

enhancements/sig-architecture/10-deletepropagationstrategy/README.md
qiujian16: That looks great, updated.

deads2k:
The use-case makes sense to me and I think the API can fulfill the use-case.  It would be good to get another set of eyes who is less deep in apimachinery to see if the APi and use-case are clearly explained for them as well.

/approve

enhancements/sig-architecture/10-deletepropagationstrategy/README.md
```diff
@@ -0,0 +1,240 @@
...
+spec:
+  workload:
+    manifests:
+    - apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: cm1
+        namespace: default
+  deletePropagationsStrategy:
+    type: Orphan
```
qiujian16, line 135: Other options
1. Deference?
2. ShallowDelete?
3. IgnoreOnDelete?

mdelder (commented):
I think the idea here is important to add support for.

I would like to keep the language consistent with the cascading delete examples as in https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#setting-the-cascading-deletion-policy

So something like:
```
deleteOptions:
  propagationPolicy: Foreground|Background|Abandon
```

qiujian16:
> I think the idea here is important to add support for.
> 
> I would like to keep the language consistent with the cascading delete examples as in https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#setting-the-cascading-deletion-policy
> 
> So something like:
> 
> ```
> deleteOptions:
>   propagationPolicy: Foreground|Background|Abandon
> ```

Updated

PR #16, opened 2021-08-09
schedule placement based on resource usage/capacity

elgnay:
The related feature request: https://github.com/open-cluster-management-io/community/issues/52
Signed-off-by: Yang Le <yangle@redhat.com>


dongwangdw (commented):
Hi @elgnay ,

I have some comments. Most of them are rewording. Please consider them.
Hope it can be  helpful to make this doc more readable.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+Add a field `ClusterResourcePreference` into the Placement spec to allow users to set the cluster resource preference.
+```go
+type PlacementSpec struct {
+	// ... ...
+
+	// ClusterResourcePreference represents the cluster resource preference configuration for
+	// the placement.
+	// This field is ignored when NumberOfClusters in spec is not specified.
+	// +optional
+	ClusterResourcePreference *ClusterResourcePreference `json:"clusterResourcePreference,omitempty"`
```
qiujian16, line 61: What is the difference between  nil and empty?

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+- If the placement has `ClusterResourcePreference` specified in the spec and its `Type` is `MostAllocatable`, the score of a managed cluster is the sum of the score for each cluster resource.  
+`score = sum((resource_x_allocatable - min(resource_x_allocatable)) / (max(resource_x_allocatable) - min(resource_x_allocatable))`  
+while
+  - `resource_x_allocatable` is the allocatable of resource `x` on the managed cluster, 
+  - `min(resource_x_allocatable)` is the minimum allocatable of resource `x` from all managed cluster.  
+  - `max(resource_x_allocatable)` is the maxmum allocatable of resource `x` from all managed cluster.  
+- Otherwise, the score for each managed cluster is 0.  
+
+Before returning the scores to the scheduler, the data should be normalized and ensure the value falls in the range between 0 and 100.  
+`normalized = (score - min(score)) * 100 / (max(score) - min(score))`
```
qiujian16, line 191: I think the test cases/version upgrade etc is required. You can set it to N/A if it is not required.

I addition I would like to see some examples, and a discussion on how it works with steady/balance plugin today.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+    cpu: '18'
+    memory: 69938Mi
+    socket_worker: '6'
+```
+In order to support other resources, like GPU, the `allocatable` and `capacity` should be included in the status of managed cluster either.
+
+### Plugin `mostallocatabletocapacityratio`
+It is a pritoritizer and scores feasible managed clusters with the process below.
+- If the placement has `ClusterResourcePreference` specified in the spec and its `Type` is `MostAllocatableToCapacityRatio`, the score of a managed cluster is the sum of the score for each resource.  
+`score = sum(resource_x_allocatable / resource_x_capacity))`
```
qiujian16, line 170: it is worth to  mention that  we treat each  resource dimension with equal weight and the reasoning.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
elgnay: Here is the difference:
- If nil, it means the placement has no `ClusterResourcePreference` at all;
- if empty, it means the placement has an empty `ClusterResourcePreference`, whose type is `MostAllocatableToCapacityRatio`. Consider `ClusterResourcePreference.ClusterResource` must have at lease one item, an empty `ClusterResourcePreference` is invalid.

```go
 	// +kubebuilder:validation:MinItems:=1
        // +required
 	ClusterResources []ClusterResource `json:"clusterResources"`
```

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+  - User wants to deploy an application on a cluster whose CPU usage is less than others;
+  - User wants to deploy an application on 2 clusters whose allocatable to capacity ratio of both CPU and memory is larger than others;
+
+#### Story 2: User is able to specify resources in placement spec to favor clusters with larger amount of resources allocatable.
+  - User wants to deploy an application on a cluster with largest memory allocatable;
+
+#### Story 3: User is able to create a placement to select clusters based on resource usage, and then keep placement decisions updated according to the changes on cluster resource usage.
+  - A CPU-intensive application would like to be deployed on the cluster with least CPU utilization rate.
+
+#### Story 4: User is able to create a placement to select clusters based on resource usage, and then ignore any resource usage change afterwards to keep the placement decisions pinning.
```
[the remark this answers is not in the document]
elgnay: > Hi @elgnay ,
> 
> I have some comments. Most of them are rewording. Please consider them.
> Hope it can be helpful to make this doc more readable.

@dongwangdw Thanks for the comments. I'll update the proposal accordingly.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+    memory: 63194Mi
+  capacity:
+    core_worker: '18'
+    cpu: '18'
+    memory: 69938Mi
+    socket_worker: '6'
+```
+In order to support other resources, like GPU, the `allocatable` and `capacity` should be included in the status of managed cluster either.
+
+### Plugin `mostallocatabletocapacityratio`
```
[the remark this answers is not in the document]
elgnay: Yes, you are right. That's the reason why we need another type `MostAllocatable` besides `MostAllocatableToCapacityRatio`. 

Actually, none of them works perfectly in all scenarios. For example, suppose there are two clusters:
- cluster1 with 15/20 cpus allocatable;
- cluster2 with 20/200 cpus allocatable;

With type `MostAllocatableToCapacityRatio`, cluster1 will have a higher score than cluster2,  while with type `MostAllocatable`, cluster2 will have a higher score. Which one makes more scenes? it depends on user's requirement.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/website/)
+
+## Summary
+
+The proposed work would enhance the Placement APIs to support scheduling based on the resource usage/capacity of managed clusters.
+
+## Motivation
+
+When scheduling placement, select managed clusters based on cluster resource usage/capacity to prevents placement of workloads on clusters that have insufficient resources to actually run the workloads.
```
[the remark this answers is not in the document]
elgnay: Updated on suggestion.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+Link to the feature request: https://github.com/open-cluster-management-io/community/issues/52
+
+### User Stories
+
+#### Story 1: User is able to specify resources in placement spec to favor clusters with larger allocatable to capacity ratio of the resources.
```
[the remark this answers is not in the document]
elgnay: Updated on suggestion.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+  - User wants to deploy an application on a cluster whose CPU usage is less than others;
+  - User wants to deploy an application on 2 clusters whose allocatable to capacity ratio of both CPU and memory is larger than others;
+
+#### Story 2: User is able to specify resources in placement spec to favor clusters with larger amount of resources allocatable.
```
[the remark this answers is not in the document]
elgnay: Updated on suggestion.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,191 @@
...
+  - User wants to deploy an application on a cluster with largest memory allocatable;
+
+#### Story 3: User is able to create a placement to select clusters based on resource usage, and then keep placement decisions updated according to the changes on cluster resource usage.
```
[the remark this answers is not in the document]
elgnay: Updated on suggestion.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
elgnay: Added.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
elgnay: Updated.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,221 @@
...
+#### Story 4: User is able to select clusters based on some resource(s) usage once, without considering the usage changes afterwards, keeping the decisions pinned.
+  - Some critical service should run on a steady set of clusters, even if the resource usage of those clusters changes;
+
+#### Story 5: Support scheduling based on usage/allocatable of resources other than CPU and Memory. For example,
+  - AI applications are expected to run on clusters with least gpu usage;
+  - An application can be scheduled to cluster with largest available network bandwidth;
+
+### Goals
+
+- Sotry 1 and 2, allow user to select managed clusters based on resource usage/capability with the Placement APIs;
```
deads2k, line 42: typo story

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Non-Goals
+- Balance workload across the fleet based on cluster resource usage;
+- Story 3 and 4 are related to churning policy of placement, and will be covered in a separated enhancement proposal;
```
deads2k, line 46: is this different then that the steady weighting?

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
deads2k: Given steady and balance weighting, it seems like we may want a more generic prioritizing configuration.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+The proposed work would enhance the Placement APIs to support scheduling based on the resource usage/capacity of managed clusters.
+
+## Motivation
+
+During placement scheduling, managed clusters will be selected based on resource(s) (like memory and cpu etc.) usage/capacity to avoid scheduling workloads on clusters with insufficient resources.
+
+Link to the feature request: https://github.com/open-cluster-management-io/community/issues/52
+
+### User Stories
```
deads2k, line 21: is there a way to express: "I must have at least this much space"?  If not, why was that excluded?

deads2k: how does a user handle it when their clusters are autoscaling?  What do they expect to happen?  This could be presented as there not being capacity available I think.  Or perhaps that capacity is tight, but will expand.

elgnay: > is there a way to express: "I must have at least this much space"? If not, why was that excluded?

No, because we can not support it. Suppose we are able to create a placement which matches all clusters with at least 10G allocatable memory. In the first scheduling cycle, cluster1 is selected by this placement for it has 12G allocatable memory. After several minutes, the allocatable memory of cluster1 is reduced to 8G.  We don't know if the reduced 4G memory is consumed by workload associated with this placement. So in the next scheduling cycle, should cluster1 be selected by this placement or not?

elgnay: > how does a user handle it when their clusters are autoscaling? What do they expect to happen? This could be presented as there not being capacity available I think. Or perhaps that capacity is tight, but will expand.

Here is what will happens if managed clusters are autoscaling.
#### Scale up
1. User creates a placement with `ClusterResourcePreference`;
2. The cluster with most allocatable to capacity ratio or most allocatable will be selected;
3. Workload will then be deployed on the selected cluster;
4. If there is no enough resource available or the capacity is tight, new node will be added;
5. Capacity/allocatable of this cluster changes. That will trigger a new scheduling cycle for placements with ClusterResourcePreference;

#### Scale down
1. An underutilized node in a managed cluster is removed;
2. Capacity/allocatable of this cluster changes. That will trigger a new scheduling cycle for placements with ClusterResourcePreference;
3. Some Placements may no longer select this cluster because of resource change;

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
elgnay: Yes, they are related. Weight of `steady` plugin should be adjusted automatically according to the churning policy of a placement. And we may also support some advanced features, like `churningSeconds`, to describe to what extent we can stand for the cluster churning.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
elgnay: Corrected.

enhancements/sig-architecture/15-resourcebasedscheduling/README.md
elgnay: A generic prioritizing configuration is provided instead of `ClusterResourcePreference`

Issue #17, opened 2021-08-10
Select clusters based on its capability


qiujian16:
relates to https://github.com/open-cluster-management-io/community/issues/74

qiujian16:
done

PR #18, opened 2021-08-10
Cluster features with labels and clusterclaims

qiujian16:

### DIFF
--- a/enhancements/sig-architecture/17-cluster-features/README.md
+++ b/enhancements/sig-architecture/17-cluster-features/README.md
@@ -1,0 +1,110 @@
+# Managed cluster feature labels
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+There are requirements about selecting `ManagedClusters` based on the certain features of the managed cluster by `Placement` API. These requirements include:
+
+1. Select a set of the clusters that a certain `ManagedClusterAddon` is enabled. For instance, a user wants to install an application that needs multicluster service connectivity, the use will need to select a set of clusters that `Submariner` addon is enabled by using `Placement` API.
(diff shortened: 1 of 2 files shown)


qiujian16:
relates to https://github.com/open-cluster-management-io/enhancements/issues/17

qiujian16:
@itdove @tpouyer PTAL

enhancements/sig-architecture/17-cluster-features/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+### Feature labels
+
+Feature labels should have the prefix of `feature.open-cluster-management.io`. Each addon will be represented as a feature label in the format of `feature.open-cluster-management.io/addon-<addon name>`. The values of the lable are `available`, `unhealthy`, `unreachable`. For example, the `application-manager` addon will be represented as a label on `ManagedCluster` as:
+
+```
+feature.open-cluster-management.io/addon-application-manager: available
+```
+
```
itdove, line 47: 1. I guess the label will be automatically removed if the addon is disabled.
2. What will the latency for the label to be present?
3. Can we have also the negative, for identity provider we are also looking to clusters which doesn't have a feature. Will that be possilbe, for example we want to know the managedcluster which doesn't have GRC.

enhancements/sig-architecture/17-cluster-features/README.md
```diff
@@ -0,0 +1,110 @@
...
+## Motivation
+
+This proposal is to standardise the labels and `ClusterClaims` that are categorized as features of a `ManagedCluster`.  The idea is inspired by [node features](https://kubernetes-sigs.github.io/node-feature-discovery/) in kubernetes.
+
+### Goals
+
+- define the format of labels and `ClusterClaims` for `ManagedClusters` to represent a feature in managed cluster
+- implement addon as feature labels in `registration`.
+
```
itdove, line 26: If implemented in registration will the label be removed when the addon is disabled?

enhancements/sig-architecture/17-cluster-features/README.md
qiujian16: label select can have `notin`

enhancements/sig-architecture/17-cluster-features/README.md
qiujian16: it will be removed when related `ManagedClusterAddon`  resource is deleted. I will add this in the doc.

PR #19, opened 2021-08-17
Addon for apiserver-network-proxy

yue9944882:
@qiujian16

### DIFF
--- a/enhancements/sig-architecture/14-addon-cluster-proxy/README.md
+++ b/enhancements/sig-architecture/14-addon-cluster-proxy/README.md
@@ -1,0 +1,260 @@
+# Addon: Apiserver Network Proxy
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+Automating the installation of [apiserver-network-proxy](https://github.com/kubernetes-sigs/apiserver-network-proxy)
+so that admins or services can access the managed-clusters' kube-apiserver
+in an imperative approach but without retaining any per-cluster credentials
+in the hub cluster.
+
+## Motivation
+
+#### Mutual Cluster Connectivity
+
+Originally in OCM a minimal requirement for the underlying network 
+infrastructure is to support one-way network connectivity (spoke->hub) so 
+that the basic cluster registration and management can work. The
+apiserver-network-proxy will provide a reverse proxy tunnel over a long 
+connection and multiplexes packets through the tunnel so that hub and spoke 
+clusters are mutually accessible w/o raising more requirements to the 
+infrastructure. Note that the reverse tunnel is only supposed to transmit 
+"mouse flows" to avoid congestion.
+
+#### Direct Multi-Cluster Resource Access
+
+In some cases where we need to directly read the presence/state of specific 
+resources in the managed clusters, however for now we have no idea of 
+those resources unless they were created via the manifests api. Moreover,
+we don't have the fine-grained information about their fields and values 
+even if they are natively delivered by OCM. These problems will be eased
+in a degree if we can establish a proxy tunnel between hub and spoke 
+clusters.
+
+#### Avoid Potential Credential Leakage
+
+Normally we're supposed to obtain the credential (w/ sufficient permission)
+from the spoke cluster before raising api calls against the spoke cluster, 
+but that requires us to copy and dump the credential out of the spoke 
+cluster which exposes the cluster under potential danger of being accessed 
+by unexpected users. Also in some cases where the hub/spoke clusters are
+mutually untrusted, copying the credentials will be strictly impossible.
+
+
+### Goals
+
+- Automate the installation and operation and apiserver-network-proxy via 
+  the addon-framework.
+- Automate the certificate rotation of both proxy-servers and proxy-agents.
+
+### Non-Goals
+
+- Customizing the source code of apiserver-network-proxy.
+
+## Proposal
+
+#### Architecture
+
+
+The following figure shows the interaction between the components:
+
+```
+   
+                     -----------------------------------------
+                     | Hub-Cluster                           |
+                     |                                       |
+  ----------------   | ----------------     ---------------  |
+  | Proxy Client |> > >| Proxy Server | <===| Proxy Addon |  |
+  ----------------   | ----------------     |  Controller |  |
+                     |      V               ---------------  |
+                     |      V                    ‖           |
+                     |    / V \                  ‖           |
+                     ----/| V |\-----------------‖------------
+                          | V |                  ‖
+                   Reverse| V |Tunnel            ‖ Install/Operate
+                          | V |                  ‖ 
+ -------------------------| V |-------------     ‖
+ | Managed-Cluster        | V |            |     ‖
+ |                        | v |            |     ‖
+ | -----------------   ---------------     |     ‖
+ | | KubeApiserver |< <| Proxy Agent | <==========
+ | -----------------   ---------------     |
+ -------------------------------------------
+
+```
+
+- A proxy-agent that establishes a long connection towards the proxy-server
+  and register itself w/ the cluster's identity.
+- A proxy-server that receives handshakes from proxy-agent and maintains 
+  tunnel over the long connection.
+- A proxy-client requesting the target kube-apiserver through the tunnel
+  over either websocket HTTP-CONNECT or grpc protocol.
+- An addon controller automates the installation of proxy servers and proxy 
+  agents by consuming the corresponding configuration custom resource. And
+  optionally the proxy-agents can aggregate and report the status upwards
+  so that admins can checkout their healthiness easily w/o even touching
+  the managed-clusters.
+
+
+#### Components
+
+According to the latest design and implementation of [addon-framework](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/8-addon-framework),
+we will bring a new agent controller named `cluster-proxy-agent-controller` that
+qualifies the interfaces defined in addon-framework and automatically does the 
+following things upon discovering a new `ManagedClusterAddon` resource:
+
+1. Signing all the required X509 certificates and keys beforehand:
+   
+    i.   a CSR requesting server certificate which will be shared across the 
+         replicas of the proxy servers.
+   
+    ii.  a CSR requesting agent certificate which will be shared by the proxy 
+         agents.
+   
+    iii. a CSR requesting client certificate for logging onto in the tunnel.
+
+2. A configuration custom resource that contains the following information: 
+
+    - the expected replicas of proxy-servers.
+      
+    - the proxy protocol.
+      
+    - the server/agent ports
+    
+3. Installing proxy-server in the cluster namespace in the hub cluster.
+   
+4. Installing proxy-agent in an arbitrary namespace in the spoke cluster with 
+   bootstrap credentials configured.
+   
+5. Periodically probing the connectivity healthiness of the tunnel.
+
+#### Expected Usage
+
+Basically all the client library supporting HTTP proxy are expected to use 
+the proxy addons w/o additional work. As a matter of fact, most of those 
+popular HTTP client libraries only supports unsecured HTTP proxy instead of 
+mTLS-secured HTTPS proxy, and neither golang SDK nor JDK has builtin support 
+for mTLS proxy. So at the cost of securing the proxy server further, we will 
+need to adapt HTTPS proxy by our own effort. The good news is that the
+apiserver-network-proxy provided packages to talk over mTLS proxy. And also 
+there are a few https client binaries that natively supports mTLS proxy e.g.
+`curl` in the latest versions as is shown below:
+
+```bash
+curl -v -p \
+     --proxy-key certs/frontend/private/proxy-client.key \
+     --proxy-cert certs/frontend/issued/proxy-client.crt \
+     --proxy-cacert certs/frontend/issued/ca.crt \
+     --proxy-cert-type PEM \
+     -x https://127.0.0.1:8090 \   # the proxy-server's address
+     http://localhost:8000/success # any target L4 address
+````
+
+
+#### API Spec
+
+ManagedClusterAddon:
+
+```yaml
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ClusterManagementAddon
+metadata:
+ name: apiserver-network-proxy
+spec:
+ addonMeta:
+   displayName: apiserver-network-proxy
+   description: "https://github.com/kubernetes-sigs/apiserver-network-proxy"
+ addonConfiguration:
+   crd: proxy.addon.open-cluster-management.io
+   name: proxy-config
+```
+
+ProxyConfiguration:
+
+```yaml
+apiVersion: proxy.addon.open-cluster-management.io/v1alpha1
+kind: ProxyConfiguration
+metadata:
+ name: apiserver-network-proxy
+spec:
+  server:
+    replicas: 2
+    protocol: [websocket|grpc]
+    tunnel-port: 8091
+    proxy-port: 8090
+  agent:
+    replicas: 2
+    server-address: 4.4.4.4
+```
+
+
+#### Framework Integration
+
+Based on the extensibility provided by the addon-framework, the new proxy addon
+will be implementing the required interfaces in the following behavior:
+
+
+```go
+var _ AgentAddon = &ApiserverNetworkProxyAddon{}
+
+func (a *ApiserverNetworkProxyAddon) Manifests(cluster *clusterv1.ManagedCluster, config runtime.Object) ([]runtime.Object, error) {
+    // 1. a service-account with fundamental spoke cluster permissions.
+    // 2. a deployment installing the proxy-agents
+}
+
+func (a *ApiserverNetworkProxyAddon) GetAgentAddonOptions() AgentAddonOptions {
+	// using the `ProxyConfiguration` above as options
+}
+
+var _ AgentAddonWithRegistration = &ApiserverNetworkProxyAddon{}
+
+func (a *ApiserverNetworkProxyAddon) HubRBAC(cluster *clusterv1.ManagedCluster, group string) (*rbacv1.Role, *rbacv1.RoleBinding) {
+	// none
+}
+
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+1. Signing all the required X509 certificates and keys beforehand:
+   
+    i.   a CSR requesting server certificate which will be shared across the 
+         replicas of the proxy servers.
+   
+    ii.  a CSR requesting agent certificate which will be shared by the proxy 
+         agents.
+   
+    iii. a CSR requesting client certificate for logging onto in the tunnel.
```
skeeey, line 121: Will these CRSs be created and approved by user in advance? or the `cluster-proxy-agent-controller` will handle this?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,260 @@
...
+  server:
+    replicas: 2
+    protocol: [websocket|grpc]
+    tunnel-port: 8091
+    proxy-port: 8090
+  agent:
+    replicas: 2
+    server-address: 4.4.4.4
+```
+
```
skeeey, line 196: We also need the [`ManagedClusterAddOn `](https://github.com/open-cluster-management-io/api/blob/main/addon/v1alpha1/types_managedclusteraddon.go) to trigger the anp installation on each managed cluster

skeeey: And I think we may consider to add some status in the config/addon apis to check the anp and its connection status

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: i think so, the agent CSRs should be auto-approved by the operator if (1) the managedcluster is present (2) the managedclusteraddon is present

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: @skeeey i just added a few details about the config coordinates spec, let me know if i missed anything

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -50,13 +50,21 @@ mutually untrusted, copying the credentials will be strictly impossible.
 
 ### Goals
 
+P1:
```
qiujian16, line 53: alpha I think

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+    rotation:
+      expiringDays: 365
+      forceReloadDays: 200
+  proxy: 
+    mode: [http-connect|grpc|uds] 
+  deploy:
+    namespace: open-cluster-management
+    agentAuthentication:
+      namespace: <validating namespace>
+      serviceAccount: <validating service-account>
```
qiujian16, line 158: I think it could be a fixed name, will you expect user mutate this field?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+
+
+##### ManagedProxyServer
+
+The `ManagedProxyServer` is supposed to be a cluster-scoped resource and holds 
+information for provisioning an active proxy-server instance. An example will 
+be:
+
+```yaml
+apiVersion: addon.open-cluster-management.io/v1alpha1
```
qiujian16, line 139: proxy.open-cluster-management.io might be better?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+- __MangedProxyAgent__: the configuration and status of proxy-agents.
+
+Additionally, the operator will be list-watching the standard `ManagedClusterAddon`
+resource and automatically installs proxy-agents w/ the expected options
+upon discovering the addon in each cluster's namespace.
+
+
+##### ManagedProxyServer
+
+The `ManagedProxyServer` is supposed to be a cluster-scoped resource and holds
```
qiujian16, line 134: Why do we need it to be a cluster scoped resource or singleton? I think it is possible that we could have multiple servers,  each with a ManagedClusterSet also?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+  serverReference: 
+    # A pointer to the server configuration so that auto-signed certificates
+    # and other related config can be imported.
+    apiVersion: "addon.open-cluster-management.io/v1alpha1"
+    resource: managedproxyservers
+    name: default
+  replicas: 2
+  proxyServerAddress: 1.1.1.1
+  clusterServiceAccount: open-cluster-management:proxy-agent:egress
+  certificates:
```
qiujian16, line 220: What  is the intention to this secret?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ManagedProxyAgent
+metadata:
+  name: default # Or any unique name, will be the name of the workload
+spec:
+  serverReference: 
+    # A pointer to the server configuration so that auto-signed certificates
+    # and other related config can be imported.
+    apiVersion: "addon.open-cluster-management.io/v1alpha1"
+    resource: managedproxyservers
```
qiujian16, line 215: Do we expect there will be other resource option here?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+	// using the `ProxyConfiguration` above as options
+}
+
+var _ AgentAddonWithRegistration = &ApiserverNetworkProxyAddon{}
+
+func (a *ApiserverNetworkProxyAddon) HubRBAC(cluster *clusterv1.ManagedCluster, group string) (*rbacv1.Role, *rbacv1.RoleBinding) {
+	// none
+}
+
+func (a *ApiserverNetworkProxyAddon) BootstrapKubeConfig(cluster *clusterv1.ManagedCluster) ([]byte, error) {
```
qiujian16, line 324: It seems not the latest addon framework interface. We simplify this so a bootstrapKubeConfig is not really needed.

qiujian16 (commented):
Thanks, @yue9944882 it looks great. Would you like to add a bit about testing plan? Also I think it would need a metadata.yaml in the same directory.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,362 @@
...
+# listing managed cluster's namespaces
+kubectl get --raw="/apis/<the addon api-group>/<the addon api-version>/<addon-resource-name>/<managed cluster name>/proxy/api/v1/namespaces"
+```
+
+And the official golang and java SDKs are supposed to talk to the managed cluster
+by simply adding an api prefix to the requests.
+
+#### Proxy Server Authorization 
+
+Even if the proxy-servers are mTLS-secured, all the client talking to the servers
```
qiujian16, line 359: I think one option is we can transparently pass the credential of the request to the spoke, if all spoke are at the same id provider.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: firstly for clarification, the things under `agentAuthentication` is optional and meant for authorizing the agent handshakes. e.g. proxy-server should only accepts agents mounting expected service-accounts. i dont completely understand the purpose of such additional authorization b/c the agent handshakes are already secured/authenticated by mTLS layer. 

> will you expect user mutate this field?

i picked a few outstanding configuration options from the whole spill. there're more flags such as loggings. i think we can start by keeping it in case any user will need additional authorization between server <-> agent. wdyt?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: can you elaborate how to make it a singleton in k8s api? practically i think singleton should be sufficient, but having forward-compatibility of multiple groups of servers isnt a bad idea, is it? i can imagine cases where we need to isolate a new tunnel so that protecting some traffic for critical usage from bandwidth congestion

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: the agent and servers talks over the mTLS layer, which requires the agents to have their client x509 certificate

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: no actually, i was just following the prior art from rbac's role and rolebinding

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: @qiujian16 thats exactly now `agentAuthentication` works we talked above. i will eliminate this section.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
skeeey: I think it is clear for me

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,381 @@
...
+information for provisioning an active proxy-server instance. An example will 
+be:
+
+```yaml
+apiVersion: proxy.open-cluster-management.io/v1alpha1
+kind: ManagedProxyServer
+metadata:
+  name: default # Or any unique name, will be the name of the workload
+spec:
+  replicas: 1
```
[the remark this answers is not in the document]
yue9944882: i presume multiple replicas will be increasing both high-availability and throughput:

- the proxy-agents is working in simply stateless active-active replicas b/c proxy-servers will be distributing traffic to those registered agents (currently in round-robin strategy).
- the proxy-servers' scaling is a bit tricky. initially proxy-agents doesnt know the actual # of proxy-servers' replicas behind the entry load-balancer, so during the process when proxy-agent is registering itself to the any of the proxy-server, it will be told the identity (a UUID by default) of the proxy-server and also an expected # of replicas via the grpc protocol. so the proxy-agent will be keep attempting the registration until it was successfully registerred to the expected number of servers.

given that all the replicas of proxy-servers and proxy-agents are active, i think scaling up will not only improve the failure/shutdown-tolerance but also the throughput (if the tunnel is working over public internet, the throughput will be converging to a bottleneck due to the bandwidth i think..)

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,381 @@
...
+  certificates:
+    signingSANs: 
+      - "1.1.1.1"
+      - "example.com"
+    rotation:
```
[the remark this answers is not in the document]
yue9944882: @deads2k the current api spec's smallest granularity is "day". do you think we will need minute-ish rotation in development?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,381 @@
...
+
+The tunnel from agent to server is secured by mTLS technique, the same way
+we're doing for securing kubernetes' apiserver. So it's almost safe to expose
+the proxy-servers' entry point even in the public networks unless it is 
+flooded by DDoS attacks.
+
+
+##### Server <-> Proxy Authorization
+
+Based on the authentication above, the proxy-server can opt-in to verify
```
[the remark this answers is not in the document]
yue9944882: okay, i thought opting-out from authz will make it easier for development tho. but yes, we can force the authz in the framework

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,381 @@
...
+the agents' identity by asking the agents to pass its service-account-token 
+in the handshake attempts. The proxy-servers can either validate the token
+by its signed namespace or the name of the service-account.
+
+
+#### Expected Usage
+
+
+##### Direct HTTP calls
```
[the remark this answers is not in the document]
yue9944882: will be adding a new section elaborating this.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,536 @@
...
+client -mTLS(2)->                                             T1: Connects proxy server   
+       --------->                                             T2: Requests a proxied stream
+       --------------------- mTLS(3) ----------------->       T3: Unwinds kube-apiserver mTLS
+       ----------------- k8s api requests ------------>       T4: Sends request payload 
+```
+
+##### Server <-> Proxy Authentication
+
+The tunnel from agent to server is secured by mTLS technique, the same way
+we're doing for securing kubernetes' apiserver. So it's almost safe to expose
```
xuezhaojun, line 345: Here is a little concern, if we expose anp proxy-server directly, in the future we want to replace another proxy approach say remotedialer, it would be painful since the tunnel layer is not interchangeable.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,536 @@
...
+  certificates:
+    signing:
+      additionalSANs:
+        - "1.1.1.1"
+        - "example.com"
+      rotation:
+        expiringDays: 365
+        forceReloadDays: 200
+    bootstrap:
+      secret:
```
qiujian16, line 163: this might not be needed. registration can help each addon to generate a csr with customized signer. So it is not needed to provide an additional bootstrap secret.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,381 @@
...
+
+The following figure shows the interaction between the components:
+
+```
+   
+                     -----------------------------------------
+                     | Hub-Cluster                           |
+                     |                                       |
+  ----------------   | ----------------     ---------------  |
+  | Proxy Client |> > >| Proxy Server | <===| Proxy Addon |  |
```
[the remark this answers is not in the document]
yue9944882: > There is a concern about proxy-server, if an agent register itself's name to proxy-server, and proxy-server cache this in memory. Then it would not be stateless.

sorry to be dense. can you elaborate what you mean by "stateless"?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: can you share me a link to the alternative proxy approach you mentioned? i agree the proxy framework doesnt have to be coupled with apiserver-network-proxy tho, we can leave a bit forward-compatibility for the alternatives.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: oh true, we can reuse the bootstrap secret from registration agent

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
xuezhaojun: Sure, we have considered adding https://github.com/rancher/remotedialer as an alternative approach in the future.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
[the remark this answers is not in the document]
xuezhaojun: Sure, in our practice, we specify the number of instances for proxy-server to 1 because I thought the proxy-server cached some registration information.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,545 @@
...
+  addonMeta:
+    displayName: apiserver-network-proxy
+    description: "https://github.com/kubernetes-sigs/apiserver-network-proxy"
+  addonConfiguration:
+    # Importing templates for creating real agent instances.
+    crd: managedproxyconfigurations.proxy.addon.open-cluster-management.io
+    name: default
+```
+
+##### CredentialProjection
```
qiujian16, line 274: let's remove this  section

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,545 @@
...
+
+##### Server <-> Proxy Authentication
+
+The tunnel from agent to server is secured by mTLS technique, the same way
+we're doing for securing kubernetes' apiserver. So it's almost safe to expose
+the proxy-servers' entry point even in the public networks unless it is 
+flooded by DDoS attacks.
+
+
+##### Server <-> Proxy Authorization
```
qiujian16, line 350: I think this part relates to mtls also?

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
```diff
@@ -0,0 +1,545 @@
...
+```
+
+
+## Graduation Critiria
+
+Alpha:
+
+- Proxy servers and agents can be automatically installed/upgrade by the new addon-manager.
+- A breif status of the proxy servers and agents can be aggregated and reflected on the
+  addon resource.
```
qiujian16, line 518: Also add a beta graduation

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: done, also eliminated connecting section in the goal list

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: i rephrased this section a bit for clarification, plz take another look, thx.

enhancements/sig-architecture/14-addon-cluster-proxy/README.md
yue9944882: done, clarified alpha & beta missions

qiujian16:
Thanks this is looks good to me with current scope and graduation plan. 

/approve
/lgtm

PR #23, opened 2021-09-14
Update 15-resourcebasedscheduling

haoqing0110:
Update 15-resourcebasedscheduling

### DIFF
--- a/enhancements/sig-architecture/15-resourcebasedscheduling/metadata.yaml
+++ b/enhancements/sig-architecture/15-resourcebasedscheduling/metadata.yaml
@@ -8,5 +8,7 @@ approvers:
   - "@qiujian16"
   - "@mdelder"
 creation-date: 2021-08-06
-last-updated: 2021-08-25
-status: provisional
+last-updated: 2021-09-17
+status: implemented
+see-also:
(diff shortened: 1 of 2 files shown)


qiujian16 (commented):
thanks, just one thing.

in Graduation Criteria, I think we should add an item for alphat->beta that we need to revisit the API shape before upgrade to beta based on user's feedback

enhancements/sig-architecture/15-resourcebasedscheduling/metadata.yaml
```diff
@@ -8,5 +8,7 @@ approvers:
   - "@qiujian16"
   - "@mdelder"
 creation-date: 2021-08-06
-last-updated: 2021-08-25
-status: provisional
+last-updated: 2021-09-17
+status: implemented
+see-also:
+  - "/enhancements/sig-architecture/15-resourcebasedscheduling/README.md"
```
qiujian16, line 14: point to the original placement proposal https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/6-placements

haoqing0110: fixed

PR #24, opened 2021-09-16
KEP: Managed ServiceAccount

yue9944882:
this KEP is decoupled from https://github.com/open-cluster-management-io/enhancements/pull/19 aiming at solving the authentication/authorization against each managed cluster. the addon proposed by this KEP should be working w/ or w/o the former proxy KEP depending on whether the hub->agent network connectivity is working.

### DIFF
--- a/enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
+++ b/enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
@@ -1,0 +1,171 @@
+# Addon: Managed ServiceAccount
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal will be adding a new OCM addon named `managed-serviceaccount` 
+which helps the administrators to automatically distribute service-accounts
+to the managed clusters and manage the lifecycle of those service-accounts 
+including token rotation, ensuring token validity, destroying the token, etc.
+The token of the distributed service-account will also the periodically
+projected / reported to the hub cluster as a secret resource with the same
+name and the same namespace as the corresponding "ManagedServiceAccount" 
+resource. A valid "ManagedServiceAccount" should be either in a "cluster 
+namespace" or a "cluster-set namespace" which is bind to a valid 
+"ManagedClusterSet" resource.
+
+## Motivation
+
+### Distribute and manage a multi-cluster service-account
+
+Creating such a "multi-cluster service-account" will help us easily distribute
+service-accounts to the managed cluster and on the other hand destroying the 
+"multi-cluster service-account" will manage the lifecycle of ensuring, 
+revoking, removing the service-accounts from each of the managed cluster. Not 
+only the addon will protect these distributed service-account from unexpected 
+modification and removal, and also it will synchronize the lifecycle of
+the service-accounts according to its parent resource "multi-cluster 
+service-account".
+
+### Token projection
+
+After ensuring the service-accounts' presence on each managed cluster, we will
+be able to dynamically asking the managed cluster for signing a token w/ expected
+token audience prescribed. The existing `authenticationv1.TokenRequest` api is 
+particularly designed for such usages, so it's rather simple for the addon 
+to request new token from each managed cluster as long as rbac permission of
+token request is granted.
+
+### Authenticate hub->spoke outbound requests
+
+In some cases when a component (such as an operator) deployed in the hub 
+cluster needs to proactively invoke the kube-apiservers of the managed 
+clusters, the outbound requests ought to be authenticated before being 
+served by the kube-apiserver. Then attaching the token to those outbound 
+requests will pass the managed cluster's authentication.
+
+
+## Goals & Non-Goals
+
+### Goals
+
+- Manage the lifecycle of service-accounts across multiple cluster.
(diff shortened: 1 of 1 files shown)


enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
```diff
@@ -0,0 +1,159 @@
...
+### A Highly-secured namespace for persisting the tokens
+
+The addon-agents need to find a place to persist the tokens after they finally
+succeed obtaining tokens via `authenticationv1.TokenRequest` api. In this 
+proposal, we're expecting all these tokens to be persisted in a single 
+highly-secured namespace such as `managed-cluster-credentials`. To protect the 
+tokens from being copied for other unexpected usage, the namespace should not
+be accessible to most of the hub cluster users except:
+
+1. the addon-agent for rotation the token. (write-only to its managing secret)
```
deads2k, line 143: can you be specific about what verbs?  YOu cannot have named create, so you're thinking about a named apply?

deads2k:
What is the intended consumer of these serviceaccount tokens?  It seems as though creating these would encourage patterns of having a controller in the hub acting on the spokes.  Is that the objective?

If you grant individual rights for spokes to write the tokens into a single project, the fanout for the number of RBAC roles and RBAC rolebindings in the destination namespace seems extremely high.

enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
yue9944882: oh rbac doesnt support named create, is it? so yes, a named server-side-apply can be better than named create.

yue9944882:
> What is the intended consumer of these serviceaccount tokens? It seems as though creating these would encourage patterns of having a controller in the hub acting on the spokes.

less about the operators in the hub, more about the web services (e.g. UI dashboard for showing the status of each managed clusters). i didnt like the operators to raise watches to the managed clusters by the tokens b/c this is conflicting w/ the original architecture of OCM --- the commands/operations should be "pull"ed by the agents instead of "push"ing by a central operator. probably we should ban the long-running requests using this token.

> If you grant individual rights for spokes to write the tokens into a single project, the fanout for the number of RBAC roles and RBAC rolebindings in the destination namespace seems extremely high.

yeah, so intentionally i didnt expect this KEP to cover the RBAC authorization upon the projected tokens. this KEP will be merely focusing on just the token, or multi-cluster authentication. as for multi-cluster authorization, i think there should be another set of API to control the distribution of RBAC policies.

enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+```yaml
+# `ManagedServiceAccount`is cluster-scoped
+apiVersion: proxy.open-cluster-management.io/v1alpha1
+kind: ManagedServiceAccount
+metadata:
+  name: my-foo-operator-serviceaccount
+spec:
+  provision: 
+    namespace: default
```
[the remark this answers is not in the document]
yue9944882: the reported token will be under the same namespace of `ManagedServiceAccount` resource

enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
```diff
@@ -0,0 +1,159 @@
...
+and additionally there will also be a few new custom resources:
+
+A sample of the `ManagedServiceAccount` will be:
+
+```yaml
+# `ManagedServiceAccount`is cluster-scoped
+apiVersion: proxy.open-cluster-management.io/v1alpha1
+kind: ManagedServiceAccount
+metadata:
+  name: my-foo-operator-serviceaccount
```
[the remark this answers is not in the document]
yue9944882: yes, or a cluster-set namespace which means the serviceaccount shall be replicated

enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
```diff
@@ -0,0 +1,159 @@
...
+Then addon-agent will be list-watch after started and dynamically provisions
+local service-accounts and persist the token to the hub cluster based on the 
+rotation policy.
+
+### A Highly-secured namespace for persisting the tokens
+
+The addon-agents need to find a place to persist the tokens after they finally
+succeed obtaining tokens via `authenticationv1.TokenRequest` api. In this 
+proposal, we're expecting all these tokens to be persisted in a single 
+highly-secured namespace such as `managed-cluster-credentials`. To protect the
```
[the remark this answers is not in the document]
yue9944882: exactly

enhancements/sig-architecture/19-projected-serviceaccount-token/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+
+The `ManagedServiceAccount` resource is expected to be created under:
+
+- "cluster namespace": a namespace with the same name as the managed cluster, 
+  which indicates that the `ManagedServiceAccount` will be bound to a service
+  account with the same name as the `ManagedServiceAccount` and under the 
+  namespace where the "managed-serviceaccount-addon-agent" deploys.
+  
+- "cluster-set namespace": a namespace with valid clusterset bounded, which
```
qiujian16, line 112: maybe we start with cluster namespace only at first. I think creating this on clusterset ns might have some inconsistent behavior. A related secret in clusterset ns will not be generated, but it will be generated in cluster ns, right?

yue9944882: @qiujian16 sure we can def defer that in the future stage, and our current poc code only applies to the cluster ns btw

qiujian16:
/approve
/lgtm
/hold

hold in case anyone else would need to review.

qiujian16:
/unhold

no additional feedback is obtained.

PR #30, opened 2021-10-20
Specify status sync in ManifestWork

qiujian16:

### DIFF
--- a/enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
+++ b/enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
@@ -227,6 +227,12 @@ status:
           version: v1
 ```
 
+The returned value must be a scalar value, and the work agent should check the type of the returned value. The work agent should
+treat the status that a value "ISNotFound" or "TypeMismatch" separately. If the path of the interestedValue is not found in the
+status of the resource, this values should be ignored. If the path of the interestedValue is valid, but the type of the value is
+not scalar, e.g a list or map, the condition of "StatusFeedbackAvailable" should be set false and a message shoud be added to
+indicate that which intrestedValue cannot be obtained
+
 #### Status update frequency
 
 Ideally the work agent should have a informer for each applied resource to update interestedValue. It will need to
--- a/enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
+++ b/enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
@@ -47,7 +47,7 @@ In the spec of `ManifestWork`, add a new field
 type ManifestWorkSpec struct {
     ...
 	// StatusFeedback defines the option to return status of applied resource
-	StatusFeedbackOptions []StatusFeedbackOption `json:"statusFeedbackOption,omitempty"`
+	StatusFeedbacks []StatusFeedbackOption `json:"statusFeedbacks,omitempty"`
 }
 
 type StatusFeedbackOption struct {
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+In the spec of `ManifestWork`, add a new field 
+
+```go
+
+// ManifestWorkSpec represents a desired configuration of manifests to be deployed on the managed cluster.
+type ManifestWorkSpec struct {
+    ...
+	// StatusFeedback defines the option to return status of applied resource
+	StatusFeedbackOptions []StatusFeedbackOption `json:"statusFeedbackOption,omitempty"`
```
deads2k, line 50: nit: "options" suffix is unnecessary

enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+type StatusFeedbackOption struct {
+	// ResourceIdentifier represents the group, resource, name and namespace of a resoure.
+	// It must match to a resource defined in spec.workload.manifests
+	// +required
+	ResourceIdentifier ResourceIdentifier `json:"resourceIdentifier"`
+
+	// FeedbackOption defines the option of how status can be returned.
+	// It is IntrestedField only today, but can be extend to others in the future.
+	FeedbackOption string `json:"feedbackOption"`
```
deads2k, line 61: put this into its own struct alongside InterstedFields

enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
```diff
@@ -0,0 +1,243 @@
...
+### Upgrade Strategy
+It will need upgrade on CRD of ManifestWork on hub cluster, and upgrade of work agent on managed cluster.
+
+### Version Skew Strategy
+- The StatusFeedback field is optional, and if it is not set, the manifestwork can be correctly treated by work agent with elder version
+- The elder version work agent will ignore the StatusFeedback field.
+
+## Alternatives
+
+### Return raw status data of the applied resource.
```
deads2k, line 235: Glad to see this section filled in.  I think this analysis is correct.

enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
```diff
@@ -0,0 +1,253 @@
...
+	// +required
+	ResourceIdentifier ResourceIdentifier `json:"resourceIdentifier"`
+
+	// FeedBackRules defines the returned resource status feedback
+	FeedbackRule FeedbackRule `json:"feedbackRule"`
+}
+
+type FeedbackRule struct {
+	// Type defines the option of how status can be returned.
+	// It is IntrestedField only today, but can be extend to others in the future.
```
deads2k, line 65: nit: typo interested

enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
```diff
@@ -0,0 +1,253 @@
...
+	Type string `json:"type"`
+
+	// InterstedFields defines the status fields to be synced for a manifest
+	// An interstedValues will be updated in certain manifest status of the manifestwork.status.
+	InterestedFields []InterestedResourceField `json:"interestedFields,omitempty"`
```
deads2k, line 70: In usage, it seems like the same sorts of fields will be requested many times.  Every deployment will likely have an available replica count of something similar.  This can result in subtly different alias names and perhaps even the same alias with different jsonpaths.

While I agree that a design goal of allowing any CustomResource to have a field extracted is valuable, it seems like making the common-case easy and consistent has significant value.  Can add a section to the "alternatives considered" section that shows what an API that allows for some well-known options like the workload status that has `Replicas`, `ReadyReplicas`, and `AvailableReplicas`

qiujian16: Added another mode WellKnownStatus so with that use does not need to specify json path. The returned value will have the fixed name.

deads2k:
> Added another mode WellKnownStatus so with that use does not need to specify json path. The returned value will have the fixed name.

I think this will help a lot with consistency of consumption.  The design looks implementable to me and the rest of the API shape looks like things we can work out in an API PR.

/lgtm
/hold

holding for 48 hours in case there are reviewers also looking at the new updates.

enhancements/sig-architecture/29-manifestwork-status-feedback/README.md
```diff
@@ -0,0 +1,243 @@
...
+        - lastTransitionTime: "2021-10-14T14:59:09Z"
+          message: Resource is available
+          reason: ResourceAvailable
+          status: "True"
+          type: Available
+        statusFeeback:
+          interestedValues:
+          - fieldValue:
+              integer: 1
+              type: integer
```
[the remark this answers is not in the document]
yue9944882: IIUC all the feedbacks will be treated as plain yaml content for any kind of resource. currently supported types in yaml are:

- null
- boolean
- float
- integer
- string
- list
- map

i think we dont need to deal w/ int-or-string specifically. just clarify the failure path when the expected type in the spec doesnt match the actual feedback will be sufficient. e.g. say if we asserted `boolean` but the result turns out to be a string, then how should that be reflected in the status? my first thought is that we should explicitly show the failures to the feedback consumers, the feedback should deliver the message accurately otherwise it can be extra complex for a consumer (e.g. a multi-cluster operator) to read the feedback.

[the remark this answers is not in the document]
qiujian16: Added a section to indicate error checking.

yue9944882 (commented):
/lgtm

Issue #32, opened 2021-11-02
placement extensible scheduling


haoqing0110:
The related feature request: open-cluster-management-io/community#52

When implementing placement resource based scheduling, we find in some cases, prioritizer needs extra data (more than the defualt value provided by ManagedCluster and ManagedClusterInfo) to calculate the score of the managed cluster.

For example, prioritizer ResourceRatioCPU depends on knowing the real-time CPU usage, which is not provided by ManagedCluster. So we want a more extensible way to support scheduling based on customized value.

haoqing0110:
close as proposal https://github.com/open-cluster-management-io/enhancements/pull/33 merged.

PR #33, opened 2021-11-02
placement extensible scheduling proposal

haoqing0110:
Ref: https://github.com/open-cluster-management-io/enhancements/issues/32
Signed-off-by: haoqing0110 <qhao@redhat.com>


enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/website/)
+
+## Summary
+
+The proposed work provides an API to represent a scalar value of the managed cluster, to support placement extensible scheduling.
+
+## Motivation
+
+When implementing placement resource based scheduling, we find in some cases, prioritizer needs extra data (more than the defualt value provided by `ManagedCluster` and `ManagedClusterInfo`) to calculate the score of the managed cluster.
```
qiujian16, line 17: remove ManagedClusterInfo here

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+
+For example, prioritizer ResourceRatioCPU depends on knowing the real-time CPU usage, which is not provided by `ManagedCluster`.
```
qiujian16, line 19: For example, there is requirement to schedule based on resource monitoring data from the cluster. ResourceRatioCPU here is confusing.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+So we want a more extensible way to support scheduling based on customized value.
+
+### User Stories
+
+#### Story 1: The placement controller could prioritize clusters by data pushed from each managed cluster.
+  - As an end user, can define prioritizer name `customized-resourceratiocpu` in placment spec.prioritizerPolicy.configurations to say he wants to schedule based on customized value resourceratiocpu.
```
qiujian16, line 25: As an end user, I can ...

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+  - On each managed cluster, there would be a customized agent to sum up the cpu usage of the cluster. Then update the value into the ManagedClusterScalar CR with name `customized-resourceratiocpu`.
```
qiujian16, line 26: let's not mention ManagedClusterScalar here, this is the proposal not the use case.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+  - On the hub, there would be a customized agent to read cpu usage from thanos. It will generate a value for each cluster and update it into ManagedClusterScalar CR with name `customized-resourceratiocpu`.
+  - On the hub, the placement prioritizer will read value from ManagedClusterScalar `customized-resourceratiocpu` and score based on it.
+
+#### Story 3: The placement controller could prioritize clusters for disaster recovery.
+  - As an end user, can define prioritizer name `customized-dr` in placment spec.prioritizerPolicy.configurations to say he wants to schedule workload based on disaster recovery score.
+  - On the hub, there would be a customized agent to give cluster in the frist DC the highest score.
+  - On the hub, the placement prioritizer will select the cluster in the frist DC. Then disaster recovery workload will be running on it.
+  - When this cluster is unavailable (a taint is added), the customized agent will give cluster in the second DC the highest value.
+  - On the hub, the placement prioritizer will re-schedule to select the cluster in the second DC. Then disaster recovery workload will be running on the second DC.
```
qiujian16, line 39: I think there would be another story that a third party controller can comprehensively evaluate cluster on latency, region, iops etc., and rate cluster according to the SLA the cluster can provide.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Goals
+- Design an new API(CRD) to contain a scalar value for each managed cluster.
```
qiujian16, line 42: s/an/a/

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Namespaced"
+// +kubebuilder:subresource:status
+
+// ManagedClusterScalar represents a scalable value (aka score) of one managed cluster.
+// Each ManagedClusterScalar only represents the score for one specific calculator type.
+// ManagedClusterScalar is a namesapce scoped resource.
+//
+// The ManagedClusterScalar name should follow the format {cluster name}-{calculator name}.
```
qiujian16, line 64: I do not think it needs {cluster name}, but it needs to mention here that the namespace of the resource is the cluster ns.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+
+	// Score contains a scalable value of this managed cluster.
+	Score int64 `json:"score,omitempty"`
+
+	// ExpireTime is the expire time the score.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=date-time
+	ExpireTime metav1.Time `json:"expireTime" protobuf:"bytes,4,opt,name=lastTransitionTime"`
```
qiujian16, line 90: I think you want it to be pointer and nil means never expire

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+
+//ManagedClusterScalarStatus represents the current status of ManagedClusterScalar.
+type ManagedClusterScalarStatus struct {
+	// Conditions contains the different condition statuses for this managed cluster score.
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// Score contains a scalable value of this managed cluster.
+	Score int64 `json:"score,omitempty"`
+
+	// ExpireTime is the expire time the score.
```
qiujian16, line 85: maybe it should be ValidUntil. And the doc should be 

ValidUntil defines the time this score is valid, after this time, the Score is considered to be invalid by placement. The controller ownning this resource should keep the score update-to-date.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+//
+// The ManagedClusterScalar name should follow the format {cluster name}-{calculator name}.
+// For example, a calculator named ResourceAllocatableMemory can calculate the totale allocatable memory
+// of one cluster.
+// So for cluster1, the corresponding ManagedClusterScalar name is cluster1-resourceallocatablememory.
+type ManagedClusterScalar struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Status represents the status of the ManagedClusterScalar.
```
qiujian16, line 72: I still think you need spec and have prioritizerPluinName in it.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+Add a new plugin called customized plugin, which can score clusters based on ManagedClusterScalar CR `customized-{prioritizer-name}`.
+
+### 4. Workflow
+Refer to [Examples](#Examples)
+
+### 5. Open Questions
+1. Is one score enough? any other missed use cases? For example, spreading depends on previous choice.
+
+## Alternative
+Using scheduler extender mechanism, to support schedule based on metrics, schedule based on SLA.
```
qiujian16, line 138: this is not very clear, would you elaborate?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+Pros: 
+  - More flexible compared with above proposal can only store one score.
+
+Cons:
+  - Calling extender by http/https may have performance issue.
+
+## Examples
+
+### 1. An agent on managed cluster update ResourceRatioCPU score to API, and placement prioritizer plugin score based on it. (Use Story 1)
+![](./pic1.png)
```
qiujian16, line 149: it is not a good way to add pic in this doc.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+```
+2. Placement-controller watch the placement changes and generate a new ManagedClusterScalar CR `customized-resourceratiocpu` to trigger customized data collection.
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ManagedClusterScalar
+metadata:
+  name: customized-resourceratiocpu
+  namespace: cluster1
+```
+3. The agent on managed cluster watch the new ManagedClusterScalar CR and update score into it. The agent on managed cluster prerefresh the score periodically before expire time.
```
qiujian16, line 172: prerefresh? :)

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+  conditions:
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  expireTime: "2021-10-29T18:31:39Z"
+  score: 74
+```
+4. Placement read value from CR and score on it. (when to trigger re-schedule???)
```
qiujian16, line 189: different clusters update this value separately. So how can the scheduler to know it has sufficient data to schedule?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+5. Placement GC will clean the CR every 10 mins.
```
qiujian16, line 190: what is the criteria to GC. I think it also need to mention why this CR is created on demand and needs to be GCed.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+  - On each managed cluster, there would be a customized agent to sum up the cpu usage of the cluster. Then update the value into the ManagedClusterScalar CR with name `customized-resourceratiocpu`.
+  - On the hub, the placement prioritizer will read value from ManagedClusterScalar `customized-resourceratiocpu` and score based on it. 
+
+#### Story 2: The placement controller could prioritize clusters by metrics collected on hub.
+  - As an end user, can define prioritizer name `customized-resourceratiocpu` in placment spec.prioritizerPolicy.configurations to say he wants to schedule based on customized value resourceratiocpu.
+  - On the hub, there would be a customized agent to read cpu usage from thanos. It will generate a value for each cluster and update it into ManagedClusterScalar CR with name `customized-resourceratiocpu`.
+  - On the hub, the placement prioritizer will read value from ManagedClusterScalar `customized-resourceratiocpu` and score based on it.
+
+#### Story 3: The placement controller could prioritize clusters for disaster recovery.
```
qiujian16, line 34: This story needs to be rephrased a bit. I think briefly:

1. a user has two clusters, a primary and backup cluster on which workload storage is configured so storage are synced from primary cluster to backup.
2. user want to deploy workloads on the primary cluster.
3. when primary cluster is down, the workload should be automatically switched to the backup cluster.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  expireTime: "2021-10-29T18:31:39Z"
+  score: 0
+```
+4. On the hub, the placement prioritizer will select the cluster in the frist DC. Then disaster recovery workload will be running on it.
+5. When this cluster is unavailable (a taint is added), the customized agent will give cluster in the second DC the highest value.
```
qiujian16, line 258: I do not think it needs agent to update the score, right?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,321 @@
...
+  namespace: cluster1
+```
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ManagedClusterScalar
+metadata:
+  name: customized-dr
+  namespace: cluster2
+```
+3. On the hub, there would be a customized agent to give cluster in the frist DC the highest score.
```
qiujian16, line 224: customized agent or customized constroller?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,349 @@
...
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  score: 74
+```
+Usually each ManagedClusterScalar {PrioritizerName} will be created in more than one cluster namespaces, different clusters update the status separately. So the prioritizer plugin which depends on ManagedClusterScalar should follow below logic to get avaliable score and sort all the clusters.
+- PreScore. We have talked about PreScore() in previous section, the prioritizer plugin should create the CR before use it.
+- Score. The prioritizer plugin should watch all the created CRs until all of them have a score, then call Score() to sort clusters.
```
qiujian16, line 192: I think you would want to define a threshold, for instance 90% of scores exist

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Version Skew Strategy
+N/A
+
+## Drawbacks
+1. Is one score enough? any other missed use cases? For example, spreading depends on previous choice.
+
+## Alternative
+### Support user-defined scheduler
+As and end user, can specify the user-defined scheduler name in placement, and let a customized scheduler to generate placement decision. In this way, can also achieve extensiable scheduling.
```
qiujian16, line 343: extensible scheduling can also be achieved.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,349 @@
...
+
+Pros: 
+  - More flexible, the customized scheduler can schedule based on resource usage, SLA, metrics, etc.
+
+Cons:
+  - More effort to write a customized scheduler.
```
qiujian16, line 349: the customized scheduler needs to understand all the spec of placement api.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,349 @@
...
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  score: 74
+```
+Usually each ManagedClusterScalar {PrioritizerName} will be created in more than one cluster namespaces, different clusters update the status separately. So the prioritizer plugin which depends on ManagedClusterScalar should follow below logic to get avaliable score and sort all the clusters.
+- PreScore. We have talked about PreScore() in previous section, the prioritizer plugin should create the CR before use it.
+- Score. The prioritizer plugin should watch all the created CRs until all of them have a score, then call Score() to sort clusters.
+- Re-Schedule. Rather than watch ManagedClusterScalar and trigger Score(), that will cause frequently reschedule. The prioritizer plugin should periodically(eg. every 5 minutes) call Score() to reschedule.
+
```
qiujian16, line 194: I think we should also define how this scalar values are formalized.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,362 @@
...
+	// +optional
+	Status ManagedClusterScalarStatus `json:"status,omitempty"`
+}
+
+// ManagedClusterScalarSpec defines the attributes of the ManagedClusterScalar.
+type ManagedClusterScalarSpec struct {
+	// PrioritizerName will be the prioritizer name used in placement.
+	// +kubebuilder:validation:Required
+	// +required
+	PrioritizerName string `json:"prioritizerName,omitempty"`
```
elgnay, line 85: Is it possible that a ManagedClusterScalar is used by two prioritizers?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,362 @@
...
+  prioritizerName: {PrioritizerNameInPlacement}
+status:
+  conditions:
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  score: 74
```
elgnay, line 188: score -> scalar?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,362 @@
...
+  prioritizerName: CustomizeResourceRatioCPU
+status:
+  conditions:
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  score: 74
```
elgnay, line 251: score -> scalar

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,362 @@
...
+spec:
+  prioritizerName: CustomizeDisasterRecovery
+status:
+  conditions:
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  score: 100
```
elgnay, line 306: score -> scalar

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,362 @@
...
+spec:
+  prioritizerName: CustomizeDisasterRecovery
+status:
+  conditions:
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: ManagedClusterScalar updated successfully
+    reason: ManagedClusterScalarUpdated
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  score: 0
```
elgnay, line 323: score -> scalar

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,362 @@
...
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  score: 74
+```
+Usually each ManagedClusterScalar CR will be created in more than one cluster namespace, different clusters update the status separately. So the prioritizer plugin which depends on ManagedClusterScalar should follow below logic to get avaliable score and sort all the clusters.
+- Prescore
+
+In prescore stage, prioritizer plugin should create the CR before use it (have talked about this in previous section).
+The plugin should also check if 80% of the CRs have a valid score (80% is a reasonable threshold hardcode so far, might make it configurable in the future). If yes, the plugin will continue to score the clusters. If not, terminate this schduling and let requeue to trigger a reschedule.
```
elgnay, line 194: Do we still have a requeue logic here? or we just need periodical rescheduling?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: yes, in the current design a ManagedClusterScalar can be used by two prioritizers.

haoqing0110: Maybe not, the prioritizers plugin should check if the CR spec.PrioritizerName equals the prioritizers plugin name.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: fixed above 4 typo.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Update the doc.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+  - As an end user, I can configure placement yaml to use this scalar value to select clusters.
+
+#### Story 3: Disaster recovery workload could be automatically switched to an avaliable cluster.
+  - A user has two clusters, a primary and backup cluster on which workload storage is configured so storage are synced from primary cluster to backup.
+  - As an end user, I want to deploy workloads on the primary cluster first. And when primary cluster is down, the workload should be automatically switched to the backup cluster.
+
+#### Story 4: Users could use the data provided by third party controllers to select clusters.
+  - A third party controller can comprehensively evaluate cluster on latency, region, iops etc. and rate cluster according to the SLA the cluster can provide.
+  - As an end user, I can configure placement yaml to use this scalar value to select clusters.
+
```
suigh, line 49: For the story, maybe `sort clusters` is better than `select clusters`, as for the current OCM usage, there will some clusters in a placement, what we will do is to sort these clsuters with reported values, and end users can select the top one or top N from the placementdecision. For `select clusters`, the clusters in the placementdecision need to be updated (add/delete) with the reported values, the basic behavior is changed, and the performance may be impacted.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+	Score(ctx context.Context, placement *clusterapiv1alpha1.Placement, clusters []*clusterapiv1.ManagedCluster) (map[string]int64, error)
+}
+```
+
+- Delete
+
+The deletion of ManagedClusterScalar CR is handled by the framework (out of all the prioritizer plugins). The ManagedClusterScalar CR is created by each prioritizer plugin on demand, and the CR should be shared by some placements cross namespaces. When the ManagedClusterScalar CR is not used by any placement, need to clean it to avoid unnecessary api call.
+
+There will be a GC() funcion in the framework, being called every 10 minutes. Go through all the Placment CRs and ManagedClusterScalar CRs, and clean not used ManagedClusterScalar CRs.
+
```
suigh, line 166: we should also notify agent and controller about this, so that they will not report more data.

suigh: Is there a timeout for the CR? E.g. for the cpu usage, if we cannot get new data anymore, how can we handle this case?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+#### An agent update ResourceRatioCPU scalar to API, and placement prioritizer plugin score based on it. (Use Story 1&2)
+1. User create a new placement as below
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: placement
+  namespace: ns1
+spec:
+  numberOfClusters: 1
```
suigh, line 219: the `numberOfClusters` should be 2 or bigger, and we will provide sorted clusters, then user can select one (usually the top one).

If the `numberOfClusters` is bigger than 1, then user can do something else with the result, E.g. get the busiest cluster and check it.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+#### Disaster recovery workload could be automatically switched to an avaliable cluster. (Use Story 3)
+1. A user has 2 clusters, one is primary and one is backup cluster. To deploy workloads on the primary cluster, define the placement as below.  
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: placement
+  namespace: ns1
+spec:
+  numberOfClusters: 1
```
suigh, line 266: same suggetsion as above, the value of `numberOfClusters` should be bigger than 1

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: > we should also notify agent and controller about this, so that they will not report more data.

When unused ManagedClusterScalar CR is removed by GC(), the agent won't watch and update scalar to status anymore.

> Is there a timeout for the CR? E.g. for the cpu usage, if we cannot get new data anymore, how can we handle this case?

Yes, the API has `ValidUntil` designed in CR status. 
```
//ManagedClusterScalarStatus represents the current status of ManagedClusterScalar.
type ManagedClusterScalarStatus struct {
	// Conditions contain the different condition statuses for this managed cluster scalar.
	// +optional
	Conditions []metav1.Condition `json:"conditions"`
	// Scalar contains a scalable value of this managed cluster.
	// +optional
	Scalar int64 `json:"scalar,omitempty"`
	// ValidUntil defines the time this scalar is valid.
	// After this time, the scalar is considered to be invalid by placement. nil means never expire.
	// The controller ownning this resource should keep the scalar up-to-date.
	// +optional
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Type=string
	// +kubebuilder:validation:Format=date-time
	ValidUntil *metav1.Time `json:"validUntil" protobuf:"bytes,4,opt,name=lastTransitionTime"`
}
```

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: This case is different from the above, as it's for the disaster recovery case and should always return only one cluster.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Do you mean even the desired cluster number is 1 or any other number, the user still wants a list of sort clusters (more than 1) from placement decision, and he picks up one by himself?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: @qiujian16 @elgnay any comments for this?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
suigh: User will not set `numberOfClusters` for most cases, it is a limitation to the user.

For example, there are total 100 managed clusters, and userA owns 5 of them, what this user want to do is to build a placement with these 5 managed clusters, and sort these clusters by some rules, then select the top one to send workload. This user also need to get the last one in the sorted list to check if there is something wrong with that cluster or do something else.  With `numberOfClusters=1`, the user can only see the status of one cluster, and other clusters will be ignored, this is why I say this is a limitation.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+So we want a more extensible way to support scheduling based on customized value.
+
+### Goals
+- Design a new API(CRD) to contain a scalar value for each managed cluster.
+- Let placement prioritizer maintain the lifecycle (creating/deleting) of the CR.
+- Let placement prioritizer support rating clusters with the scalar value of the CR.
+
+### Non-Goals
+- How a customized agent update the scalar value into the CR.
+- Placement filter will not support the new API(CRD).
```
deads2k, line 28: Why not?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+### ManagedClusterScalar API
+ManagedClusterScalar is the new API to add, to represents a scalable value (aka score) of one managed cluster.
+```go
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Namespaced"
+// +kubebuilder:subresource:status
+
+// ManagedClusterScalar represents a scalable value (aka score) of one managed cluster.
+// Each ManagedClusterScalar only represents the scalar value of one prioritizer.
```
deads2k, line 64: one prioritizer per entry is definitely the cleanest, but it seems like we're likely to want a single controller in the managedcluster reporting back several prioritizer values.  Have we considered allowing multiple scalars reported back to amortize the cost of updates with many prioritizers from a single source?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+
+//ManagedClusterScalarStatus represents the current status of ManagedClusterScalar.
+type ManagedClusterScalarStatus struct {
+	// Conditions contain the different condition statuses for this managed cluster scalar.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// Scalar contains a scalable value of this managed cluster.
+	// +optional
+	Scalar int64 `json:"scalar,omitempty"`
```
deads2k, line 97: to be weighted, we wanted to force a range, rigth?  -100/+100 is what I recall.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+
+	// Items is a list of managed clusters
+	Items []ManagedClusterScalar `json:"items"`
+}
+```
+
+### Let placement prioritizer maintain the lifecycle of the CR.
+- Create
+
+The creation of ManagedClusterScalar CR is handled by each prioritizer plugin on demand. As every update in ManagedClusterScalar CR will introduce an api call, that will bring potential performance presure to api server. So we only want to create ManagedClusterScalar CRs for enabled prioritizer in filterd cluster namespace.
```
deads2k, line 127: nit: line break on new sentences so it's easier to comment.

deads2k: Given that these prioritizers are for selecting clusters, how would you know ahead of time which clusters you want to create them in?

You have to handle the case where the data isn't present for the placement decision, but I'm not sure how you would make this choice.

deads2k: If the managedcluster has the controller present which is trying to write the `ManagedClusterScalar`, that seems like enough of a signal to collect the data.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+```go
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Namespaced"
+// +kubebuilder:subresource:status
+
+// ManagedClusterScalar represents a scalable value (aka score) of one managed cluster.
+// Each ManagedClusterScalar only represents the scalar value of one prioritizer.
+// ManagedClusterScalar is a namesapce scoped resource. The namespace of the resource is the cluster namespace.
+type ManagedClusterScalar struct {
```
deads2k, line 66: `AddOnPlacementScore` ?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+  name: {PrioritizerName}
+  namespace: cluster1
+spec:
+  prioritizerName: {PrioritizerNameInPlacement}
+```
+The name is the prioritizer name.
+The namespace is the filtered cluster namespace.
+The spec.prioritizerName is the prioritizer name used in placement yaml. 
+
+The code change includes add PreScore() in Prioritizer interface, in which to create ManagedClusterScalar CR before Score().
```
deads2k, line 144: If you treated the presence of the addon prioritizer plugin in the managedcluster as sufficient to write data back, then PreScore is unnecessary.  Also, Pre versus Post versus During isn't really material since the status is reported back asynchronously, right?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+    status: "True"
+    type: ManagedClusterScalarUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  scalar: 74
+```
+Usually each ManagedClusterScalar CR will be created in more than one cluster namespace, different clusters update the status separately. So the prioritizer plugin which depends on ManagedClusterScalar should follow below logic to get avaliable scalar and sort all the clusters.
+- Prescore
+
+In prescore stage, prioritizer plugin should create the CR before use it (have talked about this in previous section).
+The plugin should also check if 80% of the CRs have a valid scalar (80% is a reasonable threshold hardcode so far, might make it configurable in the future). If yes, the plugin will continue to score the clusters. If not, terminate this scheduling and wait for the next reschedule.
```
deads2k, line 195: is it really better to place nothing instead of placing what you can?

It seems like you have two cases
1. a placement contains a managedcluster that no longer has an addonPlacementScore.
2. a placement needs to select additional managedclusters and the only ones available have no addonPlacementScore

It seems like always collecting the data if the managedcluster has the addon installed will reduce the mental load around, "what do we do if we don't have this information".

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,363 @@
...
+
+- Score
+
+In score stage, prioritizer plugin get valid scalar from ManagedClusterScalar CR. If no scalar updated in ManagedClusterScalar CR status, treat it as 0.
+After getting all the cluster scalar, the plugin should formalized them from -100 to 100. The maximal will be 100 and minimal is -100, other score distribute in proportion.
+
+- Re-Schedule
+
+Rather than watch ManagedClusterScalar and trigger Score(), that will cause frequently reschedule. The scheduler framework should periodically (eg. every 5 minutes) reschedule the placement.
```
deads2k, line 204: why would we give this special treatment instead of using our current behavior.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: We don't support `sort clusters` in placement decision is because it will trigger a frequent update. For example, when cluster score (high to low) changes from "cluster1, cluster2, cluster3" to "cluster3, cluster2, cluster1", `sort clusters` will trigger an update, while if use `select clusters`, it keeps "cluster1, cluster2, cluster3".

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: As we don't provide a sorted cluster list in the placement decision for now, to handle the above case, a possible solution is to create 2 placements, one sets prioritizer [weight](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/types.go#L321) to a positive number and selects the top N, another sets prioritizer weight to a negative number and select the top1. 

Can this satisfy your requirement?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: yes, will try to modify this part. 

It might be like letting the API contains a group of prioritizer values. So that on the managed cluster, one controller updates cpu/mem/cpuratio/memratio values into one CR, another controller updates other customized values, eg, GPU, into another CR.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
suigh: Only show the top 1 cluster has the same issue, all the clusters need to be sorted when the cluster scores are changed.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: -100/+100 indeed makes the user easy to understand the value, while the idea in this proposal is not to set the scalar range (placement-controller prioritizer plugin will range the final score to -100/+100)

The reason is, if the controller is running on the managed cluster, it doesn't have a global view to normalize the scalar into -100/+100. Or needs to set an appropriate default min/max value to normalize the scalar, the side effect is it will lose resolution. 

What will you suggest about this?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Will modify this part to make the logic clear.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Hmm, yes. Let me update this part, we may only want PreScore to do some check before Score, not to handle the ManagedClusterScalar CR creation.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Will update the doc to handle the 2 cases.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: For example, the environment has 1000 clusters, suppose there will be 1000 ManagedClusterScalar CRs, when they update one by one, watching the changes will trigger 1000 times reschedule.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Let us think about this.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,426 @@
...
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// Type defines the type of the prioritizer.
+	// Type is either "Build-in", "AddOnScore" or "", where "" is "Build-in" by default.
+	// When type is "Build-in", you can use the build-in prioritizers.
+	// When type is "AddOnScore", the prioritizers use score provided by AddOnPlacementScore to prioritize clusters, the
+	// score source should be defined in ScoreResourceName and ScoreName.
+	// +kubebuilder:default:=Build-in
+	// +optional
+	Type string `json:"type,omitempty"`
```
qiujian16, line 157:
```go
type PrioritizerCoordinate struct {
    // Type could be BuiltIn or AddonScore
    Type string

   // It could be Balance/Steady
   BuiltIn string

  AddonScore
}

type AddonScore struct {
  ResourceName string
  ScoreName string
}
```

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,426 @@
...
+- Reschedule
+  - In PreScore(), the prioritizer checks if 80% of the CRs have a valid score.
+    - If yes, the prioritizer will continue to score the clusters. 
+    - If no, remove the managed cluster which has invalid score from the placement decision.
+  - In Score(), the prioritizer will get customized score from `AddOnPlacementScore`, and calculate a final score for each managed cluster.
+
+Future work：there might be a cluster-scoped CRD to configure the threshold of each prioritizer. For example, the frequency to trigger a re-schedule, the threshold to treat the cluster's score as valid.
+
+### How to maintain the lifecycle of the AddOnPlacementScore CRs? 
+Even this is out of scope in this proposal, 2 options to maintain the resources:
```
qiujian16, line 292: 1 . where the 3rd partcy controller resides to maintain the score (both spoke and hub)
2. how to calculate the score. There are multiple options. For instance if it is a controller on hub to maintain the score, it can normalize based on max/min value it collects. We would consider a common library to make it more easy. 
3. When the score should be updated. ValidUntil must be consider for the 3rd party controller
4. When the score should be created? It can be create with the existence of a ManagedCluster, or on demand for the purpose of reducing objects on hub.

haoqing0110: updated.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: updated with the second option.

yue9944882:
@yue9944882 (subscribing the thread to so as to track it in the "mentioned" list)

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// PrioritizerScoreCoordinate represents the configuration of the prioritizer and score source.
+	// +kubebuilder:validation:Required
+	// +required
+	PrioritizerScoreCoordinate PrioritizerScoreCoordinate `json:"scoreCoordinate"`
+
+	// Weight defines the weight of the prioritizer score. The value must be ranged in [0,10].
+	// Each prioritizer will calculate an integer score of a cluster in the range of [-100, 100].
```
yue9944882, line 156: prioritizing the clusters by simply ordering by their weight/score will be leading to "frequent-order-flipping" issues IMHO. e.g. if the scores are calculated from the metrics by an external system, mostly the metrics value will be actively changing in a range over time even if everything runs stably/normally --- say we have a policy that selects 2 clusters out of 3, and all these clusters' score are ranging in a "narrow" range e.g. [50,51] over time, in this case the prioritized order of the 3 clusters will be easily flipping which in the end makes the higher systems (e.g. a multi-cluster workload) working on that policy receiving too frequent topology update from the placement decision. 
 
in addition to that, i can also think of a cyclic-consequence problem if the clusters are merely prioritized  by their rank --- e.g. again if we select a cluster from the candidate clusters by the scores, then the score of the selected cluster can be descending b/c of being selected (e.g. if the score is calculated from the allocated resource), then the next moment this cluster can be kicked from the list b/c a candidate w/ higher score is now winning. repeatedly the list of the selected cluster will also be frequently changing over time.

so i suppose we can probably introduce a  `"margin"` concept into the prioritzer. a margin is a buffer for preventing the selected cluster list to be frequently flipping. hence, overall the clusters are ordered by their score but as for the candidate clusters, they will have to win the selected cluster $margin more scores to be chosen by the prioritizer. if the scores are ranged in [-100,100] so a reasonable/recommended value of margin can be 20 i suppose.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+	// The final score of a cluster will be sum(weight * prioritizer_score).
+	// A higher weight indicates that the prioritizer weights more in the cluster selection,
+	// while 0 weight indicates that the prioritizer is disabled.
+	// +kubebuilder:validation:Minimum:=-10
+	// +kubebuilder:validation:Maximum:=10
+	// +kubebuilder:default:=1
```
yue9944882, line 162: IIUC it doesnt make sense to default the weight?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+	// +required
+	PrioritizerScoreCoordinate PrioritizerScoreCoordinate `json:"scoreCoordinate"`
+
+	// Weight defines the weight of the prioritizer score. The value must be ranged in [0,10].
+	// Each prioritizer will calculate an integer score of a cluster in the range of [-100, 100].
+	// The final score of a cluster will be sum(weight * prioritizer_score).
+	// A higher weight indicates that the prioritizer weights more in the cluster selection,
+	// while 0 weight indicates that the prioritizer is disabled.
+	// +kubebuilder:validation:Minimum:=-10
+	// +kubebuilder:validation:Maximum:=10
```
yue9944882, line 161: the kubebuilder mark doesnt match the go doc

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+	// +kubebuilder:validation:Required
+	// +required
+	PrioritizerScoreCoordinate PrioritizerScoreCoordinate `json:"scoreCoordinate"`
+
+	// Weight defines the weight of the prioritizer score. The value must be ranged in [0,10].
+	// Each prioritizer will calculate an integer score of a cluster in the range of [-100, 100].
+	// The final score of a cluster will be sum(weight * prioritizer_score).
+	// A higher weight indicates that the prioritizer weights more in the cluster selection,
+	// while 0 weight indicates that the prioritizer is disabled.
+	// +kubebuilder:validation:Minimum:=-10
```
yue9944882, line 160: a minus weight is confusing, i suppose the weight coefficient shoudl be [0, 1]?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: @yue9944882 yes, we also have similar concerns, what we have done so far is to introduce a prioritizer called "[Steady](https://github.com/open-cluster-management-io/placement/blob/main/pkg/plugins/steady/steady.go)",  try to make the decision steady when other score changes. And user can [adjust the Steady weight](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/15-resourcebasedscheduling#4-pin-the-placementdecisions-use-story-4) in placement. It looks like similar to what you think.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Want to use a minus weight to indicate "not suggest to select', or "reverse selection". For example, a prioritizer "MostAllocatableMemory" sorts clusters by allocatable memory from high to low, if user set weight to a negative number, he can select the cluster with least allocatable memory.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+
+So we want a more extensible way to support scheduling based on customized scores.
+
+### Goals
+- Design a new API(CRD) to contain the customized scores for each managed cluster.
+- Let placement prioritizer support rating clusters with the customized scores provided by the new API(CRD).
+
+### Non-Goals
+- How to maintain the lifecycle (create/update/delete) of the CRs.
+- Placement filters will not support the new API(CRD).
```
deads2k, line 27: the details from your answer https://github.com/open-cluster-management-io/enhancements/pull/33#discussion_r755819357 are very good.  Can you add them here:

> Put filter out of scope is because:
Most of the cases we know are that users want to use customized values to prioritize clusters.
For filters, adding labels to filter is easier than supporting customized values to filter.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
deads2k: The change to allow negative weighting is interesting.  Can you provide examples where being able to select both for and against an addonplacementscore is valuable?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+	// +optional
+	Configurations []PrioritizerConfig `json:"configurations,omitempty"`
+}
+
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// PrioritizerScoreCoordinate represents the configuration of the prioritizer and score source.
+	// +kubebuilder:validation:Required
+	// +required
+	PrioritizerScoreCoordinate PrioritizerScoreCoordinate `json:"scoreCoordinate"`
```
deads2k, line 153: this is a proposed change to `Name` in the existing API?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+
+The prioritizer will get customized score "cpuratio" from "default" `AddOnPlacementScore`, and calculate a final score for each managed clusters.
+
+- Reschedule
+
+The scores inside `AddOnPlacementScore` are supposed to be updated frequently. Rather than watch the changes in `AddOnPlacementScore` and trigger reschedule, the scheduler framework would periodically (eg. every 5 minutes) reschedule the placement.
+
+**What if no valid score is generated?**
+
+What if some of the scores are invalid? For example, the AddOnPlacementScore CR is not created for some managed clusters, the score is missing in some CR, the score is expired. To handle the invalid cases, we add a PreScore() in the Prioritizer interface, to do some pre-check before Score().
```
deads2k, line 274: Why is it desireable to effectively suspend a placement because a particular addonplacementscore isn't present?  This seems to produce a situation where a placement ceases to prioritize those things that it can in the absence of complete information.  Is there a discussion of the tradeoff and why this is good?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -0,0 +1,462 @@
...
+
+- When should the score be updated?
+
+  We recommend that you set `ValidUntil` when updating the score, so that the placement controller can know if the score is still valid in case it failed to update for a long time.
+
+  The score could be updated when your monitoring data changes, or at least you need to update it before it expires.
+
+- How to calculate the score.
+
+  The score should be in the range -100 to 100, you need to normalize the scores before update it into `AddOnPlacementScore`.
```
deads2k, line 324: must be in the range

deads2k:
The API break needs to be handled before an API PR can be merged.  I'm not really clear on why its desireable to suspend a placement entirely when some of its AddOnPlacementScores aren't available.  That seems like an unfortunate failure mode and an unfortunate bootstrap mode. 

I think we can reasonably expect that the AddOnPlacementScoring controller is created before the placement attempts to use it.  And if it isn't available, then scoring at zero (no priority) seems like a more robust failure mode.  The scoring controllers themselves could then choose a balance of "freshness" without fear that stale choices will cause the placement to stop placing.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Using https://github.com/open-cluster-management-io/enhancements/pull/33#discussion_r752325410 as an example, user owns 5 clusters, he wants to send the workload to the top one, and also wants to get the last one to check if there is something wrong with that cluster or do something else. With negative weight, he can achieve it by creating 2 placement, one with weight 1 to select the top one, and one with weight -1 to select the last one.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: OK, will do it.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Yes, as this is v1alpha1 API, might add `Name` back and claim it will be deprecated.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: The original consideration is AddOnPlacementScore CRs are created on-demand, so don't want to make decisions when most of the scores are not ready. 

As you mentioned, it's reasonably expect that the AddOnPlacementScoring controller is created before the placement attempts to use it. It makes sense to give a default score 0 if it isn't available.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: fixed.

haoqing0110:
@deads2k yes, thanks for your suggestion, and I update the proposal. For the API changes, claim the Name will be deprecated. For the unavailable scores, giving a default score of 0 makes sense for current use cases, will keep the logic simple until we know more use cases and feedback.

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
deads2k: > Using #33 (comment) as an example, user owns 5 clusters, he wants to send the workload to the top one, and also wants to get the last one to check if there is something wrong with that cluster or do something else. With negative weight, he can achieve it by creating 2 placement, one with weight 1 to select the top one, and one with weight -1 to select the last one.

can you include this in the doc itself for future us?

deads2k:
The behavior looks good to me.  The API change is significant, but if @qiujian16 is ok with it, I think it's ok during alpha.

/assign @qiujian16 
/approve

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: @deads2k yes, add the description about negative weight in the proposal.

qiujian16:
/lgtm

thanks, I think it is a great extension.
I am ok with the api break change in alpha given we have not had many consumers on placement plugin yet.

PR #36, opened 2021-11-16
Running deployments of cluster-manager and klusterlet outside of Hub and ManagedCluster

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>

### DIFF
--- a/enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
+++ b/enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
@@ -1,0 +1,265 @@
+# Running deployments of cluster-manager and klusterlet outside of Hub and ManagedCluster
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+The proposed work would support running deployments of cluster-manager and klusterlet outside of Hub and ManagedCluster. 
+
+## Motivation
+
+Currently, the cluster manager and klusterlet should be deployed on the hub/managed cluster, it is mandatory to expose the hub cluster to the managed cluster even if the managed cluster resides in an insecure environment. 
+
+To address this, we propose to provide an option for running cluster-manager/klusterlet outside of the hub/managed cluster.
+
+This ability will bring some additional possible benefits:
+* We can leverage compute resources better by running several cluster-managers/klusterlets deployments on the same powerful cluster; it also enhances our management of all deployments.
+* The footprints of the hub/managed cluster can be reduced.
+* We can use this ability to run the klusterlet on the hub cluster, so we do not need to expose the hub cluster to the managed cluster, and accordingly, we expose the managed cluster to the hub.
+
+### Goals
+
+* Support deploy cluster-manager outside of the hub cluster.
+* Support deploy klusterlet outside of the managed cluster.
+
+### Non-Goals
+
+* Support deploy addons agents outside of the managed cluster.(TBD)
+* Authenticate managed cluster without using CSR(CertificateSigningRequest).
+
+## User Stories
+
+### Story1
+
+As a managed cluster administrator, I hope as few as possible workloads running in my cluster while the cluster can be managed by the hub.
+
+### Story2
+
+As an administrator of several Hubs, I want to run all deployments in one place for a better management experience,
+### Story3
+
+As a hub cluster administrator, on the premise that an untrusted cluster can be managed, I do not want the cluster to talk to the hub cluster directly. 
+
+### The Comparison Between "Detached mode" and "Push mode"
+
+They both can solve the same "untrusted cluster" issue in the network. 
+
+But the "push mode" will need listwatch on multiple spokes, which has been proven to be not scalable in kubefed since a single controller will consume a lot of memory.
+
+However, in the "detached mode", there is no such a "center controller" to "monitor" every spoke. Instead, it could even distribute the computation in multiple nodes. 
+
+## Proposal
+
+Introduce a mode of deploying, named "Detached" mode, which means deploying cluster-manager/klusterlet outside of the hub/managed cluster.
+
+### Architecture
+
+#### Architecture of Detached cluster-manager
+
+```
+        +---------------------------------------------------------+                                               
+        |  management-cluster                                     |                                               
+        |                                                         |                                               
+        |  +----------------+  +-------------+ +------------+     |                                               
+        |  |deployment:     |  | deployment: | |deployment: |     |                                               
+        |  |cluster-manager |  | registration| |placement   | etc.|                                               
+        |  |                |  |             | |            |     |                                               
+        |  +------|---------+  +------|------+ +------|-----+     |                                               
+        |         |                   |               |           |                                               
+        +---------|-------------------|---------------|-----------+                                               
+                  |                   |               |                                                           
+                  |--------------|--------------------+                                                           
+                                 |watch,create,update,delete...                                                   
+                                 |                                                                                
+        +------------------------v--------------------------------+                                               
+        |  hub-cluster                                            |       registrater       +-----------------+   
+        |                                                         <-------------------------> managed cluster |   
+        |  +-------+ +-----------+ +---------------+              |                         +-----------------+   
+        |  | CRDs  | |APIServices| |Configurations | etc.         |                                               
+        |  |       | |           | |               |              |                                               
+        |  +-------+ +-----------+ +---------------+              |                                               
+        +---------------------------------------------------------+
+```
+
+The **hub-cluster** is where hub is installed and where the cluster agent should register.
+ 
+The **management-cluster** is the cluster where deployments are actually running on. The deployments of several hub-clusters could run in one management-cluster and be divided by different namespaces.
+
+Note that cluster-manager is deploy on **management-cluster** but configed with a kubeconfig secret named ”external-hub-kubeconfig”. The secret is a cluster-admin kubeconfig of the **hub-cluster** and installed in the same namespace as other hub components.
+
+#### Architecture of Detached klusterlet
+
+```
+                      +--------------+                                                                            
+                      |  hub cluster |                                                                            
+                      +-------^------+                                                                            
+                              |                                                                                   
+                              |                                                                                   
+                              |                                                                                   
++-----------------------------|------------------------------------+                                              
+|  management-cluster                                              |                                              
+|                                                                  |                                              
+| +--------------+  +--------------------+  +---------------+      |                                              
+| | deployment:  |  | deployment:        |  | deployment:   |  etc.|                                              
+| | klusterlet   |  | registration-agent |  | work-agent    |      |                                              
+| +------|-------+  +--------------------+  +-------|-------+      |                                              
+|        |                    |                     |              |                                              
++--------|--------------------|---------------------|--------------+                                              
+         |                    |                     |                                                             
+         |                    |                     |                                                             
+         ---------------------|----------------------                                                             
+                              | watch,create,update,delete...                                                     
+                              |                                                                                   
++-----------------------------v------------------------------------+                                              
+|  managed-cluster                                                 |                                              
+|                                                                  |                                              
+|  +----------------------+   +----------------------+             |                                              
+|  | CRDs:                |   | Configurations:      |             |                                              
+|  | clusterClaim         |   | role,rolebinding,sa  |   etc.      |                                              
+|  | appliedmanifestwork  |   |                      |             |                                              
+|  +----------------------+   +----------------------+             |                                              
++------------------------------------------------------------------+  
+```
+
+- The managed-cluster is the cluster managed by the hub cluster, manifestwork will be deployed onto this cluster.
+- The management-cluster is the cluster where Klusterlet related deployments are actually running.
+
+The managed cluster will not connect to the hub cluster, all things done by the management cluster. The management cluster could be the hub cluster, it could be another third-party cluster as well.
+
+In Detached mode, the "klusterlet.spec.namespace" will be ignored, all agent components will be installed to the namespace with the same name as klusterlet. The other thing to note is that users are responsible for creating an `external-managed-kubeconfig` in the agent installation namespace, Klusterlet-operator will use this kubeconfig to get a new minimum permission kubeconfig for the managed cluster, then render it to registration-agent and work-agent.
+
+In addition, the addons agents are not considered to be deployed outside of the managed cluster currently. Here are some reasons:
+* In Detached mode, the deploying of addon-agent will be more complex: the addon-agents deployment runs on the management cluster; the role, service account, rolebindings runs on the managed cluster; maybe there will be some other things that need to be deployed and we don’t know where they should be deployed.
+* We are not sure that all addons agents pods can run on the management cluster, maybe some addons agents running on the management cluster will cause the function to fail.
+
+#### Architecture of a combined usage of both side
+
+```
+                    +-----------------------------------------------------------------+                                 
+                    | mnagement-cluster                                               |                                 
+                    |                                                                 |                                 
+                    | cluster-manager                         klusterlet              |                                 
+                    | +-------------------+                   +-------------------+   |                                 
+                    | | +-------------------+                 | +-------------------+ |                                 
+                    | | |                 | |<--------------->| |                 | | |                                 
+                    | +-|-----------------+ |                 +-|-----------------+ | |                                 
+                    |   +-------------------+                   +-------------------+ |                                 
+                    +-------|----------------------------------------------|----------+                                 
+                            |                                              |                                            
+                            |     <private cloud>  |-----------------------|----------------------
+                            |                      |  <public cloud>       |                                            
+                   watch,create,update,delete...   |         watch,create,update,delete...                               
+                            |                      |                       |                                            
+                    +-------v-----------+          |           +-----------v-----------+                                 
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/33-hosted-deploy-mode/hosted-deploy-mode.md
```diff
@@ -0,0 +1,217 @@
...
+### Architecture
+
+#### Architecture of Detached cluster-manager
+
+```
+        +---------------------------------------------------------+                                               
+        |  management-cluster                                     |                                               
+        |                                                         |                                               
+        |  +----------------+  +-------------+ +------------+     |                                               
+        |  |deployment:     |  | deployment: | |deployment: |     |
```
qiujian16, line 63: registration manager/placement manager I do no think we have work manager on hub

enhancements/sig-architecture/33-hosted-deploy-mode/hosted-deploy-mode.md
```diff
@@ -0,0 +1,217 @@
...
+        |  |       | |           | |               |              |                                               
+        |  +-------+ +-----------+ +---------------+              |                                               
+        +---------------------------------------------------------+
+```
+
+The **hub-cluster** is where hub is installed and where the cluster agent should register.
+ 
+The **management-cluster** is the cluster where deployments are actually running on. The deployments of several hub-clusters could run in one management-cluster and be divided by different namespaces.
+
+Note that cluster-manager is deploy on **management-cluster** but configed with a kubeconfig secret named ”external-hub-kubeconfig” which is a cluster-admin kubeconfig of the **hub-cluster**.
```
qiujian16, line 87: in which ns?

enhancements/sig-architecture/33-hosted-deploy-mode/hosted-deploy-mode.md
```diff
@@ -0,0 +1,217 @@
...
+* In Detached mode, the deploying of addon-agent will be more complex: the addon-agents deployment runs on the management cluster; the role, service account, rolebindings runs on the managed cluster; maybe there will be some other things that need to be deployed and we don’t know where they should be deployed.
+* We are not sure that all addons agents pods can run on the management cluster, maybe some addons agents running on the management cluster will cause the function to fail.
+
+#### Architecture of a combined usage of both side
+
+```
+                    +-----------------------------------------------------------------+                                 
+                    | mnagement-cluster                                               |                                 
+                    |                                                                 |                                 
+                    | deployments of hub                      deployments of agent    |
```
qiujian16, line 140: let's call it cluster-manager and klusterlet

enhancements/sig-architecture/33-hosted-deploy-mode/metadata.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - "@xuezhaojun"
+  - "@zhujian7"
+reviewers:
+  - "@qiujian16" 
+approvers:
+  - "@qiujian16"
+creation-date: 2021-11-16
+last-updated: 2021-11-16
+status: provisional
+see-also:
```
qiujian16, line 12: no need for this section

enhancements/sig-architecture/33-hosted-deploy-mode/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+title: hosted-deploy-mode
+authors:
+  - "@xuezhaojun"
+  - "@zhujian7"
+reviewers:
+  - "@qiujian16" 
+approvers:
+  - "@qiujian16"
```
qiujian16, line 8: add @zhiweiyin318 and @elgnay as reviewers also.

qiujian16:
I think it is clear to me
/approve

/assign @zhiweiyin318 
/assign @elgnay

clyang82:
Could you list a real use case for this proposal? for example, I have a hub cluster is installed in private network while the managed cluster is in public cloud. right now, the managed cluster cannot connect to the hub cluster. so with this enhancement, how can I deploy the management cluster to achieve my goal? Thanks.

enhancements/sig-architecture/33-hosted-deploy-mode/hosted-deploy-mode.md
xuezhaojun: Done. Replace work with placement

enhancements/sig-architecture/33-hosted-deploy-mode/hosted-deploy-mode.md
xuezhaojun: Updated.

xuezhaojun:
> Could you list a real use case for this proposal? for example, I have a hub cluster is installed in private network while the managed cluster is in public cloud. right now, the managed cluster cannot connect to the hub cluster. so with this enhancement, how can I deploy the management cluster to achieve my goal? Thanks.

Thanks for the review, in this case, I think the proper approach is to use the hub-cluster as "management-cluster" and run klusterlet and deployments on it.
But I think we're not there yet, so far we have done the job that separate CRDs and deployments, therefore deployments of klusterlet and cluster-manager could run in the same cluster. Still, the managed cluster needs to send csr to the hub cluster.

Please correct me if I got something wrong. @zhujian7 @qiujian16  And Do we need to have a discussion about how to make deployments communicate directly?

qiujian16:
@xuezhaojun I think we should describe this use case in the doc. I do not think csr matters here, because csr in this case is created on hub.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,217 @@
...
+        |  |       | |           | |               |              |                                               
+        |  +-------+ +-----------+ +---------------+              |                                               
+        +---------------------------------------------------------+
+```
+
+The **hub-cluster** is where hub is installed and where the cluster agent should register.
+ 
+The **management-cluster** is the cluster where deployments are actually running on. The deployments of several hub-clusters could run in one management-cluster and be divided by different namespaces.
+
+Note that cluster-manager is deploy on **management-cluster** but configed with a kubeconfig secret (in the `open-cluster-management` namespace) named ”<cluster-manager-name>-hub-kubeconfig” which is a cluster-admin kubeconfig of the **hub-cluster**.
```
zhujian7, line 87: `<cluster-manager-name>-hub-kubeconfig` means for the external hub kubeconfig? Please confirm it is always deployed in the `open-cluster-management`?

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,217 @@
...
+|  +----------------------+   +----------------------+             |                                              
++------------------------------------------------------------------+  
+```
+
+- The managed-cluster is the cluster managed by the hub cluster, manifestwork will be deployed onto this cluster.
+- The management-cluster is the cluster where Klusterlet related deployments are actually running.
+
+The managed cluster will not connect to the hub cluster, all things done by the management cluster. The management cluster could be the hub cluster, it could be another third-party cluster as well.
+
+In Detached mode, the "klusterlet.spec.namespace" will be ignored, all agent components will be installed to the namespace named <klusterlet's name>-open-cluster-management-agent. The other thing to note is that users are responsible for creating an external-managed-kubeconfig in the agent installation namespace, Klusterlet-operator will use this kubeconfig to get a new minimum permission kubeconfig for the managed cluster, then render it to registration-agent and work-agent.
```
zhujian7, line 128:
```suggestion
In Detached mode, the "klusterlet.spec.namespace" will be ignored, all agent components will be installed to the namespace with the same name as klusterlet. The other thing to note is that users are responsible for creating an `external-managed-kubeconfig` in the agent installation namespace, Klusterlet-operator will use this kubeconfig to get a new minimum permission kubeconfig for the managed cluster, then render it to registration-agent and work-agent.
```

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: Yes, `<cluster-manager-name>-hub-kubeconfig` means the external hub kubeconfig.  And the current approach is all of cluster-manager's external-hub-kubeconfig will be depolyed in the `open-cluster-management` namespace.

zhujian7: Users create `<cluster-manager-name>-hub-kubeconfig` in `open-cluster-management` ns, then the operator copy it to the installation ns?

xuezhaojun: It could be: operator use this <cluster-manager-name>-hub-kubeconfig to generate proper SA-kubeconfig and install them in respective installation ns.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: The commit is squshed to and sign information.

xuezhaojun:
> have a hub cluster is installed in private network while the managed cluster is in public cloud. right now, the managed cluster cannot connect to the hub cluster.

The use case is added.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,223 @@
...
+                    +-------v-----------+                     +------------v----------+                                 
+                    |hub                |                     |managed cluster        |                                 
+                    |                   |     logically       |                       |                                 
+                    | +--------+        |<------------------->| +--------+            |                                 
+                    | | CRDs...|        |                     | |CRDs... |            |                                 
+                    | +--------+        |                     | +--------+            |                                 
+                    +-------------------+                     +-----------------------+     
+```
+
+A real use case would be: The hub cluster is installed in **private network** while the managed cluster is in public cloud. right now, the managed cluster cannot connect to the hub cluster. **It also means deployments on managed cluster can not apply any resources to the hub cluster directly**.
```
xuezhaojun, line 160: @clyang82  Hello, Chunlin, we have added a real use case here, please have a review and see whether it's clear for you. Thanks~

xuezhaojun: @qiujian16 The use case is added now, please also have a review, thanks~

xuezhaojun: Updated this part, the link has outdated.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,224 @@
...
+                    +-------v-----------+          |           +-----------v-----------+                                 
+                    |hub                |          |           |managed cluster        |                                 
+                    |                   |     logic|ally       |                       |                                 
+                    | +--------+        |<- - - - -|- - - - - >| +--------+            |                                 
+                    | | CRDs...|        |          |           | |CRDs... |            |                                 
+                    | +--------+        |          |           | +--------+            |                                 
+                    +-------------------+          |           +-----------------------+     
+```
+
+A real use case would be: The hub cluster is installed in a **private network** while the managed cluster is in a public cloud. right now, the managed cluster cannot connect to the hub cluster. **It also means deployments on managed cluster can not apply any resources to the hub cluster directly**.
```
xuezhaojun, line 161: @clyang82 @qiujian16   Hello, we have added a real use case here, please have a review and see whether it's clear for you. Thanks~

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: Solved, apply secret after namespace is created

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
ryanzhang-oss: quick question, in the above case. Is there any reason that the cluster-manager controllers need to be in the management cluster?

ryanzhang-oss:
I wonder if there an implementation PR for this proposal?

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,224 @@
...
+## Motivation
+
+Currently, the cluster manager and klusterlet should be deployed on the hub/managed cluster, it is mandatory to expose the hub cluster to the managed cluster even if the managed cluster resides in an insecure environment. 
+
+To address this, we propose to provide an option for running cluster-manager/klusterlet outside of the hub/managed cluster.
+
+This ability will bring some additional possible benefits:
+* We can leverage compute resources better by running several cluster-managers/klusterlets deployments on the same powerful cluster; it also enhances our management of all deployments.
+* The footprints of the hub/managed cluster can be reduced.
+* We can use this ability to run the klusterlet on the hub cluster, so we do not need to expose the hub cluster to the managed cluster, and accordingly, we expose the managed cluster to the hub (A bit like Pull Mode to Push Mode).
```
ryanzhang-oss, line 24: I wonder if this approach makes the management cluster a performance bottleneck? What is the advantage of running Klusterlets in a management cluster vs a pure push model?

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,224 @@
...
+
+As a managed cluster administrator, I hope as few as possible workloads running in my cluster while the cluster can be managed by the hub.
+
+### Story2
+
+As a hub cluster administrator, on the premise that an untrusted cluster can be managed, I do not want the cluster to talk to the hub cluster directly.
+
+### Story3
+
+As an administrator of several Hubs, I want to run all deployments in one place for a better management experience,
```
ryanzhang-oss, line 48: IIRC, this is a fleet of a fleet case. Would it make sense to create a uber hub cluster to manage all the hub clusters as member clusters instead?

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
```diff
@@ -0,0 +1,224 @@
...
+
+## User Stories
+
+### Story1
+
+As a managed cluster administrator, I hope as few as possible workloads running in my cluster while the cluster can be managed by the hub.
+
+### Story2
+
+As a hub cluster administrator, on the premise that an untrusted cluster can be managed, I do not want the cluster to talk to the hub cluster directly.
```
ryanzhang-oss, line 44: This seems to be a call for a push model

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: That's a good point! The controller could be running in the hub cluster too in the above case.
In practice, I got a case that the management cluster is super powerful, and the customer wants to leverage it for serval hub clusters, therefore hub clusters could be very light-weigh.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: Yes, that's right. It could be solved by a push model as well.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: We had a case that a customer has a super-powerful machine and want to use it to run every deployment of each hub cluster. So base on this situation, the management cluster would be a bottleneck, but if the situation changes, we might need to use another approach. 
For the advantage, I would say the compatibility since there are no API changes. @zhujian7  Feel free to add more.
And CSR is still needed for this.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: In this case, we aim to do no management. Instead, for example,  we can easily collect logs, metrics, and other data of running pods from all hubs.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
qiujian16: @xuezhaojun I think it worth writing a section to compare it with push mode (cons and pros). I think a push mode will need listwatch on multiple spokes, which has been proven to be not scalable in kubefed since a single controller will consume a lot of memory. Running klusterlet separated distribute the computation to multiple pods. It might be good to consider even in push case, we still have workers for each different spoke.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
qiujian16: this is interesting. We had discussion before on such uber hub idea. The thing that is not clear is what scenario want to achieve with this uber hub. Do we just want to provide a index so the request can be proxied to a certain second level hub or we want a higher level coordination, or just checking the status of the clusters. I think scalability also needs to be considered with this uber hub. a single can support manage than 2k cluster today. A uber hub might need to managed 10/100 times more.

enhancements/sig-architecture/33-detached-deploy-mode/detached-deploy-mode.md
xuezhaojun: The content added.

ryanzhang-oss: good point, the push model shouldn't require to run all of them in the same pod.

PR #37, opened 2021-11-30
proposal for clusterset override

ldpliu:
https://github.com/open-cluster-management/backlog/issues/18304
Signed-off-by: ldpliu <daliu@redhat.com>

### DIFF
--- a/enhancements/sig-architecture/30-clusterset-override/README.md
+++ b/enhancements/sig-architecture/30-clusterset-override/README.md
@@ -1,0 +1,400 @@
+# ManagedClusterSet Override
+
+## Release Signoff Checklist
+ 
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+ 
+## Summary
+The proposed work enhances the managedClusterSet API to support managedClusterSet Override.
+
+## Motivation
+Today, managedClusterSet is designed to group the managedClusters, a managedCluster can only be included in one managedClusterSet, by setting a label `cluster.open-cluster-management.io/clusterset` on the managedCluster.
+A user must have explicit RBAC rules, `managedclustersets/join`, to add such labels.
+There are also raising requirements on using managedClusterSet for the following purposes. Within a purpose, the managedClusterSet must be exclusive, but for different purposes it can overlap.
+- As cluster admin, I want to create a managedClusterSet which includes all clusters in the apac region. Then I can apply certain configurations to all clusters in apac.
+- As a SRE team member, I want to use a managedClusterSet for Disaster Recovery purposes. The managedClusterSet should select two or more clusters in different regions. 
+- As a DEV team member, I want to use a managedClusterSet to select clusters based on the middleware enabled on these clusters, so I could run special applications in these clusters.
+
+## Proposal
+
+### Changes on managedClusterSet APIs.
+
+Currently, managedClusterSets use the label `cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>` to select managedClusters. 
+But for some use cases, I want to use other labels to select managedClusters in a managedClusterSets.
+
+So, In this proposal, we change the managedClusterSets spec and want to provide a flexible way to select managedClusters. 
+
+```go
+type ManagedClusterSetSpec struct {
+    // Selector represents a selector of ManagedClusters by labels and names.
+    ClusterSelector ManagedClusterSelector  `json:"clusterSelector"`
+}
+
+type ManagedClusterSelector struct{
+    // "" means to use the current mechanism of matching label <cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>.
+    // "LabelSelector" means to use the LabelSelector to select target managedClusters
+    // "ExclusiveLabel" means to use a particular cluster label. It is guaranteed that clustersets with same label key are exclusive with each others
+    // +optional
+    SelectorType SelectorType `json:"selectorType"`
+
+    // ExclusiveLabel defines one label which clusterset could use to select target managedClusters. In this way, we will:
+    // 1. Guarantee clustersets with same label key are exclusive
+    // 2. Enable additional permission check when cluster joining/leaving a clusterset (the label key should start with the reserved prefix "cluster.open-cluster-management.io/" and "info.open-cluster-management.io/");
+    ExclusiveLabel *ManagedClusterLabel `json:"exclusiveLabel"`
+
+    // LabelSelector define the general labelSelector which clusterset will use to select target managedClusters
+    LabelSelector *metav1.LabelSelector `json:"labelSelector"`
+}
+
+type SelectorType string
+
+const (
+    LabelSelector       SelectorType = "LabelSelector"
+    ExclusiveLabel        SelectorType = "ExclusiveLabel"
+)
+
+//ManagedClusterLabel defines one label
+type ManagedClusterLabel struct {
+    Key string `json:"key"`
+    Value string `json:"value"`
+}
+```
+
+### RBAC
+Currently, When I want to create a managedClusterSet, I only should have `create` permission to the managedClusterSet. 
+In the managedCluster part, label `cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>` means the cluster is in a managedClusterSet.
+If I want to add the label `cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>` to a managedCluster, I must have `create` permission for subresource `managedclustersets/join`. 
+If I want to add/update other labels, I just need to have `update` permission to the managedCluster resource.
+
+In this proposal, managedClusterSet could use any labels to select managedClusters. And if I add a label for a managedCluster, it may lead to the managedCluster added to a managedClusterSet. So we need to rethink about the rbac control of managedClusterSet
+
+We will use the following questions to consider the RBAC control.
+
+#### What purposes will managedClusterSet may be used for
+a. Create a managedClusterSet for disaster recovery purposes, which should include specific clusters.
+b. Create a managedClusterSet which should include all clusters in one region or cloud provider.
+c. Create a managedClusterSet which has the same capacity(like: all clusters enabled middleware).
+d. Create a managedClusterSet for each squad for resource group purposes.
+
+#### Who can create a managedClusterSet
+Same as the current way, Anyone who has `create` permission to `managedclustersets` could create managedClusterSet. 
+But generally, Cluster admin should not give this permission to others. So only cluster admin could create the managedClusterSet.
+
+#### What kinds of roles exist related to managedClusterSets and managedClusters
+In General, there are three kinds of roles related to managedClusterSet.
+1. Cluster admin
+   - Cluster admin could `create`, `get`, `list`, `update`, `delete` all managedClusterSets.
+   - Cluster admin could `bind` all managedClusterSets to all namespaces.
+   - Cluster admin could `create`, `get`, `list`, `update`, `delete` all managedClusters.
+   - Only Cluster admin could `accept` managedClusters.
+
+2. Controller/agent which run in managedClusters
+   - The Controller/agent should not `create`, `get`, `list`, `update`, `delete` managedClusterSets.
+   - The Controller/agent should not `bind` managedClusterSet.
+   - The Controller/agent only should `get` related managedCluster and `update` the managedCluster status and labels.
+
+3. Non Cluster admin users
+   - These users should `get` some managedClusterSets, and could not `create`, `update`, `delete`, `list` managedClusterSets.
+   - These users should `bind` some managedClusterSets to their namespaces.
+   - These users may need to `create` managedClusters(For Resource Group Purpose managedClusterSet[d], in each group, the group members may need to create the managedClusters). They could also `delete` these managedClusters, and `update` these managedClusters labels.
+
+#### Who can add labels for managedClusters
+1. Cluster admin
+Cluster admin could add any labels to any managedClusters.
+
+2. Controller/agent which run in managedClusters
+Currently, the controller/agent may add labels like: `region: apac`, `cloud: Amazon`, `middlewareEnabled: true`. And these labels mean these clusters have some attributes.
+- `region: apac` means the cluster is provisioned in apac region
+- `cloud: Amazon` means the cluster is provisioned in Amazon cloud provider
+- `middlewareEnabled: true` means this cluster enables the middleware.
+
+3. Non Cluster admin users
+Currently, non cluster admin users may also `add/update` some managedClusters labels. 
+
+#### Add restrictions for adding/updating/deleting labels to managedClusters
+In this proposal, managedClusterSet could use any labels to select managedClusters. And the new added labels may lead to this managedClusters added to a managedClusterSet.
+
+But for some managedClusterSet(like: Disaster Recovery ManagedClusterSet[a], Resource Group ManagedClusterSet[d]), we do not want any users to add their managedClusters to these managedClusterSets. 
+
+So we propose adding restrictions for some labels, and if a user wants to add these kinds of labels, he/she must have a "special" permission.
+Then the managedClusterSets which do not want anyone to join could use these labels to select managedClusters.
+
+We propose adding restrictions for labels which key have prefix `cluster.open-cluster-management.io/` and `info.open-cluster-management.io/`
+- `cluster.open-cluster-management.io/` is encouraged for non cluster admin users. Like for "Resource Group ManagedClusterSet[d]", the DEV team users may add a label like: `cluster.open-cluster-management.io/clusterset: dev`
+- `info.open-cluster-management.io/` is encouraged for spoke agents, and the agents should have permissions on their own managedcluster to modify these kinds of labels.
+  The label key should be: `info.open-cluster-management.io/region`, `info.open-cluster-management.io/cloud`, `info.open-cluster-management.io/vendor`
+
+Then we need a way to define the "special" permission.
+Investigate the existing rbac rule in current k8s, we use a rule which is similar to [csr-approver](https://github.com/kubernetes/website/blob/3ca34b9a3be454055ba234f1d2ff7d55809b5040/content/zh/examples/access/certificate-signing-request/clusterrole-approve.yaml#L25)
+
+We define a subresource `managedclusters/label`, which resource name is the label of managedCluster. This subresource means user could add/remove the label to managedClusters.
+
+For example: 
+- This permission means the users could add/remove the label `cluster.open-cluster-management.io/clusterset:dev` to their managedClusters.
+
+```yaml
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclusters/label"]
+   resourceNames: ["cluster.open-cluster-management.io/clusterset:dev"]
+   verbs: ["create"]
+```
+
+- This permission means the users could add/remove any labels to their managedClusters.
+```yaml
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclusters/label"]
+   resourceNames: ["*"]
+   verbs: ["create"]
+```
+
+- This permission means the users could add/remove labels with `cluster.open-cluster-management.io/clusterset` key and arbitrary value to their managedClusters.
+```yaml
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclusters/label"]
+   resourceNames: ["cluster.open-cluster-management.io/clusterset:*"]
+   verbs: ["create"]
+```
+
+Notes: The `managedclustersets/join` permission will be deprecated after the proposal applied, and the existing `managedclustersets/join` permission could be replaced with the following permission.
+```yaml
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclusters/label"]
+   resourceNames: ["cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>"]
+   verbs: ["create"]
+```
+
+#### Who can bind the managedClusterSet to a namespace
+Same as the current way, anyone who has `managedclustersets/bind` permission could bind a managedClusterSet to a namespace.
+
+### Examples
+### Example1: As cluster admin, I want to create a managedClusterSet which includes all clusters in the apac region. Then I can apply certain configurations to all clusters in apac.
+1. For each cluster, there should be an agent running on the managedCluster. The agent should have the following permissions:
+- `update` permission to its related managedCluster
+- `create` permission to subresource `managedclusters/label` with resourceName `info.open-cluster-management.io/region:*`
+2. For each cluster in the apac region, related agents should add a label: `info.open-cluster-management.io/region:apac` to these managedClusters.
+3. As cluster admin, I could create a managedClusterSet to select all managedClusters in `apac` region
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: ManagedClusterSet
+metadata:
+  name: apacset
+spec:
+  clusterSelector:
+    selectorType: ExclusiveLabel
+    exclusiveLabel:
+      key: info.open-cluster-management.io/region
+      value: apac
+```
+
+### Example2: As a DEV team member, I want to use a managedClusterSet to select clusters based on the middleware enabled on these clusters, so I could run special applications in these clusters.
+1. For each cluster, there should be an agent running on the managedCluster. The agent should have the following permissions:
+- `update` permission to its related managedCluster
+- `create` permission to subresource `managedclusters/label` with resourceName `info.open-cluster-management.io/middlewareEnabled:true`
+2. For each cluster which enables the middleware, related agents should add a label: `info.open-cluster-management.io/middlewareEnabled:true` to these managedClusters.
+3. Cluster admin create a managedClusterSet `middlewareenabledset`
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: ManagedClusterSet
+metadata:
+  name: middlewareenabledset
+spec:
+  clusterSelector:
+    selectorType: LabelSelector
+    labelSelector:
+      matchLabels:
+        info.open-cluster-management.io/middlewareEnabled: true
+```
+4. Cluster admin should give following permissions to DEV team, so these team members could run applications in the managedClusterSet's clusters.
+- `get` permission to managedClusterSet `middlewareenabledset`
+- `create` permission to subresource `managedclustersets/bind` to managedClusterSet `middlewareenabledset`
+```yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+ name: clustersetrole
+rules:
+ - apiGroups: ["cluster.open-cluster-management.io"]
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,239 @@
...
+rules:
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclustersets/exclusiveKey"]
+   resourceNames: ["region"]
+   verbs: ["create"]
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclustersets"]
+   resourceNames: ["apac"]
+   verbs: ["*"]
+```
```
elgnay, line 91: I don't think you need to repeat the required RBAC rules for each example because they have already been covered in the above section.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,239 @@
...
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ManagedClusterSet
+metadata:
+name: middlewareenabled
+spec:
+exclusiveKey: middlewareAttribute
+```
+ 
+### Risks and Mitigation
+Currently, if I want to create a managedClusterSet, I must get permissions about two kinds of resources, `managedclustersets/exclusiveKey` and `managedclustersets`. When there are multiple resources for both `managedclustersets/exclusiveKey` and `managedclustersets`, it may cause permission leak.
```
elgnay, line 164: There is another risk. Since permission check will no longer be done when adding a cluster into a clusterset. User may add a cluster into a clusterset incidentally by adding a label to cluster, which is the exclusiveKey of that clusterset.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ManagedClusterSet
+metadata:
+name: aws
+spec:
+exclusiveKey: region
+```
+
+2. Currently, their is no permission check when adding a managedcluster to clusterset. User may add a cluster into a clusterset incidentally by adding a label to cluster, which is the exclusiveKey of that clusterset.
+
```
qiujian16, line 172: I think we need to discuss about mitigation approach here.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+### Test Plan
+- Unit tests will cover the functionality of the controllers.
+- Unit tests will cover the new APIs.
+- And e2e tests will cover the following cases.
+ - Create/update/delete managedClusterSet with permission `managedclustersets/exclusiveKey`
+ - Create/update/delete managedClusterSet without permission `managedclustersets/exclusiveKey`
+ - Update managedClusterSet spec exclusiveKey
+ 
+### Upgrade / Downgrade Strategy
+N/A
```
qiujian16, line 183: upgrade must be considered

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+ 
+## Proposal
+1. Changes on ManagedClusterSet APIs.
+ 
+Add a field into ManagedClusterSet spec
+ 
+```go
+type ManagedClusterSetSpec struct {
+ // ExclusiveKey defines the key of the label which the managedClusterSet will use to select managedCluster.
+ // The managedClusterSet will select ManagedClusters based on the managedCluster's label "<ExclusiveKey>:<ManagedClusterSetName>".
```
deads2k, line 30: Why is this more desireable than a general label selector?

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+ 
+## Summary
+The proposed work enhances the managedClusterSet API to support managedClusterSet Override.
+ 
+## Motivation
+Today, managedClusterSet is designed to be exclusive. A managedCluster can only be included in only one managedClusterSet, by setting a label `cluster.open-cluster-management.io/clusterset` on the managedCluster. A user must have explicit RBAC rules, `managedclustersets/join`, to add such labels.
```
deads2k, line 15: Can you add a section to the proposal that explains why the exclusivity was originally a design goal, why we are changing that behavior, and how we expect this to impact clients.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+ ExclusiveKey string `json:"exclusiveKey"`
+}
+```
+ 
+Notes:
+As managedClusterSet uses the label `<ExclusiveKey>:<ManagedClusterSetName>` to filter managedclusters. So if I want to use one label key as exclusiveKey, the label value must match the k8s name rule.
+ 
+2. RBAC change
+ 
+Given the flexibility introduced by exclusive key, instead of control whether a cluster can be added into a managedClusterSet with sub resource of `managedclustersets/join`, it might make more sense to control whether a user can create a managedClusterSet with a certain exclusive key. We can define a subresource of `managedclustersets/exclusiveKey`, to control the creation of certain managedClusterSet.
```
deads2k, line 43: Can you add a paragraph describing which actors are currently able to label ManagedClusters by default?

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+metadata:
+name: apac
+spec:
+exclusiveKey: region
+```
+
+This managedClusterSet includes all clusters with label `region: apac`, so I could use this managedClusterSet to select all clusters in apac, then apply a certain configuration to these clusters.
+ 
+### I want to create a managedClusterSet for Disaster Recovery purposes, the managedClusterSet should select two or more clusters in different regions.
+1. I need to manually select one cluster in each region: `apac`,`us`,`europe`, and set label: `clusterAttribute=backup` for these clusters
```
deads2k, line 89: This connection seems to imply that the client setting the label value on the managedclusters is also the client creating the manageclusterset.  General label selection tries to decouple the client setting the label values (the property or attributor provider) from the client consuming the properties for a purpose (the managedclusterset creator in this example).

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,191 @@
...
+1. Changes on ManagedClusterSet APIs.
+ 
+Add a field into ManagedClusterSet spec
+ 
+```go
+type ManagedClusterSetSpec struct {
+ // ExclusiveKey defines the key of the label which the managedClusterSet will use to select managedCluster.
+ // The managedClusterSet will select ManagedClusters based on the managedCluster's label "<ExclusiveKey>:<ManagedClusterSetName>".
+ // If the exclusiveKey is not set, we will use `cluster.open-cluster-management.io/clusterset` as exclusiveKey.
+ // The value can not change once the managedClusterSet is created.
```
yue9944882, line 32: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1101-immutable-fields

the feature can help us to make this field immutable

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata: 
+  name: clusterset-validation
+  namespace: open-cluster-management-hub
+data:
+  region: apac
+  region: us
```
qiujian16, line 218: I do not think you can create a comfigmap like this.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,284 @@
...
+metadata:
+ name: clustersetrole
+rules:
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclustersets"]
+   resourceNames: ["apacSet"]
+   verbs: ["create"]
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclustersets/clusterlabel"]
+   resourceNames: ["region:apac"]
```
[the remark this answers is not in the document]
ldpliu: Yes, the format must be `foo:bar`, if we need to support arbitrary value. may be we could use label like `foo:*`

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,284 @@
...
+metadata:
+ name: clustersetrole
+rules:
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclusters"]
+   resourceNames: ["mc1"]
+   verbs: ["create"]
+ - apiGroups: ["cluster.open-cluster-management.io"]
+   resources: ["managedclusters/label"]
+   resourceNames: ["region:apac"]
```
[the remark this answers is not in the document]
ldpliu: If we need to support arbitrary values, I think the `resourceNames` could be `region:*`

ldpliu:
I also add a option2 which has no external rbac control for clusterset creation and updating cluster label.
@elgnay @qiujian16 please help to review it.

enhancements/sig-architecture/30-clusterset-override/README.md
showeimer: Will we run into an issue with identical values in label key:value pairs and ManagedClusterSet being cluster-scoped;

for example:

foo=bar
MCS name = bar, exclusiveKey = foo

other=bar
MCS name = bar, exclusiveKey = other

ldpliu: We have changed the spec, and use `clusterLabel` to select managedClusters.

We need to make sure that, managedClusterSets should be exclusive within a purpose, but for different purposes it can overlap.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,235 @@
...
+If I want to add the label `cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>` to a managedCluster, I must have `create` permission for subresource `managedclustersets/join`. 
+If I want to add/update other labels, I just need to have `update` permission to the managedCluster.
+
+In this proposal, managedClusterSet could use any labels to select managedClusters. And if I add a label for a managedCluster, it may lead to the managedCluster added to a managedClusterSet. So we need to rethink about the rbac control of managedClusterSet
+
+We will use the following questions to consider the RBAC control.
+
+#### What kinds of roles exist related to managedClusterSet
+In General, there are three kinds of roles related to managedClusterSet.
+- Cluster admin
```
qiujian16, line 60: can you elaborate what actions each role will make here?

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,235 @@
...
+- `cloud: Amazon` means the cluster is provisioned in Amazon cloud provider
+- `middlewareEnabled: true` menas this cluster enables the middleware.
+
+The managedClusterSet could use these labels to select target clusters.
+
+3. Normal users
+Currently, Any user who has `update` permission to the `managedclusters` could add any labels to the managedCluster. And these labels may lead to the managedClusters added to a managedClusterSet.
+
+But for some managedClusterSet(like: disaster recovery managedClusterSet[a], resource group managedClusterSet[d]), we do not want any users to add their clusters to these managedClusterSets.
+For these managedClusterSets, we hope these managedClusterSets' `spec.clusterLabel.key` should have the same prefix. This prefix may be configured by the user. But in this proposal, we use `cluster.open-cluster-management.io/` as the prefix to identify the managedClusterSet which needs external permissions to add to.
```
qiujian16, line 89: Why it comes to this conclusion?

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+If a user wants to add these kinds of labels, he/she should have an external permission. So we need a way to define this permission.
+Investigate the existing rbac rule in current k8s, we use a rule which is similar to [csr-approver](https://github.com/kubernetes/website/blob/3ca34b9a3be454055ba234f1d2ff7d55809b5040/content/zh/examples/access/certificate-signing-request/clusterrole-approve.yaml#L25)
+We define a subresource `managedclusters/label`, which resource name is the label of managedCluster. This subresource means user have permission to add this label to managedClusters.
```
qiujian16, line 93: still need to elaborate how to define if a user can add any label keys or values?  Will it require any controller/agent to need this permission also?

elgnay:
The doc does not mention that `managedclustersets/join` will be deprecated.

ldpliu:
> The doc does not mention that `managedclustersets/join` will be deprecated.

I add a note to describe that at the end of this section https://github.com/ldpliu/enhancements-1/tree/clusterset-override/enhancements/sig-architecture/30-clusterset-override#who-can-add-a-managedcluster-to-a-managedclusterset

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: changed the description

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+2. Controller/agent which run in managedClusters
+   - The Controller/agent should not `create`, `get`, `list`, `update`, `delete` managedClusterSets
+   - The Controller/agent should not `bind` managedClusterSet.
+   - The Controller/agent should `get`, `update` related managedCluster which is running on.
+
+3. Non Cluster admin users
+   - These users should `get` some managedClusterSets
+   - These users should `bind` some managedClusterSets to their namespaces
+   - These users should `create`, `get`, `update`, `delete` some managedClusters
```
qiujian16, line 73: This is questionable to me. I think it should callout.

A non-cluster admin user cannot create clusterset.

Also it needs to examplain why 
A non-cluster admin user can create/delete cluster. The update permission should be changed. I do not think a controller or agent or non-cluster admin user can `accept` a cluster.

A controller or agent can update the label of the managed cluster.
A non-cluster user may need to update some cluster's label (not all)

ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,279 @@
...
+   - These users may need to `create` managedClusters(For Resource Group Purpose managedClusterSet[d], in each group, the group members may need to create the managedClusters). They could also `delete` these managedClusters, and `update` these managedClusters labels.
+
+#### Who can join a managedCluster to a managedClusterSet
+In this proposal, managedClusterSet could use any labels to select managedClusters. If one user adds a label to a managedClusters, it may lead to this managedClusters added to a managedClusterSet. So this question also means "who can add/update a label for managedClusters".
+
+1. Cluster admin
+Cluster admin could add any labels to any managedClusters.
+
+2. Controller/agent which run in managedClusters
+The controller/agent may add labels like: `region: apac`, `cloud: Amazon`, `middlewareEnabled: true`
```
deads2k, line 93: Is this the case currently or the case after this proposal?  Is there a specific prefix reserved?  should there be?

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,279 @@
...
+- `cloud: Amazon` means the cluster is provisioned in Amazon cloud provider
+- `middlewareEnabled: true` menas this cluster enables the middleware.
+
+3. Non Cluster admin users
+Currently, non cluster admin users may `add/update` some managedClusters labels. And these labels may lead to the managedClusters added to a managedClusterSet.
+
+But for some managedClusterSet(like: Disaster Recovery ManagedClusterSet[a], Resource Group ManagedClusterSet[d]), we do not want any users to add their managedClusters to these managedClusterSets. 
+
+For these managedClusterSets, we need a way to identify them. 
+We popose use the prefix `cluster.open-cluster-management.io/` in managedClusterSet `spec.clusterLabel.key` to identify the managedClusterSet which need "special" permission to join. (This prefix may be configured. But in this proposal, it is const)
```
deads2k, line 105: This is a good idea. We can add restrictions to labels in our domain, because we own that area.

deads2k, line 105: if a user adds, `something-else.open-cluster-management.io/foo` should that be subject to a secondary RBAC check or should any user be allowed to do that?

enhancements/sig-architecture/30-clusterset-override/README.md
deads2k: > I think we need to discuss about mitigation approach here.

mitigation or migration?

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,279 @@
...
+So, In this proposal, managedClusterSets could use any labels to select managedClusters. 
+
+So, we add a field `ClusterLabel` into managedClusterSet spec
+
+```go
+ type ManagedClusterSetSpec struct {
+ // ClusterLabel defines the label which the managedClusterSet will use to select managedClusters.
+ // If the field is not set, we will use "cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>" as default value.
+ // +optional
+ ClusterLabel ClusterLabel  `json:"clusterLabel"`
```
deads2k, line 38: why not a metav1.LabelSelector?

deads2k, line 38: part of me predicts this will not be the last grouping mechanism we choose.  How about


```go
type ManagedClusterSetSpec struct {
    Selector ManagedClusterSetSelector
}

type ManagedClusterSetSelector struct{
    // "" means to use the current mechanism of matching by name
    // "LabelSelector" means to use the label 
    // A theoretical "Named" means to use "Named []string" where the managedclustername is set
    SelectorType string 

    LabelSelector *metav1.LabelSelector

}

Comment (elgnay) [reply enhancements/sig-architecture/30-clusterset-override/README.md:38]: @deads2k Yes, you are right. I'd like to add another selector type `ClusterLabel` to
- Guarantee clustersets with same cluster label key are exclusive;
- Enable additional permission check when cluster joining/leaving a clusterset (the label key should start with the reserved prefix);

```go
type ManagedClusterSetSelector struct{
    // "" means to use the current mechanism of matching by name
    // "ClusterLabel" means to use a particular cluster label. It is guaranteed that clustersets with same label key 
    // are exclusive with each others;  
    // "LabelSelector" means to use the label 
    // A theoretical "Named" means to use "Named []string" where the managedclustername is set
    SelectorType string 

    LabelSelector *metav1.LabelSelector

    ClusterLabel *ClusterLabel
}

//ClusterLabel defines one cluster label
type ClusterLabel struct {
	Key string `json:"key"`
	Value string `json:"value"`
}
``` 

What do you think?

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: This is the case currently.
After the proposal, if user want to restrict the label, he/she could add labels like: `cluster.open-cluster-management.io/region:apac`. And add the permission to controller/agent
`
 - apiGroups: ["cluster.open-cluster-management.io"]
   resources: ["managedclusters/label"]
   resourceNames: ["cluster.open-cluster-management.io/region:*"]
   verbs: ["create"]
`

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: Any user be allowed to do that, we only check the label which key has prefix: `cluster.open-cluster-management.io/`.

enhancements/sig-architecture/30-clusterset-override/README.md
deads2k: > What do you think?

I think that looks good.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,358 @@
...
+```go
+type ManagedClusterSetSpec struct {
+    // Selector represents a selector of ManagedClusters by labels and names.
+    ClusterSelector ManagedClusterSelector  `json:"clusterSelector"`
+}
+
+type ManagedClusterSelector struct{
+    // "" means to use the current mechanism of matching label <cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>.
+    // "LabelSelector" means to use the LabelSelector to select target managedClusters
+    // "ClusterNames" means to use "ClusterNames" where the managedClusters Name is set
```
deads2k, line 40: This was just to prove the utility of the discriminated union.  If no one has a use-case for it, I'm ok not adding it in this iteration.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,358 @@
...
+    // +optional
+    SelectorType SelectorType `json:"selectorType"`
+
+    LabelSelector *metav1.LabelSelector `json:"labelSelector"`
```
deads2k, line 44: in the godoc here, please indicate that `cluster.open-cluster-management.io/*` have secondary ACL checks and that other labels do *not* have any secondary ACL protection.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,358 @@
...
+Currently, non cluster admin users may also `add/update` some managedClusters labels. 
+
+#### Add restrictions for adding/updating/deleting labels to managedClusters
+In this proposal, managedClusterSet could use any labels to select managedClusters. And the new added labels may lead to this managedClusters added to a managedClusterSet.
+
+But for some managedClusterSet(like: Disaster Recovery ManagedClusterSet[a], Resource Group ManagedClusterSet[d]), we do not want any users could add their managedClusters to these managedClusterSets. 
+
+So we propose adding restrictions for some labels, and if user want to add these kinds of labels, he/she must have a "special" permission. Then the managedClusterSet which do not want anyone to join could use these labels to select managedClusters.
+
+We propose adding restrictions for labels which key have prefix `cluster.open-cluster-management.io/`. (This prefix may be configured. But in this proposal, it is const)
```
deads2k, line 115: I suggest adding another prefix that the poke agents themselves are encouraged to use and have permissions on their own managedcluster to modify. this will give a "trust labels under this prefix to have been set by the managedcluster", which is likely useful for certain kinds of label selection.  This would be comparable to things like the kubelet architecture labels.

enhancements/sig-architecture/30-clusterset-override/README.md
deads2k: I agree that describing migration is useful.

1. This could impact clients who expect mutually exclusive sets of managedclusters: this is likely to be small and faced no matter what we do.
2. As long as the old RBAC check for setting `cluster.open-cluster-management.io/clusterset` still functions, old clients changing the label will be ok.  the fact that a new RBAC check could also grant permission won't stop those clients from working.

Any other cases to be concerned about?

enhancements/sig-architecture/30-clusterset-override/README.md
qiujian16: +1

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: removed the Named selector

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: add the ClusterLabel selector.

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: done

enhancements/sig-architecture/30-clusterset-override/README.md
ldpliu: I investigate the kubelet [doc](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) add another prefix `info.open-cluster-management.io/` for spoke agent.

enhancements/sig-architecture/30-clusterset-override/README.md
qiujian16: yes, I mean migration. I think we should also discuss the impact to the consumers during this migration.

deads2k:
I think this design looks good.  I think @qiujian16's question about a migration approach is still outstanding: https://github.com/open-cluster-management-io/enhancements/pull/37/files#r762771876

The concern may be about previous API consumers.  Without the discriminated union to start, some api consumers may believe that the only way the clusterset can be matched is based on the single label.  If those API consumers are also using this API to find matching clusters, the results the produce will be different. 

I suspect the concern could be mitigated by finding API clients that are generating matching managed clusters based on manageclustersets and updating them *first*.  This coudl be accomplished by merging just the discriminated union, rippling out, then merging the rest of the options later.

This only works if the number of current consumers developing their own matches is small.

qiujian16:
currently managedClusterSet only have 3 consumers
1. placement
2. ACM
3. submariner

> I suspect the concern could be mitigated by finding API clients that are generating matching managed clusters based on manageclustersets and updating them _first_. This coudl be accomplished by merging just the discriminated union, rippling out, then merging the rest of the options later.
> 

this makes a log of sense. @ldpliu I would be good if we add a migration section, and describe the migration path as above.

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+type SelectorType string
+
+const (
+    ClusterLabel        SelectorType = "ClusterLabel"
+)
+
+//ClusterLabel defines one cluster label
+type ClusterLabel struct {
+	  Key string `json:"key"`
```
qiujian16, line 356: key by default is cluster.open-cluster-management.io/clusterset and can only be cluster.open-cluster-management.io/

enhancements/sig-architecture/30-clusterset-override/README.md
```diff
@@ -0,0 +1,401 @@
...
+    Value string `json:"value"`
```
qiujian16, line 357: Value can only be empty or the name of clusterset

ldpliu: done

qiujian16:
/approve
/hold

hold in case anyone needs to review this.

PR #38, opened 2021-12-21
Add the Securing Sensitive Data in Hub Policy Templates proposal

mprahl:


enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
+# Securing Sensitive Data in Hub Policy Templates
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management/website/)
```
mikeshng, line 10:
```suggestion
      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
```
Sorry about this one. I have a PR open now to fix the template url.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+all policies as non-compliant. There are some workarounds for this, but it is
+proposed to skip this for the upcoming release and provide documentation to
+manually rotate the key in the event a customer detects that it has been
+compromised. This will likely be made automatic in a future release.
+
+#### RBAC
+
+Since the encryption key is stored in a secret in the managed cluster namespace,
+it is important that non-privileged users do not have access to read this value.
+The default ACM roles of `open-cluster-management:view:<managed_cluster_name>`
```
mikeshng, line 167:
```suggestion
The default OCM roles of `open-cluster-management:view:<managed_cluster_name>`
```
Can you verify in OCM is the same set of roles?

mikeshng (commented):
Minor nits.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Proposal
+
+To allow users to configure managed clusters using policies containing arbitrary
+sensitive data, it is proposed to add a template function called `protect`,
+which will allow the user to encrypt a string in a Hub policy template. For
+example:
+
+```yaml
+'{{hub "(lookup "v1" "Secret" "default" "my-hub-secret").data.message | protect
```
qiujian16, line 59: please clarify the function parameter here. if "default" is ns, why it is set here? What is allowed value, to be encrypted? can it be any interface{}?

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+To make things easier for the user, a modified version of the `fromSecret`
+template function will be made available for the Hub policy templates. It’ll
+accept the same input as its managed cluster counterpart but output a Base64
+encoded string of the encrypted value, just as the previous example looked like
+on the replicated policy. If a user needs to alter the secret value such as
+base64 decode it, they would need to use the first example. Here is an example
+of the modified `fromSecret` template function:
+
+```yaml
+'{{hub fromSecret "default" "my-hub-secret" "message" hub}}'
```
qiujian16, line 83: please also explain the function parameters and allowed values here

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+The AES-256 key needs to be generated and stored both on the Hub cluster and the
+managed cluster. There would be a unique key per managed cluster. This would
+mean that if a single managed cluster is compromised, it does not mean that all
+encrypted values in the other managed clusters have also been compromised.
+
+The Policy Propagator should be responsible for generating the AES-256 key for
+each managed cluster. Each key would be stored in a secret called
+`policy-encryption-key` in each respective managed cluster namespace. The key
+would be stored on the secret’s `key` property. This secret should also have the
+`cluster.open-cluster-management.io/backup: 'policy'` label set so that the key
```
qiujian16, line 140: I do not think we need to callout the backup part.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+is backed up during backup and restore situations. With this in place, the
+Policy Propagator is able to encrypt the data using the `protect` template
+function, however, the managed cluster also needs the key when it decrypts the
+data. The Policy Spec Sync controller should also sync the
+`policy-encryption-key` secret to the managed cluster in the managed cluster
+namespace by watching the managed cluster namespace for secrets on the Hub. It
+is also desired to allow customers to provide their own keys. This would be
+limited to AES-256 keys unless a customer requests otherwise.
```
qiujian16, line 148: If this key is modified or deleted, does it mean the policy on spoke can not be decrypted? Will it transport key once or multiple times?

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
qiujian16: we do not have this settings in upstream, I think we should remove this part to avoid confusion

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+It is proposed that this be considered stable in the first release after the
+implementation is complete.
+
+### Upgrade / Downgrade Strategy
+
+N/A
+
+### Version Skew Strategy
+
+N/A
```
qiujian16, line 203: I think version skew should be discussed, For instance, what if you have a high version hub control plane and a lower version agent?

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+   replicated policy and it can be decrypted on the managed cluster.
+
+### Graduation Criteria
+
+It is proposed that this be considered stable in the first release after the
+implementation is complete.
+
+### Upgrade / Downgrade Strategy
+
+N/A
```
qiujian16, line 199: if we modify the existing function, the upgrade should be discussed.

qiujian16 (commented):
Thanks, I think there are some parts to be clarified
1. who will be the consumer for this function, will a controller be the consumer and what is the impact?
2. What should be handled is a user modifies or deletes the secret for private key on hub or spoke?
3. how to handle version skew?

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
```diff
@@ -0,0 +1,268 @@
...
+### Resolve All Sensitive Data at the Point of Enforcement
+
+The general idea with this approach is that any sensitive data on the hub that
+needs to be propagated to the managed cluster would need to be stored in the
+managed cluster namespace. Then when the Configuration Policy controller
+assesses the policy on the managed cluster, the sensitive data is retrieved on
+the Hub. This means that the sensitive data is never stored in a policy in any
+form.
+
+From a user experience perspective, this is not ideal since it means that a user
```
qiujian16, line 224: I am not sure I fully understand this. I think your function can read the original secret and create a new secret containing only the interested data that user cares in the managed cluster ns? A single policy could store all these data in a single secret if there are multiple function defined in this policy.

The benefit is you do not need to maintain a key for each cluster and the total data payload is reduced.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
mprahl: The namespace has to be the namespace in which the policy is created in. This is an existing restriction for Hub policy templates that does not change with this feature.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
mprahl: The secret is constantly synced from the Hub to the managed cluster. I'll add that detail in the doc.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
mprahl: This alternative does not account for a desired use-case where a single secret contains the configuration for all managed clusters. For instance, if I have a secret like the one shown below, I'd have to copy that secret to all managed cluster namespaces that need only a couple of the values, thus exposing sensitive data that isn't required for that cluster. Alternatively, I'd have to create 1000 secrets to configure 1000 managed clusters and they'd all individually have to be in the managed cluster namespace. With the actual proposed solution, you could just use `{{hub printf "%s.ssl-key" .ManagedCluster |  fromSecret "" "cluster-config" hub}}` in a policy to get the single key for this managed cluster. It'd then get encrypted and stored in the replicated policy.

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: cluster-config
  namespace: my-config
data:
  cluster1-ssl-cert: .....
  cluster1-ssl-key: .....
......
  cluster1000-ssl-cert: .....
  cluster1000-ssl-key: .....
```

mprahl: @qiujian16 after rereading what you suggested, this would not work because it'd cause a `Secret` to be created for every time a Hub policy template uses `fromSecret` or `protect`. So if I have five policies which use this feature that apply to 1000 managed clusters, I'd end up with 5000 secrets being created on the hub.

My proposal just creates the additional secret per each managed cluster to store the encryption key and no more.

Additionally, my proposal allows the ACLs of the managed cluster service account which copies over the secret to be restricted to the `policy-encryption-key` secret in the managed cluster namespace. The approach you suggested would end up requiring ACLs to read all secrets in the managed cluster namespace. This could be a security issue depending on what secrets other components of ACM create in the managed cluster.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
mprahl: I'll mention this in the doc for clarity.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
qiujian16: this is an interesting point. Because in one hand, we increase the payload size in policy, and on the other hand we add additional secret. 

I think it need some clarification on why the the first approach will be better than the second. 
From payload size part, it seems the two approach are similar. 
From the number of update call, I think you will still have to watch the secret and update content of per cluster ns policy? 

ACL is an interesting case. I think the hub controller can set permission for the policy agent so it can only get a subset of the secrets in cluster ns.

qiujian16 (commented):
thanks @mprahl I think there are still two parts that need more clarification
1. the error handling in the template function.
    e.g. how to handle If a secret payload or object field does not exists?
2. more comparison to the alternate approach.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
mprahl: @qiujian16 in regards to the ACLs, you can set the managed cluster service account to have access to a subset of secrets, which is what my proposed implementation uses, where the restriction is set on the `policy-encryption-key` secret in the managed cluster namespace. You cannot, however, limit the ACLs to a subset using a wildcard or regex. Some other process would need to dynamically update the ACLs to include/remove secret names and that would be complex and lead to race conditions and security concerns. Additionally, the Policy Propagator would need to be granted access to read, create, update, and delete *all* secrets on the Hub since we could not restrict it by the Secret name since the value isn't static, and the managed cluster namespaces are not static, so we cannot restrict these ACLs to just those namespaces.

I would also argue that the payload size in the policy is less than creating a Secret for each managed cluster the policy applies to. For instance, the encrypted value in the policy is `$ocm_encrypted:<base 64 of encrypted value>`. Since this is using AES encryption, the resulting encrypted value is the same size as the plaintext excluding the additional padding added to the plantext value so it matches the AES block size. Then the Base64 encoding of the encrypted value does introduce some waste alongside the `$ocm_encrypted:` prefix.

That being said, a Secret will always have the `apiVersion`, `kind`, `metadata.name`, `metadata.namespace`, and the key/field names for the secret values set. Depending on how this is stored in etcd under the hood, it could lead to more wasted storage. It'd also add additional indexes in etcd which wouldn't be necessary in my preferred proposal.

Also, a value in each replicated policy would need to be put as a placeholder so that when the policy is processed on the managed cluster, it knows which secret and key/field to use when finding the value. This alone would often be more than the waste associated with the encryption proposal.

The additional API requests would increase significantly over the preferred proposed solution. The reason being that in the preferred solution, only a single Secret per managed cluster needs to be watched, which limits the resource consumption of the watch significantly. It is only updated as part of a key rotation every 30 days or if a cluster administrator mistakenly updates/deletes the key. On the retrieval side, the Policy Propagator would just read it once and cache the value (the controller that handles the key rotation updates this cache when the key changes). The same can be said on the managed cluster side when processing the policies.

With this other solution, every time the policy is propagated on the Hub, it would need to get the original Secret, compare it to the copied Secret in the managed cluster namespace and make updates to it as necessary. On the managed cluster side, it's not reasonable to cache all these secrets, so it would lead to the managed cluster reading the API for these Secrets every time the policy is processed, which is currently every 10 seconds.

I'm happy to update the doc with this, but I think this is too much detail and the doc captures this well enough.

mprahl:
> thanks @mprahl I think there are still two parts that need more clarification
> 
> 1. the error handling in the template function.
>    e.g. how to handle If a secret payload or object field does not exists?
The error handling does not change compared to existing Hub policy templates (e.g. `fromConfigMap`). If the Secret can't be retrieved, then the policy becomes noncompliant with an error stating what happened.

What in particular are you looking for?

> 2. more comparison to the alternate approach.
I responded to this in your other comment.

qiujian16:
With error handling, i am saying when an error happens in this func

```
lookup "v1" "Secret" "default" "my-hub-secret").data.message | protect
```

how do you let user know. non-compliant does not help much I think, since user will not know whether it is a configuration error or execution error on spoke. Will there be a certain status condition that indicate this? What if you have multiple func above called in one policy? I think this should be clarified on the doc.

enhancements/sig-architecture/33-hub-templates-securing-sensitive-data/README.md
qiujian16: What i mean is you could have a secret in cluster ns per policy. For instance a user creates a policy (policy1) which calls `lookup "v1" "Secret" "default" "my-hub-secret" ` multiple times. The policy controller can create a secret "policy1" in the cluster ns and store the secret payload defined in the the policy. 

The Policy Propagator does not need to watch all secrets in this cluster ns. A simple way is to define a label on secret. since this secret is created only by policy hub controller, so adding a label should be fine. Another more complicated approach is to set permission for Policy Propagator so it can only watch secret with certain names in this cluster ns.

> With this other solution, every time the policy is propagated on the Hub, it would need to get the original Secret, compare it to the copied Secret in the managed cluster namespace and make updates to it as necessary. 

But you also need to do the similar thing with encrypted approach, right? Watch all the secrets on the hub and update the encrypted payload. 

> On the managed cluster side, it's not reasonable to cache all these secrets

In encrypted approach, you actually cache it in the spoke already. right? Why we cannot cache secret?

I agree that the payload size is bit higher, and ACL management is more complicated. I think the benefit of referencing is the simplicity on API definition and it will be easier for other controller to reason about this policy API in the cluster ns.

mprahl: > What i mean is you could have a secret in cluster ns per policy. For instance a user creates a policy (policy1) which calls `lookup "v1" "Secret" "default" "my-hub-secret" ` multiple times. The policy controller can create a secret "policy1" in the cluster ns and store the secret payload defined in the the policy.
> 
> The Policy Propagator does not need to watch all secrets in this cluster ns. A simple way is to define a label on secret. since this secret is created only by policy hub controller, so adding a label should be fine. Another more complicated approach is to set permission for Policy Propagator so it can only watch secret with certain names in this cluster ns.

@qiujian16 I wasn't saying that the Policy Propagator needs to watch all secrets. I'm saying it requires the ACLs to create, update, and delete *all* secrets on the Hub since the secret names and namespaces are not known at OCM installation time.

I would argue that it may need to watch all these copied secrets since if a managed cluster admin accidentally modifies or deletes this secret, the Policy Propagator would need to recover. In the encryption approach, only a single secret is watched per managed cluster and the ACLs are scoped to that. The actions performed on the secret are just verifying the key is valid, and perform key rotation every 30 days. This is very lightweight.

As I said in my previous comment, every time the Policy Propagator processes the policy or in the case a new controller is added that watches all these copied secrets, it'd have to query all the secrets that are referenced in the `fromSecret` template function in the policy, merge them all into a single secret and verify that the copied secret matches. This is resource intensive on the Kubernetes API considering a single policy being processed could apply to a 1000 managed clusters and this would have to be done for all 1000 managed clusters. The encryption approach does not have this overhead. The encryption algorithm introduces some overhead, but it's localized to the CPU of the Policy Propagator container and not the Kubernetes API.

> 
> > With this other solution, every time the policy is propagated on the Hub, it would need to get the original Secret, compare it to the copied Secret in the managed cluster namespace and make updates to it as necessary.
> 
> But you also need to do the similar thing with encrypted approach, right? Watch all the secrets on the hub and update the encrypted payload.

The encrypted payload only changes when the key is rotated, once every 30 days. The controller that does the key rotation targets only the policies that need to be reprocessed and marks them for the main Policy Propagator controller to reprocess them.

> 
> > On the managed cluster side, it's not reasonable to cache all these secrets
> 
> In encrypted approach, you actually cache it in the spoke already. right? Why we cannot cache secret?

It is safe to cache the encryption key on the spoke since for one, it's only 256 bits. Additionally, the spoke knows when the cache is invalid because when it tries the decryption, it will fail, then the cache will be updated and the decryption tried again.

If you are caching all the copied secrets on the spoke, you would need to have the controller perform a watch on the secrets in the spoke's namespace so that the cache is always up to date.

> 
> I agree that the payload size is bit higher, and ACL management is more complicated. I think the benefit of referencing is the simplicity on API definition and it will be easier for other controller to reason about this policy API in the cluster ns.

As stated above, the ACL management is not only more complicated, it is quite a bit less safe.

What do you mean by the simplicity on the API definition? It's just as custom of an approach since the replicated policy on the spoke/managed cluster would still need some kind of custom policy template to reference the copied secret. I think it's a lot more complex to reason with. On the encryption approach, it's quite simple. You use a policy template and the value is stored as encrypted in the policy. On the spoke/managed cluster, the encrypted value is detected and then encrypted when processing the policy.

mprahl:
> With error handling, i am saying when an error happens in this func
> 
> ```
> lookup "v1" "Secret" "default" "my-hub-secret").data.message | protect
> ```
> 
> how do you let user know. non-compliant does not help much I think, since user will not know whether it is a configuration error or execution error on spoke. Will there be a certain status condition that indicate this? What if you have multiple func above called in one policy? I think this should be clarified on the doc.

Like I said, this behavior does not change from the existing Hub policy templates implementation which is why I think this is out of scope of this design document. If there is an error processing the templates, the policy is marked as noncompliant and the noncompliant message on the policy says what the error is.

qiujian16:
/approve
/lgtm
/hold

It looks fine in general. I would suggest to update the alternative section with more comprehensive analysis. hold in case you need anyone else to review.

mprahl:
@qiujian16 I just changed the document to include some of our discussion in the GitHub comments.

Could you please take another look?

PR #41, opened 2022-02-09
KEP: Work executor subject

yue9944882:


qiujian16 (commented):
it looks good in general to me

/assign @deads2k 
David, any suggestion on this?

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,102 @@
...
+  
+## Proposal
+
+### API Changes
+
+This proposal adds a new field `executorGroupName` to the `ManifestWork` api, 
+which is defaulted to (the string) "default" if absent. Meanwhile, upon starting
+a work agent we can set its executor group via an optional flag:
+
+> `--executor-group=<group name>`: If the work agent is running with a non-empty
```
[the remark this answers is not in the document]
yue9944882: the work agents under the same group are supposed to work under "leader-election". there're multiple instances but only one will take effect at a time.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,146 @@
...
+> executor group, it will only manage those manifest-work resources that assigned
+> the same group name. By default, an empty group means the agent is the default
+> executor.
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata: ...
+spec:
+  executorGroupName: default|...
```
qiujian16, line 72: how about

```yaml
executor:
 group: default
```

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Controlling availability (blast-radius) of the work agents
+
+Currently, for each of the managed clusters we will deploy the work agents
+in a single deployment resource which is sharing the availability via "leader-
+election", so there's merely one instance of work agent taking effect at a time.
+Under this pattern, we can't break down the availability of the work agent into
+finer units so that:
+
+- During the upgrade, we can progressively reduce the impact from the agent's
```
deads2k, line 29: During an upgrade, wouldn't we roll over a single deployment pod at a time, so that the outage would be negligible?  If not, is something wrong with the deployment structure or perhaps the lease configuration (perhaps graceful release is disabled?)

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,147 @@
...
+Currently, for each of the managed clusters we will deploy the work agents
+in a single deployment resource which is sharing the availability via "leader-
+election", so there's merely one instance of work agent taking effect at a time.
+Under this pattern, we can't break down the availability of the work agent into
+finer units so that:
+
+- During the upgrade, we can progressively reduce the impact from the agent's 
+  down-time.
+- By grouping the agents according to their use-cases, we can control the 
+  blast-radius so that the unavailability will not propogate between the groups.
```
deads2k, line 32: Is this related to upgrade, related to node availability, or related to a controller bug?  Sharding inside of a single cluster isn't a thing we've had to do before for performance or availability reasons.  There are examples of sharding (ingress for instance) based on capability and configuration differences and underlying tech.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Separate the identity and its held permission for the work agents
+
+There can be different actors using the work api to distribute resource from 
+the hub cluster. Currently the work agent has the cluster-admin access to the
+managed cluster which means that every hub users with write access to the
+work resources will be able to dispatch any resources to the managed cluster.
+By making different groups of work agents mount different service accounts and
```
deads2k, line 40: This is a cool use-case.  The goal being to control permissions of the hub users on the spokes?

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,147 @@
...
+> the same group name. By default, an empty group means the agent is the default
+> executor.
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata: ...
+spec:
+  executor:
+    group: default|...
```
deads2k, line 73: Will there be restrictions about who can use certain executor groups?  

if not, what is the value of 
> By making different groups of work agents mount different service accounts and granting different scopes of permission to those service accounts, we will be able to further define the expected purpose/scope of the work agent groups.

it becomes a benefit of organization on the spokes, but not a restriction of power from hub?  That's ok with me, but I'd like to be sure I understand it.  Perhaps someone could write custom rules restricting usage in their hub.


If yes, this appears to complicate the handling of any abstraction on top of manifestwork.  I don't think I'd recommend restricting the ability to set this field out of the box.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: i removed the agent instance splitting in the update. by leveraging the impersonation, we can switch the identities from one instance on the fly.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: removed the instance sharding. in the first phase the dynamic impersonation looks more feasible and light-weight to bring into implementaion. the controller sharding is a tough topic, i narrowed the scope of this proposal to dynamic idenitity mgmt.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: the permission control can be done by introducing a validating admission webhook that authorizes whether a hub user can use a executor with the SubjectAccessReview api. i clarified it in the latest revision, PTAL.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: comparing to the relationship between the original service-account and the pods, the kube-apiserver doesnt authorize whether a hub user can mount a service-account when creating pods. but they can opt-in that authorization by developing a custom admission webhook. that's similar for the proposed api change in the work api, the work agent isnt able to authorize that until we plumb a validating admission webhook to the hub. i think the alpha phase wont cover the implementation of that webhook, but we can expect that in the beta. wdyt?

deads2k:
> removed the instance sharding. in the first phase the dynamic impersonation looks more feasible and light-weight to bring into implementaion. the controller sharding is a tough topic, i narrowed the scope of this proposal to dynamic idenitity mgmt.

So the overall object is about scoping permissions?  Are you most worried about
1. a hub user being able to create things he shouldn't on the spokes
2. the controller on the spoke having too much permission on the spokes

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+Currently all the API requests raised by work agent in the managed cluster are
+from a fixed service account ("open-cluster-management-agent/klusterlet-work-sa"
+under the default installation). However in some cases, we need to clarify the
+source/owner identity of the `ManifestWork` resource so that we can record the
+api audits in a finer granularity. This will be helpful for the managed 
+cluster's admin to be capable of knowing "who is operating that `ManifestWork`
+from the hub cluster".
+
+### Separate the identity and its held permission for the work agent
```
deads2k, line 31: this motivation is phrased in terms of actors on the hub cluster that should have different levels of permission in the spoke cluster.  It does not seem directly concerned with the work controller in spokes having high privileges.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,219 @@
...
+- This proposal doesn't cover which identity to impersonate in the `ManifestWork`
+  api, i.e. it differs depending on use-cases.
+  
+## Proposal
+
+### API Changes
+
+This proposal adds a new field `executorSubject` to the `ManifestWork` api, 
+which is defaulted to nil if absent. A nil executor subject indicates that the 
+work agent doesn't perform any impersonation before requesting the hosting
```
deads2k, line 60: Rather than impersonation, could we use a subjectacessreview for a particular verb, resource, subject tuple?

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+A sample of the new `executorSubject` field in Yaml is shown below:
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata: ...
+spec:
+  executorSubject:
+    type: [ServiceAccount|User] # union discriminator
```
deads2k, line 83: I recommend against adding user, but discriminator makes sense to keep.  I could imagine a future with a `SpokeWorkRole` for instance.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Motivation
+
+### Auditing the API request history from work agent
+
+Currently all the API requests raised by work agent in the managed cluster are
+from a fixed service account ("open-cluster-management-agent/klusterlet-work-sa"
+under the default installation). However in some cases, we need to clarify the
+source/owner identity of the `ManifestWork` resource so that we can record the
+api audits in a finer granularity. This will be helpful for the managed
```
deads2k, line 27: How firm a requirement is this?  Impersonation is a high privilege, but your second motivation could be achieved similar to the openshift templateinstance API.  Only I would suggest constraining yourself to service accounts.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: yes the work-agent will always need high privileges, what we're expecting is to muzzle that high privilege by adding a client-side authorizer to the work client.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: replaced impersonation w/ proactive authorizer via SAR calls

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: yes in the executor subject we're referencing a "role" instead of a "user". i removed "User" type and keep only "ServiceAccount".

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: in the latest design, i completely removed impersonation in favor of client-side authorization

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,211 @@
...
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata: ...
+spec:
+  executorSubject:
+    type: [ServiceAccount|Username] # union discriminator for extensibility
+    serviceAccount:
+      name: ...
+      namespace: ...
```
spadgett, line 75: What stops me from specifying the username of a cluster admin and still applying any changes to the managed cluster?

yue9944882: it is supposed to be defensed by an admission webhook, which is also elaborated in the proposal. the detail of admission webhook will not be covered in the current design, and it's always recommended to use `ServiceAccount` rather than `Username`. for `ServiceAccount` type executor, a feasible API model of the policy configuration of the admission webhook is `ManagedServiceAccountBinding`, e.g.:

```yaml
apiVersion: authentication.open-cluster-management.io/v1alpha1
kind: ManagedServiceAccountBinding
spec:
  ref:
    namespace: <namespace of the managed service-account>
    name: <name of the managed service-account>
  subjects:
    - kind: User
      name: dave
    - kind: Group
      name: my-group
    - kind: ServiceAccount
      namespace: <namespace>
      name: <name>
```

the binding api will define the mapping/accessibility between real hub users and the virtual service-account.

jnpacker: If there is 1 or more `ManagedServiceAccountBinding` present, would you block the use of `executorSubject` for that ManagedCluster, or selectively applying the admission webhook (to ManagedCluster namespaces) for backwards compatibility?

yue9944882: @jnpacker i think the webhook can do the selective applying via label selecting. e.g. the admission webhook only applies to those `ManifestWork` that has `work.open-cluster-management.io/executor-subject-gated: "true"`. so the existing works need to opt-in to the webhook by adding the label otherwise they're not protected, which keeps the backward compatibility. 

in abstract, by introducing the webhook, we will divide the `ManifestWork`'s write permission into two finer permissions:
- (1) the work executor who cannot editing the current executor subject but can modify all the other fields on the work. otherwise a hub user can unexpectedly leak the ownership of the `ManifestWork` and can't do anything to get it recovered.
- (2) the work owner who has complete access to the work e.g. create/delete the work, edit exeuctor subject 

the role (2) can be either a hub cluster global admin or a higher controller orchestrating the work across clusters. the role (1) can be an application domain owner. this is what will happen in detail when an arbitrary hub user trying to edit the `ManifestWork`:

1. does that hub user has the write permission upon the work in terms of Kubernetes RBAC? if no, rejected.
2. is that hub user writing upon a work resource w/ non-empty executor subject? if no, pass (for the backward compatibility).
3. is that hub user tweaking the current executor subject?
    - if yes, the hub user should have the RBAC permission to `manifestworks/assign` virtual subresource.
    - if no, fallthrough
4. is that hub user bound to the current executor subject? (which is checked by the admission webhook)
    - if the hub user has matching `ManagedServiceAccountBinding` bound, the edit passes.
    - if no, reject
    
 note that in the above process, the webhook is responsible for checking step 2 3 4.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,211 @@
...
+the work agent keeps using the original service account.
+
+
+### (Optional) Authorizing hub-user to executor subjects
+
+We can do an optional authorization in the hub cluster to check whether the
+manipulator of the `ManifestWork` can use a `ServiceAccount` as the executor
+subject by the integration between `ManagedServiceAccount` api. The 
+authorization works by plugging an admission webhook to the hub cluster.
+The webhook list/watches the `ManagedServiceAccountBinding` resource which
```
deads2k, line 159: Could you actually use RBAC in the hub cluster to manage this based on a virtual verb (`useOnManagedCluster`?) permission?

enhancements/sig-architecture/34-work-executor-group/README.md
deads2k: > replaced impersonation w/ proactive authorizer via SAR calls

These SAR requests look like could work.  It might be easier to add the admission plugin at the same time as the field instead of afterwards.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+A sample of the new `executorSubject` field in Yaml is shown below:
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata: ...
+spec:
+  executorSubject:
+    type: [ServiceAccount|Username] # union discriminator for extensibility
```
deads2k, line 72: I like the structure here, but I suggest starting with only serviceaccount.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: added a new section about authorizing hub users via SAR in an validating admission webhook

enhancements/sig-architecture/34-work-executor-group/README.md
deads2k: Users are short lived entities compared to serviceaccounts (they come and go from companies for instance).  Having this structure to choose a different subject type eventually is good, but I'd suggest extending it as part two.  Because you need compatibility with the existing API, you cannot start with a mechanism to default this to "me".

yue9944882: @deads2k fair, i removed username type in the latest update

yue9944882:
@qiujian16 @deads2k i just added a "function spec" clarifying the intention of this proposal. also, i updated the SAR request template in the admission webhook to clarify that we dont actually do the multi-cluster resource permission management in the hub cluster, the RBAC related validating are all supposed to be done in the spoke cluster. while in the hub cluster, we merely do the authorization of "which hub user can manipulate which work resource belonging to a executor subject" i.e. the mapping the hub users and executor subject.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Work-agent client permission muzzling
+
+The supported executor subject types will be merely "ServiceAccount".
+For "ServiceAccount" type, intuitively the referenced service account is 
+working like an "ambassador" identity, and the work agent will need to 
+explicitly check whether that "ambassador" has sufficient RBAC permission 
+in the managed cluster via SubjectAccessReview api. Due to the limit of 
+the SubjectAccessReview api, we need to send (# of contained objects) times 
+SAR requests to the spoke cluster, but this should be fine b/c the work-agent
```
deads2k, line 107: Yes, and you'll be able to cache a result if you like.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,253 @@
...
+subject specified is applied to the hub cluster:
+
+0. Create a "ManifestWork" with non-nil executor subject
+1. The work-agent observes the new resource via watch event
+2. The work-agent builds a SubjectAccessReview request which:
+   - The target subject is the executor subject from `ManifestWork`
+   - The target attribute should contain the "*" verbs for all the inlined
+     objects in the spec of `ManifestWork`. Note that the "resourceNames"
+     should also be specified.
+3. The work-agent writes the objects to the managed cluster if SAR passes.
```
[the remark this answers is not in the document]
deads2k: Followup, will deletion be prevented if identity cannot delete or is cleanup immune?  Thinking about the options, I think you must honor the identity restrictions on deletion to avoid someone placing an item and then removing it to trigger a deletion they cannot manage.  You should specifically enumerate this.

In addition, perhaps you should avoid creating a thing if the identity cannot delete it on cleanup.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,253 @@
...
+  user: <requester username>
+  groups: <requester groups>
+  resourceAttributes:
+    # virtual verb
+    verb: execute-as
+    # the GVR tuple of the requesting resources
+    group: work.open-cluster-management.io
+    version: v1
+    resource: manifestworks
+    namespace: <cluster namespace>
```
deads2k, line 189: I think I understand the reasoning for this. This is chosen instead of the service account namespace because this allows for delegation inside of a single managed cluster.  this reasoning is worth commenting here.

deads2k (changes requested):
This is looking very good to me.  I have some comments requesting clarification, but I think the cases are well covered now.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: added that in the new commit

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+Note that the authorization via subject access review api doesn't require the
+service account to be actually present in the managed cluster, so in the cases
+where you have too many executor subject to manage, the number of service 
+accounts resources in the managed cluster is bounded.
+
+A sample of the requesting `SubjectAccessReview` payload is:
+
+```yaml
+apiVersion: authentication.open-cluster-management.io/v1alpha1
```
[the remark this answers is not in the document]
yue9944882: oops typo

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,253 @@
...
+kind: SubjectAccessReview
+spec:
+  # requester is the referenced service-account in executor subject
+  user: "system:serviceaccount:<sa namespace>:<sa name>"
+  groups: ["system:serviceaccounts", "system:authenticated"]
```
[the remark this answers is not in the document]
yue9944882: fixed

enhancements/sig-architecture/34-work-executor-group/README.md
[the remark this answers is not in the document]
yue9944882: @deads2k am leaning on option 1. imagine if we're deploying a combo of deployment+service in a work object, and in an update i updated the mounting service-account for that deployment from "default" to a "foo", then we will run into failure due to insufficient permission for writing service-account. in this case if we allow partial updates, the updated deployment will be constantly failing b/c that "foo" SA doesnt exist. generally i think even tho the atomic strategy is a bit clumzy but it will save some problems from happening. i clarified that in the new commit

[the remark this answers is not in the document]
yue9944882: > In addition, perhaps you should avoid creating a thing if the identity cannot delete it on cleanup.

@deads2k i also clarified the deletion sequence in the appended commit. short version: the deletion flow will also work with an atomic strategy. the work-agent wont delete any objects until the executor subject matches all the pre-conditions.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: yes the managed cluster admin should also be able to grant/revoke accessibility of an executor-subject from the hub users. i added that in the new commit.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+
+A sample of the new `executorSubject` field in Yaml is shown below:
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata: ...
+spec:
+  executorSubject:
```
qiujian16, line 90: I think it would be better as

```yaml
spec:
  executor:
    subject:
      type: ServiceAccount
      ...
```

It give us space to add other field for executor in the future. But we could discuss this in the api PR.

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,275 @@
...
+5. The work-agent reports the results back to the hub cluster.
+   
+
+##### Delete sequence
+
+1. Delete a "ManifestWork" with non-nil executor subject
+2. The work-agent builds a SubjectAccessReview request which:
+    - The target subject is the executor subject from `ManifestWork`
+    - The target attribute should be the "delete" verbs for all the inlined 
+      objects in the spec of `ManifestWork` with  "resourceNames" specified.
```
qiujian16, line 172: nit: this will need us to return a condition about deletion of manifestwork. We haven't done't this yet.

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: @qiujian16 updated, on the other hand i have a bit complex feeling about overly expanding the struct, which might involve complexity for the api consumers. but expanding this one may be fine, i can imagine we add other fields like `exectuor#rateLimit` or `executor#namespacedOnly` in the future

enhancements/sig-architecture/34-work-executor-group/README.md
yue9944882: @qiujian16 do mean we can report the deletion failure as a condition in the work's status? currently we have the finalizers, a remaining finalizer on the deleted object indicates that something is failed/unfinished during the background deletion

deads2k:
During implementation, I think you'll want a local cache of authorization decisions for performance reasons.


/lgtm

enhancements/sig-architecture/34-work-executor-group/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+1. A hub admin can limit the scope of the dispatch'able resources for a 
+   `ManifestWork` by setting a non-empty executor subject, so the work agent
+   in the spoke cluster can report failure when applying/modifying out-of-scope 
+   resources.
+2. A spoke admin can expand the permission scope of the executor subject 
+   (e.g. a service-account) by binding more ClusterRole/Role to it.
+3. (beta phase) A hub user should pass the authorization before writing
+   the resources in the work object. The hub cluster holds a policy configuration
+   of "which hub user has write access to which work resource" and a validating
```
ryanzhang-oss, line 70: IIRC, this means check if "user A can create/update resource B", right? 
Is it also possible to check if "user A can create/update resource B in namespace C"?

enhancements/sig-architecture/34-work-executor-group/README.md
ryanzhang-oss: How to check if the hub user actually owns the service account in the spoke cluster?

PR #50, opened 2022-04-11
Manifestwork Update Strategy

qiujian16:

### DIFF
--- a/enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
+++ b/enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
@@ -1,0 +1,223 @@
+# Manifest Update Strategy in ManifestWork
+
+## Release Signoff Checklist
+
+- [] Enhancement is `implemented`
+- [] Design details are appropriately documented from clear requirements
+- [] Test plan is defined
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Today work agent do a fully update of each manifest defined in the `ManifestWork`. There is some special treatments for different resources and fields:
+
+### `ApplyDirectly` for known types
+
+`ApplyDirectly` in `openshift/library-go` is used for resources with known types, including those resources in core API group, CRDs,
+APIServices etc. In general, `ApplyDirectly` will get the existing resources; compare the spec of existing and required, and update
+the existing if there is diff. `Labels` and `Annotations` will be merged from required to the existing resource. 
+In addition, there are special treatments for certain types. For example, if `Secret.type` is changed in the required resource, `ApplyDirectly` will delete the existing and create a new `Secret` as required, since `Secret.type` is an immutable field.
+
+### Apply unstructured
+
+If a resource is not in the known type for `ApplyDirectly`, the resource will be regarded as an unstrcuture object. `Labels` and `Annotations` will also be merged similar as `ApplyDirectly`. For fields not in `metadata` and `status`, work agent will do a
+deepequal and fully update the resource if there is diff. Apply unstructured covers much more resource types including `apps` API
+group and all the custom resources.
+
+However apply unstructure method brings several issues for work agent:
+
+#### Too many update calls
+
+Kube APIServer will commonly set default values for resources, and there will always be diff when work agent compare existing with
+required resource. It will cause many unnecessary update calls. For example, when a `deployment` is created, kube-apiserver will set
+default service account for the deployment spec. Since the manifest defined in `ManifestWork` does not have this field, work-agent
+will update deployment constantly.
+
+#### Apply failure
+
+Some field of the resources has both defaulting and immutable set, with which work agent is unable to update the resource. `Job` is an example. If user does not set `spec.Selector` of job, the kube-apiserver will set a default selector and label on pod. 
+These fields are immutable, so work agent will fail later when it compares the job manifest and tries to update the job.
+
+#### Fight with other actors
+
+A resource might not only managed by work agent, but also other actors. Other actors would want to mutate some field in the resource
+which will later be rolled back by the work agent. Some use cases may need these actors to control some fields of the resources other than work agent.
+
+A possible solution to resolve the first two problems is that work-agent records the hash of required spec and the generation of the last applied resource, and instead of comparing required to the existing resources, work-agent compares hash (if required changes) and generation (if existing changes). However, it cannot resolve the third problem involving multiple actors.
+
+## Motivation
+
+Provide an interface for user to control how resource defined in `ManifestWork` should be updated.
+
+## Proposal
+
+We propose to update `ManifestWork` API so the user can specify upstrategy for each resource manifest in the `ManifestWork`. Possible strategy includes:
+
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+	// serverSideApply defines the configuration for server side apply. It is honored only when
+	// type of updateStrategy is ServerSideApply
+	// +optional
+	ServerSideApply *ServerSideApplyConfig `json:"serverSideApply,omitempty"`
+}
+
+type ServerSideApplyConfig struct {
+	// Force represents to force apply the manifest.
+	// +optional
+	Force bool `json:"force"`
```
yue9944882, line 91: do we need to clarify the field-manager here?

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+example, if there is an HPA controller on the spoke that take the ownership of the `spec.replicas` field. The condition of `ManifestWork`
+should show a status condition of applied failed as below:
+
+```yaml
+status:
+  resourceStatus:
+    manifests:
+      - conditions:
+        - lastTransitionTime: "2021-10-14T14:59:09Z"
+          message: Manifest ownership conflict
```
yue9944882, line 148: we can probably reflect the name of the conflicting field-manager in the status message for clarity?

yue9944882: additionally, we can provide a better message to those users whose managed cluster doesnt support SSA. this condition can be detected by parsing the returned error after sending SSA requests.

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
yue9944882: and are users expected to flip the "force" bit to false after the force apply worked for one time?

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+```go
+type UpdateStrategy struct {
+	// type defines the strategy to update this manifest, default value is Update.
+	// Update type means to update resource by an update call.
+	// None type means do not update resource based on current manifest.
+	// ServerSideApply type means to update resource using server side apply as fieldManager of work-controller.
+	// If there is conflict, the related Applied confition of manifest will in the status of False with the
+	// reason of ApplyConflict.
+	// +kubebuilder:default=60
```
yue9944882, line 76: nit: default=Update

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
qiujian16: the field-manager will be `work-controller` or `work-agent`, I do not think of a use case that use needs to set another field manager yet. Thought?

yue9944882: >  I do not think of a use case that use needs to set another field manager yet.

agree, i just think it would be better if we can reflect the default value in the api object

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+## Motivation
+
+Provide an interface for user to control how resource defined in `ManifestWork` should be updated.
+
+## Proposal
+
+We propose to update `ManifestWork` API so the user can specify upstrategy for each resource manifest in the `ManifestWork`. Possible strategy includes:
+
+1. by default, work-agent update resource as what is implemented today.
+2. user can choose not to update the resource, work-agent will only check the existence of the resource, but not keep its spec update.
```
deads2k, line 58: create-only resources seem unusual for the purpose of manifestwork

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+3. user can choose to use server side apply when work-agent update the resource.
+4. to support lowner version of kube, we may also cosider allowing user to choose json merge approach to update the resource.
```
deads2k, line 60: s/lowner/older/ ?

I think jsonmerge is likely to be fraught.

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+Provide an interface for user to control how resource defined in `ManifestWork` should be updated.
+
+## Proposal
+
+We propose to update `ManifestWork` API so the user can specify upstrategy for each resource manifest in the `ManifestWork`. Possible strategy includes:
+
+1. by default, work-agent update resource as what is implemented today.
+2. user can choose not to update the resource, work-agent will only check the existence of the resource, but not keep its spec update.
+3. user can choose to use server side apply when work-agent update the resource.
```
deads2k, line 59: This seems like a good option.

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,223 @@
...
+4. to support lowner version of kube, we may also cosider allowing user to choose json merge approach to update the resource.
+
+### Design Details
+
+#### API change
+
+We could add an updateStraty for each manifest as below:
```
deads2k, line 66:
```suggestion
We could add an updateStrategy for each manifest as below:
```

deads2k:
I like the idea presented here.  I think CreateOnly is more descriptive than None.  And I'd like to be sure that has a use-case behind it as opposed to just "this might be handy".  That appears to be splitting ownership between a spoke admin and a hubuser.

enhancements/sig-architecture/47-manifestwork-updatestrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+Provide an interface for user to control how resource defined in `ManifestWork` should be updated.
+
+## Proposal
+
+We propose to update `ManifestWork` API so the user can specify upstrategy for each resource manifest in the `ManifestWork`. Possible strategy includes:
+
+1. by default, work-agent update resource as what is implemented today.
+2. user can choose not to update the resource, work-agent will only check the existence of the resource, but not keep its spec update.
+3. user can choose to use server side apply when work-agent update the resource.
+4. to support orlder version of kube, we may also cosider allowing user to choose json merge approach to update the resource.
```
deads2k, line 60:
```suggestion
4. to support older version of kube, we may also cosider allowing user to choose json merge approach to update the resource.
```

deads2k:
I'm still interested in the use-case for createOnly, but the design looks good to me.

qiujian16:
> I'm still interested in the use-case for createOnly, but the design looks good to me.

add more clarification on when create-only should be used.

yue9944882 (commented):
LGTM

deads2k:
>add more clarification on when create-only should be used.

Thanks.  Be sure the godoc on the API itself provides similar help for users to choose a good value for themselves.

/lgtm

PR #51, opened 2022-04-11
KEP: Work Placement

jnpacker:
Signed-off-by: Joshua Packer <jpacker@redhat.com>
* Initial submission for Work placement enhancement

### DIFF
--- a/enhancements/sig-architecture/35-work-placement/work-placement.md
+++ b/enhancements/sig-architecture/35-work-placement/work-placement.md
@@ -66,12 +66,12 @@ Work placement should contain enough status information to determine if the mani
 #### API
 ```yaml
 apiversion: work.open-cluster-management.io/v1alpha1
-kind: PlacementWork # ApplicationWork?
+kind: PlacementManifestWork # ApplicationWork?
 metadata:
   name: app
   namespace: default
 spec:
-  manifestWorkSpec:
+  manifestWorkTemplate:
     workloads:
       manifest:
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/35-work-placement/work-placement.md
```diff
@@ -0,0 +1,192 @@
...
+          namespace: default
+        spec: … 
+      - apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: test02
+          namespace: default
+        spec: … 
+  ## The following is a spike ##
+  distributionStrategy:
```
serngawy, line 90: can we elaborate more on the distributionStrategy ? what is the type Split ? and why is the distributionStrategy is repeated under test02 item as the Type is split ?

ivan-cai:
Should we need different configuration across clusters? For example, we want to distribute some objects like ResourceQuota/LimitRanger to multiple clusters，but the object value is different in different clusters.

jnpacker:
That could definitely be a spike after the initial merge.  The API is done I will submit the PR this week. I hopefully will also port the controller to the ManifestWork repo (currently in its own) this week as well to get some feedback.

enhancements/sig-architecture/35-work-placement/work-placement.md
jnpacker: Once we have the base controller, we can start working on the strategy spike.

Issue #54, opened 2022-04-22
Policy repository to assist in creating/maintaining Policy Controllers


JustinKuli:
Enhancement connected to: https://github.com/open-cluster-management-io/community/issues/150

Issue #43, opened 2022-03-11
Governance addon controller


dhaiducek:
Related issue:
- https://github.com/open-cluster-management-io/community/issues/122

PR #55, opened 2022-04-22
Enhancement proposal: Core Policy Module

JustinKuli:
Related issues:
- https://github.com/open-cluster-management-io/enhancements/issues/54
- https://github.com/open-cluster-management-io/community/issues/150


enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,157 @@
...
+pull RBAC permissions on methods into the project. So RBAC requirements on
+common functions (eg `get namespaces` for selecting namespaces) will need to
+be well-documented.
+
+### Risks and Mitigation
+
+(unknown)
+
+## Design Details
+
```
qiujian16, line 79: I think it needs more details on what you would like to put in this module. Some examples will be helpful

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,157 @@
...
+Create a new repository in open-cluster-mangement-io to provide implementations
+of common APIs and functions for controllers to work with the policy framework.
+
+### User Stories
+
+#### Story 1
+
+As a developer of a policy controller, I can ensure consistent behavior with
+other policies by incorporating the common API types.
```
qiujian16, line 53: can you explain more on this? what consistent behavior you are expecting?

qiujian16 (commented):
I think the still need more details on what we really plan to do.

1. is it a duck type with some validation func? The reasoning of these validation
2. Is it some common helper funcs? What helper funcs we are expecting?

I would also want some example like how a developer can use this library to build the controller.

qiujian16:
we should also consider other alternatives in the proposal like:
1. should we have a common sdk library but not only for policy only? Will it be suitable in this use case?
2. can we build the helpers in the existing repo? we should clarify why a separated repo is better.

JustinKuli:
Thanks for the comments @qiujian16, you make some good points. I've brough in more design details from my prototype into the design section here, but I'll also try to address each point you made in this comment.

> is it a duck type with some validation func? The reasoning of these validation

Part of the intent is for a golang struct that new policy types would embed/inline, like most CRDs do with `metav1.TypeMeta`. Fields inside the struct would be the ones required or used by the policy framework.

For example, we can look at the [ConfigurationPolicy type definition](https://github.com/open-cluster-management-io/config-policy-controller/blob/main/api/v1/configurationpolicy_types.go), and the policy types in stolostron: [CertificatePolicy](https://github.com/stolostron/cert-policy-controller/blob/main/api/v1/certificatepolicy_types.go) and [IamPolicy](https://github.com/stolostron/iam-policy-controller/blob/main/api/v1/iampolicy_types.go). All of these re-define `type RemediationAction string`, and add kubebuilder validation. But if one of them drifts from what the policy-framework expects, they could run into trouble. Actually we just ran into an issue like this where one would not allow `enforce`, but the template-sync was still trying to create CRs with that specified, causing errors.

There are other fields like this that are in some ways required by the policy framework, but there are also pieces that are more like "best practices" that we should keep track of. For example: the framework has no hard requirements for the format or content of policy status fields, but if they were consistent, then other tools could read them nicely (eg a common UI).

> Is it some common helper funcs? What helper funcs we are expecting?

Some of the fields in the common struct should have some common behavior that each policy controller should not need to re-create / copy. For example, the `namespaceSelector`. Currently each controller has a function to parse the fiels and find the selected namespaces, like [this](https://github.com/open-cluster-management-io/config-policy-controller/blob/main/controllers/configurationpolicy_controller.go#L1387). The implementations have diverged slightly, which could be confusing in the future.

We have also discussed adding a label selector to this field, in addition to the current `includes` and `excludes` lists, and currently each controller would need to copy that in order to stay consistent. So a common place for this behavior would help add new features in the future.

> I would also want some example like how a developer can use this library to build the controller.

I think an example would be good, but I wasn't sure if I should include it in the proposal. If you'd like, I can move the tutorial out of the "non-goals".

> should we have a common sdk library but not only for policy only? Will it be suitable in this use case?

We currently have the [common library for addons](https://github.com/open-cluster-management-io/addon-framework), is that the kind of thing you mean? I imagined eventually using that library in this one, since authors of policy controllers will want a way to deploy and manage the controllers. It just wasn't an initial goal of mine.

I think the policy components are separate enough that it could be confusing to include the policy library in the addon library. Like how we don't currently have the policy APIs in the [api repo](https://github.com/open-cluster-management-io/api), although maybe that could be a separate discussion.

> can we build the helpers in the existing repo? we should clarify why a separated repo is better.

The current policy repostories are: [spec-sync](https://github.com/open-cluster-management-io/governance-policy-spec-sync), [status-sync](https://github.com/open-cluster-management-io/governance-policy-status-sync), [template-sync](https://github.com/open-cluster-management-io/governance-policy-template-sync), and [propagator](https://github.com/open-cluster-management-io/governance-policy-propagator) (I think the current addon-controller repo is a placeholder for https://github.com/open-cluster-management-io/enhancements/issues/43). Each of these is a controller, responsible for specific behavior in the policy framework, but none of them describe the entire framework. In stolostron, we have a [governance-policy-framework repo](https://github.com/stolostron/governance-policy-framework), which describes the overall behavior, and some of the API, but we don't have that in this community yet.

I feel like a new repository in this community is needed to explain the policy framework overall, and this library would be able to do that.

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+new controllers, and assist in maintenance of existing ones. This repository
+will also describe the policy-framework behavior for developers, since the
+website/docs are currently more user-focused.
+
+## Motivation
+
+Current policy controllers rely on detailed knowledge from the policy framework 
+implementation, that is not clear to new developers. Most notably, the format
+for events to be picked up by the policy framework is not well documented. Other
+details will be more consistent if maintained in a single place.x
```
mprahl, line 24:
```suggestion
details will be more consistent if maintained in a single place.
```

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+all policy types so they can be passed down from the parent policy. Putting
+these in a struct that policy types embed will ensure they are implemented.
+1. Common methods: behavior like selecting namespaces and emitting compliance
+events for the policy framework will have a common implementation. This will
+reduce bugs from inconsistency, and allow policy controllers to focus on their
+specific behavior, not the details of interacting with the framework.
+
+### Non-Goals
+
+1. Custom Controller Tutorial: it will be easier to create custom controllers,
```
mprahl, line 40: I would like this to be a goal.

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+common functions (eg `get namespaces` for selecting namespaces) will need to
+be well-documented.
+
+### Risks and Mitigation
+
+(unknown)
+
+## Design Details
+
+Here is an part of the common struct and validation from a
```
mprahl, line 88:
```suggestion
Here is a part of the common struct and validation from a
```

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+At least one additional function would cover creating the compliance event that
+the status-sync component watches for.
+
+### Open Questions
+
+1. What policy features should be included? Types / behavior required by the
+policy framework should be included, but what about things that aren't required,
+like status fields on policy types?
```
mprahl, line 146: This would be good as a separate struct to embed.

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+creating the events that it is watching for, and the template-sync could expose
+a type with the common fields it requires like `remediationAction`. In this
+case, the policy controllers would need to import each piece of the framework,
+and keep each updated to match each other if changes were made. There would
+also not be a clear place for non-required, but commonly used fields like the
+`spec.namespaceSelector`, and their behavior.
+
+## Infrastructure Needed
+
+A new repository, name TBD. Maybe `policy-framework`?
```
mprahl, line 227: Let's not take that name in case we upstream the stolostron policy-framework repo.

How about `policy-controller-runtime` or `policy-controller-core`?

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+A common repository / module will contain common API validation, structs, and
+methods that multiple policy controllers can use. This will simplify creating
+new controllers, and assist in maintenance of existing ones. This repository
+will also describe the policy-framework behavior for developers, since the
```
mprahl, line 16: Keep this scoped to generic statements about policy controllers. Anything more in depth can be on the website a page focused about the architecture of the Policy Addon.

enhancements/sig-policy/54-core-policy-module/README.md
JustinKuli: The stolostron policy-framework repo is (almost entirely) tests and documentation. IMO, we could merge that content into this community repo if we wanted to "upstream" it.

I would prefer something without "core" - that would lead to stuttering in the type names. For example, in the current proposal I have `policycore.PolicyCoreSpec`, which I don't like. Something like `framework.PolicyCoreSpec` would make me happier. Alternatively, if we changed the name of `PolicyCoreSpec`, that could help.

Both `policy-controller-runtime` and `policy-controller-core` feel too close to well-known k8s libraries `sigs.k8s.io/controller-runtime` and `k8s.io/api/core/v1`.

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Design Details
+
+Here is a part of the common struct and validation from a
+[prototype](https://github.com/JustinKuli/policycore-test/blob/main/api/v1/policycore_types.go)
+of this proposal:
+
+```golang
+type PolicyCoreSpec struct {
+	Severity          Severity          `json:"severity,omitempty"`
```
qiujian16, line 101: It is necessary to doc the behavior of each field here. And I am not sure how to ensure two different customized policy controllers behave the same for these fields.

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+Common behavior could be implemented in the common library through methods on 
+these new types, for example, utilizing the `spec.namespaceSelector` field could
+be a method with this signature and doc:
+
+```golang
+// GetNamespaces fetches all namespaces in the cluster and returns a list of the
+// namespaces that match the NamespaceSelector. The client.Reader needs access
+// for viewing namespaces, like the access given by this kubebuilder tag:
+// `//+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch`
+func (sel NamespaceSelector) GetNamespaces(ctx context.Context, r client.Reader) ([]string, error) {
```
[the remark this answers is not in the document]
JustinKuli: No 😓 , the existing controllers use an object with `Include` and `Exclude` lists, each of which supports (some) wildcards, so you might have something like:
```yaml
spec:
  namespaceSelector:
    include: ["*"]
    exclude:
      - "kube-*"
      - "openshift*"
```

See doc in https://github.com/open-cluster-management-io/config-policy-controller and the implementation in https://github.com/open-cluster-management-io/config-policy-controller/blob/main/pkg/common/namespace_selection.go.

Improving the namespace selector to work with a `metav1.ListOptions{LabelSelector: ...}`, is one of my personal motivations for this proposal. But it would need to be backwards-compatible with what we already have.

enhancements/sig-policy/54-core-policy-module/README.md
JustinKuli: I agree that would be documented in the actual module. Should I put all those details in the proposal? I agree they could use more explanation, but I think we would start from what is currently documented in the configuration policy: https://github.com/open-cluster-management-io/config-policy-controller/blob/main/api/v1/configurationpolicy_types.go

For consistent behavior, for some things I think we can only document. For example, the severity field is currently just informational in the controllers - other things like alerting systems could use it but it shouldn't change the behavior of the policy. But NamespaceSelector would have a method that all the controllers can use.

enhancements/sig-policy/54-core-policy-module/README.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+With this struct, a new policy controller could implement their type like this,
+in order to get the same validation automatically:
+
+```golang
+import policycore "github.com/open-cluster-management-io/policycore/api/v1" // example import - this repo does not exist
+
+type MyNewPolicySpec struct {
+    policycore.PolicyCoreSpec `json:",inline"`
```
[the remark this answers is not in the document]
JustinKuli: I think the fields need to be directly under `spec` in JSON/YAML - I'm not sure if that is possible if it's nested in another struct? Maybe if they both use `json:",inline"`? But other controllers treat these as unstructured, and look for for things at a certain spot - for example the template-sync controller overrides remediationAction like this: https://github.com/stolostron/governance-policy-template-sync/blob/f4ec854ea43564a8c5cbfed5d2d87b7eebeb862a/controllers/template_sync.go#L327

I don't think we're planning a validation admission hook, just the CRD validations that come from using these types.

enhancements/sig-policy/54-core-policy-module/README.md
qiujian16: I think if there are two controllers that embed this spec in their own spec. how they behave on these fields is questionable. I wonder if we will have a common agent that consumes any customized policy and consumes field e.g severity. Maybe some description here will be helpful.

JustinKuli:
Added some more details to the example struct, although I would expect the exact wordings to change in the actual library.

From some other discussions a new name for the repository came up as well: "policy-nucleus", capturing the idea of a central/core piece that doesn't conflict with existing "core" or "framework" names.

@dhaiducek @mprahl what do you think?

mprahl (approved):

mprahl:
> 

I'm fine with this repo name, but the final decision can be made outside of this enhancement proposal.

PR #56, opened 2022-04-25
enhancement proposal of cluster deletion

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>

### DIFF
--- a/enhancements/sig-architecture/11-cluster-deletion/README.md
+++ b/enhancements/sig-architecture/11-cluster-deletion/README.md
@@ -1,0 +1,123 @@
+# Enhancement of cluster deletion 
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+Currently, when we remove a cluster from the Hub, the `managedClusterAddons` and `manifestWorks` in the cluster namespace need to be deleted manually, and cannot ensure that the resources deployed on the Spoke cluster can be cleaned up. This proposal is to define clearly what happened when a cluster is deleted.
+
+## Motivation
+
+Today, when we delete a `managedCluster`, only the `rolebinding` for the registration/work agent is cleaned up on the hub, and the resources including the cluster namespace,`managedClusterAddons` and `manifestWorks` are left over. It will bring some issues like: 
+1. We cannot delete `manifestWork` directly because the finalizer cannot be removed by the work-agent.
+2. The resources deployed by `manifestWork` on the spoke cluster cannot be cleaned up even if its `manifestWork` is deleted.
+3. The `managedClusterAddon` with pre-delete job cannot be deleted because the `manifestWorks` cannot be deleted too.
+
+There is a typical use case that the lifecycle of the `klusterlet` on the spoke cluster is taken over by the `manifestWork` too. There is a `manifestWork` to deploy klusterlet manifests itself after the cluster is registered. And the klusterlet manifests of the managed cluster will be cleaned up if the klusterlet manifestWork is deleted on the hub.
+The cleanup starts when the `managedCluster` is deleting.
+
+1. Delete the `managedClusterAddons` and `manifestWorks` firstly except the `klusterlet manifestWork`.
+2. Delete the `klusterlet manifestWork` after the other `managedClusterAddons` and `manifestWorks` are deleted.
+3. Start to delete the cluster namespace when there is no `managedClusterAddons` and `manifestWorks`. The cluster ns will be deleted after the resources related with Hive like job, clusterDeployment, etc are deleted. Because the Hive will create some jobs in this ns when the provisioned cluster is destroyed.
+
+The deletion process is a typical use case and example for the community. This proposal is to enhance and standardize the cluster deletion process in OCM.
+
+### Goals
+
+- Enhance and standardize the cluster deletion process and clean up the resources including `managedClusterAddons`, `manifestWorks` created by addon-framework    and namespace in order.
+
+### Non-Goals
+
+- Clean up the `manifestWorks` which are not created by addon-framework automatically.
+
+   These customized `manifestWorks` should be cleaned up by other controllers, because there are cases like these `manifestWorks` could be evicted and rescheduled to another `managedCluster` when the `managedCluster` is deleting, or they have pre-delete jobs to do.
+
+## Proposal
+
+### Design Details 
+
+Add a new feature gate named `ClusterLifeCycle` for this cluster deletion feature, which is disabled by default.
+#### registration-controller 
+
+1. Add a condition type `ContentDeleteSuccess` to the `managedCluster`, which indicates whether the cluster resources are remaining during the `managedCluster` deletion process.
+
+2. Define a label named `cluster.open-cluster-management.io/keep-upon-cluster-deletion`. The cluster namespace with this label will not be deleted by the registration during the `managedCluster` deletion process. 
+   
+3. Add options for the registration-controller binary. The registration-controller will handle the deletion process.
+   ```
+      /registrion
+      -- controller
+      -- resources-to-monitor-for-deletion: resource.version.group,resource.version.group.com...
+   ```
+
+   * `resources-to-monitor-for-deletion` specifies a list of resources GVR. registration-controller will not delete anything until all resources defined here are deleted by other controller. The common format of string like either `resource.version.group` or `resource.version.group.com`. 
+
+4. Add a managedCluster deletion controller in registration-controller to handle the resource cleanup. The deletion process like this:
+   
+   1. The cluster admin deletes the `managedCluster`. The `managedCluster` is under deleting since there is a cleanup finalizer on it.
+   2. `deletion-controller` begins to monitor predeleted resources defined in `resources-to-monitor-for-deletion` at first, do not delete anything until all resources defined are cleaned up by other controller.
+   3. `deletion-controller` begins to delete all `managedClusterAddons`.
+   4. `deletion-controller` begins to delete `roleBindings` of registration-agent and work-agent after all `managedClusterAddons` and `manifestWorks` are cleaned up.
+   5. `deletion-controller` begins to delete the cluster namespace without the label `cluster.open-cluster-management.io/keep-upon-cluster-deletion`.
+   6. `deletion-controller` remove the finalizer `cluster.open-cluster-management.io/api-resource-cleanup` on the `managedCluster`.
+
+#### manifestworkvalidators-webhook 
+
+1. manifestworkvalidators webhook will deny the creation of all `managedClusterAddons` when the `managedCluster` is deleting.
+
+2. manifestworkvalidators webhook will deny the creation of all `managedClusterAddons` and `manifestWorks` when there is no `managedCluster` with the same name as the namespace.
+
+### Upgrade strategy 
+   
+1. We will put this new feature into a feature gates named `ClusterLifeCycle` which is disabled by default.
+
+2. The user should add the label `cluster.open-cluster-management.io/keep-upon-cluster-deletion` to the cluster namespace which do not want to be deleted by registration.
+
+### User Stories
+
+#### Story 1
+
+As a cluster admin, the cluster namespace should be deleted after I delete the `managedCluster`.
+
+   Deletion process:
+   1. The cluster admin deletes the `managedCluster` on the hub.
+   2. `deletion-controller` deletes all `managedClusterAddons`.
+   3. The cluster admin or other controller deletes all customized `manifestWorks`.
+   4. `deletion-controller` deletes `roleBindings` of registration-agent and work-agent after there is no `managedClusterAddons` and `manifestWorks`.
+   5. `deletion-controller` deletes the cluster namespace.
+   6. The cluster admin deletes the `klusterlet` on the spoke cluster.
+
+#### Story 2
+
+As a cluster admin, the cluster namespace should not be deleted after I delete the `managedCluster`, because I want to remain other resources in this namespace.
+
+  Prerequisites:
+   1. Add the label `cluster.open-cluster-management.io/keep-upon-cluster-deletion` to the cluster namespace.
+   
+  Deletion process:
+   1. The cluster admin deletes the `managedCluster` on the hub.
+   2. `deletion-controller` deletes all `managedClusterAddons`.
+   3. The cluster admin or other controller deletes all customized `manifestWorks`.
+   4. `deletion-controller` deletes `roleBindings` of registration-agent and work-agent after there is no `managedClusterAddons` and `manifestWorks`.
+   5. The cluster admin deletes the `klusterlet` on the spoke cluster.
+
+#### Story 3
+
+As a cluster admin, the `managedClusterAddons` and cluster namespace should be deleted after I delete all applications in the cluster namespace during the cluster deletion.
+  Prerequisites:
+   1. Add the application GVR to `resources-to-monitor-for-deletion` of registration-controller.
+   
+  Deletion process:
+   1. The cluster admin deletes the `managedCluster` on the hub.
+   2. The cluster admin or other controller deletes applications in the cluster namespace.
+   3. `deletion-controller` deletes all `managedClusterAddons` after all applications are deleted.
+   4. The cluster admin or other controller deletes all customized `manifestWorks`.
+   5. `deletion-controller` deletes `roleBindings` of registration-agent and work-agent after there is no `managedClusterAddons` and `manifestWorks`.
+   6. `deletion-controller` deletes the cluster namespace.
+   7. The cluster admin deletes the `klusterlet` on the spoke cluster.
--- a/enhancements/sig-architecture/11-cluster-deletion/metadata.yaml
+++ b/enhancements/sig-architecture/11-cluster-deletion/metadata.yaml
@@ -1,0 +1,14 @@
+title: enhancement-of-cluster-deletion
+authors:
+  - "@zhiweiyin318"
+reviewers:
+  - "@qiujian16"
+  - "@yue9944882"
+  - "@skeeey"
+approvers:
+  - "@qiujian16"
+  - "@yue9944882"
+creation-date: 2022-04-25
+last-updated: 2022-04-25
+status: provisional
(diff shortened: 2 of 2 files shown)


enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. The resources deployed by the `manifestWorks` should be cleaned up by OCM. 
+2. The resources deployed by other owners can be cleaned up optionally by their owners like the subscriptions deployed by application addon.
+3. The `Klusterlet` CR should be deleted.
+
+### The behaviors when a cluster is removed from the Hub.
+
+1. The first step to remove a cluster is to delete the `managedCluster` on the Hub.
+2. Delete the `managedClusterAddons` and `manifestWorks` in the cluster namespace.
+3. Delete the `roleBindings` of registration-agent and work-agent in the cluster namespace after all of `managedClusterAddons` and `manifestWorks` are deleted.
+4. Delete the cluster namespace.
```
qiujian16, line 48: we do not delete cluster namespace today. Do we need to do this? Or we should by default delete ns and allow user to choose not to delete ns?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,59 @@
...
+5. Delete the `Klusterlet` CR on the Spoke cluster.
```
qiujian16, line 49: How can you prevent user deleting klusterlet at first? Or what should we do when klusterlet is deleted at first?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,59 @@
...
+On the Spoke cluster
+
+1. The resources deployed by the `manifestWorks` should be cleaned up by OCM. 
+2. The resources deployed by other owners can be cleaned up optionally by their owners like the subscriptions deployed by application addon.
+3. The `Klusterlet` CR should be deleted.
+
+### The behaviors when a cluster is removed from the Hub.
+
+1. The first step to remove a cluster is to delete the `managedCluster` on the Hub.
+2. Delete the `managedClusterAddons` and `manifestWorks` in the cluster namespace.
```
qiujian16, line 46: this needs more consideration. 

1. what should we do when cluster status is unknown. should we remove addon/work in this scenario?
2. should we define a grace period to remove these resources?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,59 @@
...
+3. Delete the `roleBindings` of registration-agent and work-agent in the cluster namespace after all of `managedClusterAddons` and `manifestWorks` are deleted.
+4. Delete the cluster namespace.
+5. Delete the `Klusterlet` CR on the Spoke cluster.
+
+### Enhancements
+
+1. Prevent the `managedClusterAddon` and `manifestWork` from being created for the `managedCluster` when the `managedCluster` is deleted in the work-webhook.
+2. Add a `unjoining` type for the `ManagedClusterJoined` condition of `managedCluster` to indicate the deleting status in registration.
+3. Delete the `managedClusterAddons` and `manifestWorks` when the `managedCluster` is deleting in work.
+4. Delete the cluster namespace until there is no `managedClusterAddons` and `manifestWorks` in it when the `managedCluster` is deleting in registration-operator.
```
qiujian16, line 56: This need further consideration, we cannot directly delete ns in some cases.

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,59 @@
...
+2. Delete the `managedClusterAddons` and `manifestWorks` in the cluster namespace.
+3. Delete the `roleBindings` of registration-agent and work-agent in the cluster namespace after all of `managedClusterAddons` and `manifestWorks` are deleted.
+4. Delete the cluster namespace.
+5. Delete the `Klusterlet` CR on the Spoke cluster.
+
+### Enhancements
+
+1. Prevent the `managedClusterAddon` and `manifestWork` from being created for the `managedCluster` when the `managedCluster` is deleted in the work-webhook.
+2. Add a `unjoining` type for the `ManagedClusterJoined` condition of `managedCluster` to indicate the deleting status in registration.
+3. Delete the `managedClusterAddons` and `manifestWorks` when the `managedCluster` is deleting in work.
```
qiujian16, line 55: I think we need a more detailed flow: which actor under which condition does what action.

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. The first step to remove a cluster is to delete the `managedCluster` on the Hub.
+2. Delete the `managedClusterAddons` and `manifestWorks` in the cluster namespace.
+3. Delete the `roleBindings` of registration-agent and work-agent in the cluster namespace after all of `managedClusterAddons` and `manifestWorks` are deleted.
+4. Delete the cluster namespace.
+5. Delete the `Klusterlet` CR on the Spoke cluster.
+
+### Enhancements
+
+1. Prevent the `managedClusterAddon` and `manifestWork` from being created for the `managedCluster` when the `managedCluster` is deleted in the work-webhook.
+2. Add a `unjoining` type for the `ManagedClusterJoined` condition of `managedCluster` to indicate the deleting status in registration.
```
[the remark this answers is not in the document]
zhiweiyin318: following the status of namespace/pod, we could add a `phase` filed in the status of managedCluster, which is the current lifecycle phase of managedCluster.
`Pending` phase means the managedCluster is waiting to be joined.
`Joined` phase means the managedCluster is joined.
`Terminating` phase means the Terminating is being deleted.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: I think we can force delete all managedClusterAddons and manifestWorks 10 minutes after the managedCluster is unavaliable.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: no, we cannot prevent user deleting klusterlet at first.
I think we could do force deletion until the cluster is unavailable if klusterlet is deleted at first.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: considering to add a `deleteOption` filed in the spec of managedCluster.
Set `deleteAll` to delete the cluster namespace after the managedCluster is deleted.
Otherwise, the cluster namespace will not be deleted by default.

and define an annotation `open-cluster-management.io/delete-by-user`. The managedClusterAddons and manifestWorks with this annotation will not be deleted by OCM during the deletion process.

enhancements/sig-architecture/11-cluster-deletion/README.md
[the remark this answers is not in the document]
qiujian16: why it is a phase but not another status condition?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Summary
+
+Currently, when we remove a cluster from the Hub, the `managedClusterAddons` and `manifestWorks` in the cluster namespace need to be deleted manually, and cannot ensure that the resources deployed on the Spoke cluster can be cleaned up. This proposal is to define clearly what happened when a cluster is deleted.
+
+## Motivation
+
+### Goals
+
+- Define which resources related the deleted cluster should be cleaned up by OCM.
```
qiujian16, line 19: rephrase?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,86 @@
...
+## Design Details 
+
+1. Add a `deleteOption` filed in the spec of `managedCluster`. 
+   1. With the value `deleteAll`, means to delete the cluster namespace after the `managedCluster` is deleted. 
+   2. Otherwise, the cluster namespace will not be deleted by default.
+2. Add a `phase` filed in the status of `managedCluster`, which is the current lifecycle phase of `managedCluster`. 
+   1. `Pending` phase means the `managedCluster` is waiting to be joined.
+   2. `Joined` phase means the `managedCluster` is joined.
+   3. `Terminating` phase means the `Terminating` is being deleted.
+3. Define an annotation `open-cluster-management.io/delete-by-user`. The `managedClusterAddons` and `manifestWorks` with this annotation will not be deleted by OCM during the deletion process.
```
qiujian16, line 62: what is the intention of this?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,86 @@
...
+#### Story 2
+
+As a cluster admin, the cluster namespace and some specified `manifestWorks` should not be deleted after I delete the `managedCluster`, because I want to delete some resources in the cluster namespace by myself.
+
+## Design Details 
+
+1. Add a `deleteOption` filed in the spec of `managedCluster`. 
+   1. With the value `deleteAll`, means to delete the cluster namespace after the `managedCluster` is deleted. 
+   2. Otherwise, the cluster namespace will not be deleted by default.
+2. Add a `phase` filed in the status of `managedCluster`, which is the current lifecycle phase of `managedCluster`.
```
qiujian16, line 58: phase should only be use for user interface

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,86 @@
...
+   2. Otherwise, the cluster namespace will not be deleted by default.
+2. Add a `phase` filed in the status of `managedCluster`, which is the current lifecycle phase of `managedCluster`. 
+   1. `Pending` phase means the `managedCluster` is waiting to be joined.
+   2. `Joined` phase means the `managedCluster` is joined.
+   3. `Terminating` phase means the `Terminating` is being deleted.
+3. Define an annotation `open-cluster-management.io/delete-by-user`. The `managedClusterAddons` and `manifestWorks` with this annotation will not be deleted by OCM during the deletion process.
+
+### The behaviors when a cluster is removed from the Hub.
+
+The cluster admin deletes the `managedCluster` on the Hub.
```
qiujian16, line 66: we should consider two cases
1. user delete cluster at first on hub.
2. user remove klusterlet at first on managed cluster.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: for ACM scenario that we use manifestWork to controller klusterlet, we can add this annotation to the manifestWork of klusterlet to make it deleted after other manifestworks are deleted.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: > 2. user remove klusterlet at first on managed cluster.

if the managedCluster is not deleted after user removed klusterlet,  we should do nothing.
I think we should consider 1 case that the managedCluster is deleted or not.

enhancements/sig-architecture/11-cluster-deletion/README.md
qiujian16: I think we should consider defining a validation webhook, that work of klusterlet is a critical work that deletion should be denied until all work depending on this is removed.

enhancements/sig-architecture/11-cluster-deletion/README.md
qiujian16: you probably need to consider a grace period on hub when klusterlet is removed already, then do a force delete?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,122 @@
...
+   1. Add the label `cluster.open-cluster-management.io/delete-by-other` to the special `manifestWork`.
+   
+   Deletion process:
+   1. The cluster admin deletes the `managedCluster` on the hub.
+   2. `deletion-controller` deletes all `managedClusterAddons`.
+   3. `deletion-controller` deletes all `manifestWorks` without the label `cluster.open-cluster-management.io/delete-by-other`.
+   4. Other controller deletes the special `manifestWorks` after the other `manifestWorks` are deleted.
+   5. `deletion-controller` deletes `roleBindings` of registration-agent and work-agent.
+   6. `deletion-controller` force deletes the `managedClusterAddons` and `manifestWorks` after the `forceDeletionGracePeriodSeconds`.
+   7. `deletion-controller` deletes the cluster namespace.
```
qiujian16, line 122: please add upgrade strategy for this feature. Also we should put this in a feature gates.

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+4. Add a managedCluster Deletion controller in registration-controller to handle the resource cleanup. The deletion process like this:
+   
+   1. The cluster admin deletes the `managedCluster`. The `managedCluster` is under deleting since there is cleanup finalizer on it.
+   2. `deletion-controller` begins to monitor predeleted resources defined in `resources-to-monitor-for-deletion` at first, do not delete anything until all resources defined are cleaned up by other controller.
+   3. `deletion-controller` begins to delete all `managedClusterAddons`.
+   4. `deletion-controller` begins to delete all `manifestWorks` without the label `cluster.open-cluster-management.io/delete-by-other` after all `managedClusterAddons` are deleted.
+   5. `deletion-controller` begins to delete `roleBindings` of registration-agent and work-agent.
+   6. `deletion-controller` begins to delete cluster namespace without the label `cluster.open-cluster-management.io/delete-by-other` after all `manifestWorks` are deleted.
```
qiujian16, line 67: maybe cluster.open-cluster-management.io/keep-upon-cluster-deletion

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+1. Add a condition type `ContentDeleteSuccess` to the `managedCluster`, which indicates whether the cluster resources are remaining during the `managedCluster` deletion process.
+
+2. Define a label named `cluster.open-cluster-management.io/delete-by-other`. The cluster resources with this label will not be deleted by the ACM during the `managedCluster` deletion process.
+   
+3. Add options for the registration-controller binary. The registration-controller will handle the deletion process.
+   ```
+      /registrion
+      -- controller
+      -- forceDeletionGracePeriodSeconds: 600
```
qiujian16, line 52: why do we need this force delete?

enhancements/sig-architecture/11-cluster-deletion/README.md
```diff
@@ -0,0 +1,122 @@
...
+   7. `deletion-controller` begins to force delete the `managedClusterAddons`,`manifestWorks` and cluster namespace without the label `cluster.open-cluster-management.io/delete-by-other` after the `forceDeletionGracePeriodSeconds`.
+   8. `deletion-controller` remove the finalizer `cluster.open-cluster-management.io/api-resource-cleanup` on the `managedCluster`.
+
+5. We only guarantee that all `managedClusterAddons`, `manifestWorks` and `rbac` in the cluster namespace are deleted. The user should clean up the remaining resources in the cluster namespace manually if the namespace is failed to be deleted because of the contents remaining.
+
+
+#### manifestworkvalidators-webhook 
+
+manifestworkvalidators webhook will deny the creation of all `managedClusterAddons` and `manifestWorks` except addon `manifestWorks` when the `managedCluster` is deleting. Because `registration-controller` starts the resources' deletion process in cluster namespace when the `managedCluster` is deleting. But the pre-delete addon `manifestWork` is an exception since it is created when the `managedClusterAddon` is deleted.
```
qiujian16, line 76: these type of manifestworks may need another label. Also what if a cluster is not found?

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: done

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: if the spoke cluster is destroyed before start to delete the managedCluster, the manifestWorks cannot be deleted because there is on one to remove the finalizer on them.
in this case, we need force delete the manifestWorks when the cluster is unavailable for a grace period, like 10min.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: how about add a label `cluster.open-cluster-management.io/created-by-addon` for the addon manifestWorks created by add-framework ?

added the case that webhook will deny the creation of all `managedClusterAddons` and `manifestWorks` when there is no `managedCluster` with the same name as the namespace.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: done

enhancements/sig-architecture/11-cluster-deletion/README.md
qiujian16: I do not think we can diff between disconnected or cluster is down. Registration may not fit to handle this case, it may need higher level scheduler to move the workload. I am thinking maybe we should not let registration to delete work, but only addon (addon is managed by registration). registration should only check if work is cleaned already.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: 1. manifestworkvalidators webhook will deny the creation of all `managedClusterAddons` when the `managedCluster` is deleting.

2. manifestworkvalidators webhook will deny the creation of all `managedClusterAddons` and `manifestWorks` when there is no `managedCluster` with the same name as the namespace.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: updated as we discussed.

enhancements/sig-architecture/11-cluster-deletion/README.md
zhiweiyin318: will not delete the cluster ns with the label `cluster.open-cluster-management.io/keep-upon-cluster-deletion`

PR #57, opened 2022-06-06
Selective policy enforcement

imiller0:
Proposal to create a mechanism through which an inform policy can be selectively enforced on a subset of the clusters it is bound to.


JustinKuli (commented):
I think I like the idea of using an annotation on the replicated policy in order to determine whether to update that policy's enforcement. That, and possibly having a separate controller to manage those annotations, seems like a good solution that can be adjusted based on users' needs.

But I feel like this proposal asserts that the only field that should ever be selectively rolled out is the `remediationAction`. What if users want to selectively roll out changes to policies, while keeping them in "enforce" mode?

Along those lines, is the new `remediationAction` value of "controlled" required? Or could this just be implemented via an annotation on replicated policies, and by adding logic to the propagator so that *any* updates to the parent policy are passed to the replicated policy only if the annotation allows it? The annotation might behave more like a "pause" feature in that case.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,415 @@
...
+policy until the user, controller, or higher level orchestration
+enables remediation directly on the child policy.
+
+The mechanism to enable enforcement of a child policy would be to add
+an annotation to the child policy which enables enforcement
+
+```
+metadata:
+  annotations:
+    versioned-policy-enable: [any|<generation>]
```
mprahl, line 106: Instead of an annotation, I think we could do something a little simpler. We could set the `remediationAction` to `controlled` on the parent policy and the replicated policies are created with the `remediationAction` of `inform` as you said.

Then the user would set the `remediationAction` to `enforce` on the subset of replicated policies as desired. The controller (governance-policy-propagator) that manages those replicated policies would ignore changes made to the `remediationAction` on the replicated policies when the parent policy's `remediationAction` is `controlled`.

If the replicated policy gets reevaluated and the spec changes (e.g. parent policy change or Hub policy template resolves differently), then the `remediationAction` gets reset to `inform`.

What do you think?

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,415 @@
...
+
+To improve user visibility the dashboard view of policies should be
+enhanced to show the current enforcement state of the policy. This
+would be reflected in the Clusters tab of the Policies/_policyName_
+page and the specific child policy view
+Policies/_policyName_/_cluster-policy-name_
+
+### User Stories
+
+Detail the things that people will be able to do if this is
```
mprahl, line 143: This paragraph is placeholder text and can be removed.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,415 @@
...
+   clusters which should receive the configuration update.
+1. My controller/orchestrator annotates the child policy for the
+   selected clusters with `versioned-policy-enable: <value>`
+1. The GRC policy controller remediates the policy on only the
+   selected clusters according to existing "enforce" mode rules. The
+   other bound clusters continue evaluating the policy under "inform"
+   mode rules.
+1. My controller recognizes the child policy moving to the compliant
+   state and removes the `versioned-policy-enable` annotation.
+1. The GRC policy controller reverts to existing "inform" mode
```
mprahl, line 167: My comment (https://github.com/open-cluster-management-io/enhancements/pull/57/files#r892335110) doesn't account for this. Perhaps a different `remediationAction` on the parent policy of `controlledInformOnCompliant` would suffice to capture this additional desired behavior.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
JustinKuli: Under the current proposal, I think the `remediationAction` on the replicated policies would be `controlled`, and the framework would infer whether to use `inform` or `enforce` based on the annotation. Is that right @imiller0?

I think the alternative mentioned here would lose out on the `any` value functionality. And I think it might be unclear to the user why the policy reverted to `inform` after they changed it - with the generation annotation and keeping the `remediationAction=controlled`, it might be more clear that something else in the spec has changed, not just that this feature is buggy or something.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,415 @@
...
+when an update is desired.
+
+The implementation is relatively simple and defers the “selection”
+flexibility to the user/orchestrator/controller
+
+To improve user visibility the dashboard view of policies should be
+enhanced to show the current enforcement state of the policy. This
+would be reflected in the Clusters tab of the Policies/_policyName_
+page and the specific child policy view
+Policies/_policyName_/_cluster-policy-name_
```
dhaiducek, line 139: This section isn't relevant for upstream since a UI hasn't been developed yet here.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
mprahl: > Under the current proposal, I think the `remediationAction` on the replicated policies would be `controlled`, and the framework would infer whether to use `inform` or `enforce` based on the annotation. Is that right @imiller0?

That's my understanding too.
> 
> I think the alternative mentioned here would lose out on the `any` value functionality. And I think it might be unclear to the user why the policy reverted to `inform` after they changed it - with the generation annotation and keeping the `remediationAction=controlled`, it might be more clear that something else in the spec has changed, not just that this feature is buggy or something.

That's a good point on the `any` functionality. I think a separate item in the spec of `remediationActionOptions` that could have the keys `informOnUpdate` and `informOnCompliant` (https://github.com/open-cluster-management-io/enhancements/pull/57#discussion_r892339771) would cover these cases.

I just think that an annotation that contains the generation is hard to use because each replicated policy has a different generation. The user also has to be aware of what a generation is which is an obscure Kubernetes concept in my opinion.  It also means that the behavior of resets is different per cluster, which is complex. We also don't have CRD validation on annotations without a validation webhook, which we don't currently have one.

JustinKuli: I don't think other, more specific remediationAction options would be better. I think it would add complexity at the wrong level. That's personal preference though.

I don't see a basic user trying to use a generation annotation, I think `any` would be more likely if this was being done manually. For a controller adding the annotation, I don't think the generation is an obscure feature, and the fact that it's different on each replicated policy is no problem.

I'm not sure I know what you mean by "the behavior of resets is different per cluster" ... the propagator will always update the spec of the replicated policy to match the root. There's no resetting, unless I misunderstand the proposal. The policy might only be *enforced* at a certain generation, but it will never be out of sync with the root policy.

You bring up a good point with the spoke cluster. I'm now imagining that the spec-sync is where the inform/enforce decision could be made. When it looks at the replicated policy on the hub, and if the remediationAction is `controlled`, then it can look at the annotation & generation and decide whether to set it to `inform` or `enforce`. (A self-managed hub, with no spec-sync, would need extra logic)

mprahl: > I don't think other, more specific remediationAction options would be better. I think it would add complexity at the wrong level. That's personal preference though.

This at least centralizes the complexity in the root policy rather than different behavior per replicated policy.
> 
> I don't see a basic user trying to use a generation annotation, I think `any` would be more likely if this was being done manually. For a controller adding the annotation, I don't think the generation is an obscure feature, and the fact that it's different on each replicated policy is no problem.

If a user is manually using this, I think they want the behavior of enforce until the policy changes. This is the most conservative approach and allows users to opt-in to enforcing on a per-cluster basis. The current proposal in the doc requires you to set the generation in the annotation to get this behavior.
> 
> I'm not sure I know what you mean by "the behavior of resets is different per cluster" ... the propagator will always update the spec of the replicated policy to match the root. There's no resetting, unless I misunderstand the proposal. The policy might only be _enforced_ at a certain generation, but it will never be out of sync with the root policy.

I mean that in the current proposal, a user must set the annotation per replicated policy, which means some can have generation set and some can have `any` set, which is complex to understand.
> 
> You bring up a good point with the spoke cluster. I'm now imagining that the spec-sync is where the inform/enforce decision could be made. When it looks at the replicated policy on the hub, and if the remediationAction is `controlled`, then it can look at the annotation & generation and decide whether to set it to `inform` or `enforce`. (A self-managed hub, with no spec-sync, would need extra logic)

Ideally, a UI or something could just look at the `remediationAction` on the replicated policy on the Hub to know the current state. Calculating the behavior based on the annotation matching the generation of the replicated policy is complex.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
imiller0: Removed.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
jc-rh: I like Matt's proposal. We didn't go this way assuming special ignore logic in the sync process was undesired. However, as Matt pointed out, managing different generation values on spokes ties us back to the sync process if we want to achieve consistent behavior.

The main motivation behind this proposal is to have the controlled remediation. The default behavior should be reset on change. We could leave the any case for future and add an annotation if we need to.

imiller0: Would this be a good summary of the proposal?
- remediationAction: [controlled | inform | enforce ]
- controlled will generate child policies with explicit: `remediationAction: inform`
- user/controller/orchestration can set `remediationAction: enforce` on any child policy
- Any change/update to the parent policy resets all child policies to `remediationAction: inform`
- [optional/future] annotation can be added to change the "reset" behavior

serngawy: I believe the main use-case here is for a policy. I want to have `remediation action as inform` for a group of clusters and for the same policy I want to have `remediation action as enforce` for another group of clusters. 

I suggest  it is  better to decouple (or allow override) the remediation Action within the policy definition not by modifying the child policies. To clarify this; As an end user I don’t see (create/update/delete) the child policies I only work with the parent policy. So by adding `remediation Action: controlled` it is not clear how we will let the end user select and modify the child policies that he/she wants to change ?

I believe we better expose the remediation action override API within the policy definition instead of delegating that to the child policies. Like could we have a `clusterSelector` within the policy definition that selects a subset of clusters (from the placement/placementRule’s clusters) And override the `remediation Action` for those clusters subset.

jc-rh: > Would this be a good summary of the proposal?
> 
>     * remediationAction: [controlled | inform | enforce ]
> 
>     * controlled will generate child policies with explicit: `remediationAction: inform`
> 
>     * user/controller/orchestration can set `remediationAction: enforce` on any child policy
> 
>     * Any change/update to the parent policy resets all child policies to `remediationAction: inform`
> 
>     * [optional/future] annotation can be added to change the "reset" behavior

Yes, this is exactly what I meant by "settle on an approach that ignores the sync status". My worry is that clusters could potentially get skipped due to the race between an update/sync and an action override, as they are done by two independent players. The user (or TALO) is not aware of the sync status. If there is a sync pending for that child policy, the action override at the child policy level gets reset right away. It may seem like a low runner case, but it makes me rather uncomfortable with this approach.

Initially both Ian and I were trying to avoid another level of grouping/placement/selector kind of approach. Now I feel it seems to be inevitable if consistency across spokes for one rollout process is a strong requirement. I think it's probably a desired (if not mandatory) feature for telco customers at least. Copying policy does give us that.

imiller0: I think the primary use of this feature will be for higher level tools/controllers to control progressive rollouts of changes based on whatever logic/ordering/grouping/timing they define for their use case. In that model enabling one cluster at a time does not seem to be a hard task. With the enabling of enforcement on the child policy there is a range of flexibility in how those higher level tools define logical groups and move clusters into/out of the group. As an example, some real-world rollout strategies would attempt to keep a maximum allowable number of clusters updating at all times. As one cluster completes remediation another one would be enabled (added into the group). These scenarios need the enforcement control to be at a single-cluster granularity (or a flexible grouping construct that allows clusters to be added/removed).

Your point is good though that for a human user to make use of this feature this is a harder interface to manage since each cluster will need to be enabled manually. If there is a way to support both granular levels of control for flexibility as well as group operations that would be good. The current proposal has a bias towards the flexibility and simplicity (eg no additional CRs)

One other consideration, it was intentional in the proposal to avoid any additional "cluster selection" mechanism or secondary set of placement specifications (mentioned in [lines 190-195](https://github.com/open-cluster-management-io/enhancements/pull/57/files#diff-3b6e09ea9bfe2c7568b14f36e957e0792f6af36e5ede839f279d0499b8b02ed1R190-R195)). For example, if users specify "clusterX" in a selector for "enforcement" but that cluster is not included in the placement API the policy won't be enforced and the user will have to debug why.

mprahl: > Would this be a good summary of the proposal?
> 
> * remediationAction: [controlled | inform | enforce ]
> * controlled will generate child policies with explicit: `remediationAction: inform`
> * user/controller/orchestration can set `remediationAction: enforce` on any child policy
> * Any change/update to the parent policy resets all child policies to `remediationAction: inform`
> * [optional/future] annotation can be added to change the "reset" behavior

I like this approach, however, my preference is still as @serngawy stated about using a placement to override the enforcement behavior.

To expand on that, we could have the following new option on the parent policy:

```yaml
remediationAction: inform
remediationActionOverrides:
  enforce:
    - apiGroup: policy.open-cluster-management.io
      kind: PlacementRule
      name: <my placement>
```

For now, we could limit the `remediationActionOverrides` to only allowing the key of `enforce` but the structure would allow us flexibility for other options in the future.

The higher-level controller would still determine which clusters it wants to enforce the policy on if simple label filtering isn't sufficient and it could then create/modify the placement to target clusters by name.

What do you think?

mprahl: > By saying end user I mean any human/controller/tool and I believe in anyway of doing it we will end up having clusterSelector in order to apply the policy remediation for the subset of the bigger policy->placement->clusters group . So how this selection suppose to happen and validated from the policy->placement->clusters space ? may be giving a full use-case example could make this more clear

In my example as shown below, the idea is that your policy would be applied in inform mode against all managed clusters that match the placements bound to the policy using placement bindings like it is today.

Additionally, to select a subset of clusters to enforce, you would specify a `Placement` and/or `PlacementRule` under `remediationActionOverrides.enforce`. If this placement matches clusters that are not in the overall placement of the policy, those clusters would be ignored. Does this make sense?

```yaml
remediationAction: inform
remediationActionOverrides:
  enforce:
    - apiGroup: policy.open-cluster-management.io
      kind: PlacementRule
      name: <my placement>
```

serngawy: yes, make sense @mprahl. its just i didn't saw your comment before i post my comment (we make comment at same time i guess) . What do you think @imiller0 ? I believe this approach will avoid handling the child policy by another tool/controller

imiller0: There are a couple aspects of this override approach that I like but also a couple issues. 

Benefits:
- Using a PlacementRule like this has the advantage of allowing the user/controller/orchestrator to easily select multiple clusters for enforcement. The current proposal requires enabling enforcement on a per-cluster basis. 
- The effect is directly visible in the parent policy. This makes it easier to see what is enabling enforcement.

Issues:
- Duplicate placement APIs. There is a bit of explanation provided under the [implementation details](https://github.com/open-cluster-management-io/enhancements/pull/57/files#diff-3b6e09ea9bfe2c7568b14f36e957e0792f6af36e5ede839f279d0499b8b02ed1R190-R195), but having two sets of placement APIs for the same policy can get confusing/ambiguous when the sets are not completely overlapping. If the inform placement API includes clusters A, B, and C, and the enforce placement APIs include clusters B, E, and F what would be the resulting behavior? The logic would either take the intersection (leaving E and F not enforced) or as distinct rules (enforcing on E and F even though they weren't originally bound to the inform policy). The latter case  would make it very easy to enforce the policy against an unintended cluster when the enforce placement API uses a label that is not a subset of the inform selection.
- Shared ownership of the parent policy. In our use case the parent policy is created through gitops. To enable enforcement the modification by the user/controller/orchestrator would be done directly on the parent policy (not via gitops) leaving the state of the CR shared between the gitops operator and the user/controller/orchestrator. This is very similar issue as occurs when modifying the child policy (in the proposal I summarized above).

imiller0: GitOps can be configured to ignore differences between the source (in git) and the CR on the cluster, but it partly bends the model of gitops (what's in git is reflected in what is on the cluster) and likely increases the amount of work that needs to be done while gitops is reconciling the CRs (deeper inspection and filtering of fields). We could avoid this "2 controllers fighting over the same CR" issue by moving the enforce override into the PlacementBinding rather than the Policy. The user/controller/orchestrator would create both an PlacementBinding (with enforce override) and PlacementRule. This keeps their contribution entirely outside the gitops generated/synced content. This has a couple downsides but perhaps the improved "richness" of the selection syntax makes up for them:
- The user/controller/orchestrator now needs to create, manage, and clean-up these extra CRs. With the approach of manipulating the child policy there aren't any additional CRs needed.
- The issue noted above that when the enforce PlacementRule is not entirely a subset of the inform, the user may be surprised that some clusters are not acted on as they expect. With additional status/logging we should be able to make the reason apparent to a user who goes looking
- The enforce action is decoupled from the parent policy. Anyone looking for why a child policy is being enforced would need to be aware of an override that exists solely within the PlacementBinding. Can the status field of the policy be updated to include information on an override?

In terms of implementation, would a second PlacementBinding result in an additional child policy? For scaling reasons this would be an issue.

From the perspective of a controller that would use this, I'm curious what @jc-rh thinks from prior work in this area?

Something like:
```yaml
apiVersion: policy.open-cluster-management.io/v1
kind: PlacementBinding
...
placementRef:
  apiGroup: apps.open-cluster-management.io
  kind: PlacementRule
  name: override-placementrule
subjects:
- apiGroup: policy.open-cluster-management.io
  kind: Policy
  name: test-policy-1
  remediationActionOverride: [enforce|inform]
```

jc-rh: Just realized that the dependecy between the rules should be inverted:
```
apiVersion: apps.open-cluster-management.io/v1
kind: PlacementRule
metadata:
  name: enforce-placementrule1
  namespace: consumer-controller-managed-ns
spec:
  baseRuleRef:
    name: inform-placementrule1
    namespace: gitops-managed-ns
  clusterSelector: 
...
```
We would still need to create a new binding. It would be great if the binding can specify subjects from a different namespace.

jc-rh: Just realized that the dependecy between the rules could be inverted:
```
apiVersion: apps.open-cluster-management.io/v1
kind: PlacementBinding
metadata:
  name: enforce-placementbinding1
  namespace: consumer-controller-managed-ns
spec:
  placementRef:
    apiGroup: apps.open-cluster-management.io
    kind: PlacementRule
    name: enforce-placementrule1
    basePlacementRef:
      name: inform-placementrule
      namespace: gitops-managed-ns
  subjects:
  - apiGroup: policy.open-cluster-management.io
    kind: Policy
    name: test-policy-1
    namespace: gitops-managed-ns
...
```
The new binding has a ref to the base rule in addition to the override rule. This way there won't be share ownership on the binding. But there is no guarantee on whether the base rule is indeed associated with the policy ...

mprahl: @jc-rh that's what I was thinking. Could you just have a placement rule override per policy to avoid conflicts if it's too complex for TALO to sync it across all policies?

jc-rh: I was assuming one override rule per policy. The sharing case is where multiple CGUs remediating the same policy on different set of clusters. With this approach, TALO will have to update the cluster list in this shared rule without overwriting the clusters from other CGU. That's why I had multiple rules in my override through binding example.

Single rule vs multiple rule aside, I feel that binding is better holder for the override rule than policy for a couple of reasons

- It's consistent with the general placement design
- It allows us to link the override rule with the base rule so that it's clear the override rule acts as a subfilter on top of a particular base rule (consider a policy can be placed through multiple bindings/rules)

However, I am not sure if there is reliable way of finding the binding that does the placement of the policy on a particular cluster ...

serngawy: @jc-rh  I think it will be much simpler if you consider moving the managedCluster in/out from the `remediationActionOverrides` placement/placementRule. For example; let's assume the policy placmentRule has a clusterSelector  as below

`  clusterSelector: 
  matchExpressions: 
  - key: common
    operator: In
    values:
    - "true"
`
AND for the `remediationActionOverrides` placement/placementRule has a clusterSelector as below

`  clusterSelector: 
  matchExpressions: 
  - key: enforce
    operator: In
    values:
    - "policyName"
`
So what TALO can do is update the ManagedCluster CR labels to include the matching label `enforce: policyName`  then the cluster will be part of `remediationActionOverrides` placementRule clusters list. once the policy becomes compliant TALO remove/update the label `enforce: policyName`  from the ManagedCluster CR.

jc-rh: I think we just completed a full circle. We said it might be better to have a group thing so we don't label child polices one by one. Now you are saying we label managed cluster one by one :)

mprahl: Where did we end up on this discussion?

I'm thinking that I don't want to support a user/operator modifying the replicated policy's remediation action. I want to keep that replicated policy as an implementation detail and not as a new interface for users to further alter policy behavior.

My preference is to allow a placement binding to override the remediation action of a subset of clusters that the overall placement selects.

@JustinKuli what do you think?

mprahl: > To be honest I got a little lost in this discussion... we might need a summary of the current options. Sorry sweat_smile

I don't blame you haha.

I'll try to summarize.

## Option 1

Specify an additional placement in the parent policy of the subset of clusters that should have the remediation action set to enforce.

```yaml
remediationAction: inform
remediationActionOverrides:
  enforce:
    - apiGroup: policy.open-cluster-management.io
      kind: PlacementRule
      name: <my placement>
```

## Option 2

Specify a remediation action override for a subset of clusters in the `PlacementBinding`.

```yaml
apiVersion: policy.open-cluster-management.io/v1
kind: PlacementBinding
...
spec:
  placementRef:
    apiGroup: apps.open-cluster-management.io
    kind: PlacementRule
    name: inform-placementrule
    overridePlacementRef:
      - apiGroup: apps.open-cluster-management.io
        kind: PlacementRule
        name: enforce-placementrule1
      - apiGroup: apps.open-cluster-management.io
        kind: PlacementRule
        name: enforce-placementrule2      
  subjects:
  - apiGroup: policy.open-cluster-management.io
    kind: Policy
    name: test-policy-1
```

## Option 3

This is the proposal in the doc in the PR. This proposes setting the parent policy remediation action to `controlled` and allow users and other processes to set the `remediationAction` in each replicated policy.

JustinKuli: Call me old school, but I like the original proposal - option 3.

I don't mind that a user/controller would be able to modify a field on the replicated policies. It feels like the proposal has considered how to do this safely, and if a rollback is needed it is very easy to remove the `controlled` bit from the parent policy. The main advantage I see is that as a user, I can just look at the Policy objects on the hub and understand what is going on.

imiller0:
Summarizing an enforcement approach discussed with @mprahl @gparvin @JustinKuli @serngawy @jc-rh and @imiller0.
Enforcement is by creation of additional PlacementRule/PlacementBinding:
- remediationActionOverride in PlacementBinding – can only be set to enforce
- By default the override PB/PR selects from the full set of managed clusters
- Optional field in PlacementBinding allows sub-filtering to select from only managed clusters bound via non-sub-filtered placement rules

An update to the enhancment for review will be posted shortly

enhancements/sig-policy/28-selective-policy-enforcment/README.md
imiller0: Resolving this thread based on the discussion and subsequent updated proposal.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,415 @@
...
++ The value of `any` indicates that changes to the parent policy are
+immediately passed through to the enforcement of the child policy. As
+long as this annotation has the value of `any` remediation is enabled
+on the child policy and changes to the parent policy are immediately
+enforced to the cluster.
++ An integer `generation` value indicates that the child policy will
+continue enforcing as long as the child policy `metadata/generation`
+value is equal to the specified value. In this way the user of this
+feature has control over whether changes to the parent policy happen
+immediately or require additional intervention by the user. At the
```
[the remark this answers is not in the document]
imiller0: Resolving this conversation based on the updated selection mechanism.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,603 @@
...
++ For the bound Policy, all selected clusters will have the
+  `remediationAction` overridden.
+
+`true` this option restricts the set of clusters considered for a
+remediationActionOverride. When true, the set of clusters that are
+considered when evaluating the PlacementRule for the specific
+PlacementBinding is restricted to the set that are selected by all
+other PlacementBindings for the policy where the `subFilter` field is
+false (examples below).
+
```
gparvin, line 175: I think it should be mentioned that the enhancement will work as expected regardless of whether the PlacementBinding is attached to a Policy or to a PolicySet.

gparvin (commented):
I had one minor comment to mention this should work with PolicySet also.  I don't think it really changes anything in the proposal.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
imiller0: I updated the proposal to call out that it applies to PolicySet as well. Thanks.

mprahl (changes requested):

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,604 @@
...
+  kind: Policy
+  name: test-policy-1
+```
+
+The `subFilter` field is false so all clusters in all placementrules
+which are bound to the policy are considered. In this case the bound
+PlacementRules are placementrule-initial and placementrule-extended
+(in this binding).
+
+Clusters A, B, C, D, E, and F are all set to enforce mode.
```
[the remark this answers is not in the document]
imiller0: Yes, good catch. Thank you!

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,604 @@
...
+The proposal extends the API with `remediationActionOverride` and
+includes two configurable elements as shown in this sample:
+
+```
+apiVersion: policy.open-cluster-management.io/v1
+kind: PlacementBinding
+...
+remediationActionOverride:
+    remediationAction: [null|enforce]
+    subFilter: [fale|true]
```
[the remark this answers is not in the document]
imiller0: Fixed, thanks.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,604 @@
...
++ The behavior of multiple PlacementBindings for the same policy and
+  cluster remains unchanged (they are additive).
+  + For reducing resources the implementation should de-duplicate any
+    child policies created
++ A policy bound to the same cluster under multiple PlacementBindings
+  will be set to `enforce` if _any_ of those bindings has
+  `remediationActionOverride.remediationAction` set to enforce.
++ A policy with `remediationAction` already set to `enforce` is
+  unaffected by a PlacementBinding with
+  `remediationActionOverride.remediationAction` set to enforce.
```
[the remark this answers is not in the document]
imiller0: Made this change.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,604 @@
...
+
+### Sub filtering option in PlacementBinding
+The `remediationActionOverride.remediationAction` field allows users
```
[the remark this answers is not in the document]
imiller0: Fixed, thanks.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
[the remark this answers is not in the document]
imiller0: Updated.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,604 @@
...
+that a second PlacementBinding used to enforce the policy to a smaller
+number of clusters will respect the intent of the initial binding.
+
+The behavior of the `remediationActionOverride` when the `subFilter`
+is `false` is:
+
++ For the bound Policy or PolicySet, all selected clusters will have
+  the `remediationAction` overridden.
+
+`true` this option restricts the set of clusters considered for a
```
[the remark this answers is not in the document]
imiller0: Yes, that would be my expectation of the behavior as well with the additional axis of "for a given policy". I've added this algorithm to cover this case. Thanks!

mprahl (approved):
Nice job! This is a great and well thought out feature.

/hold for others to review

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -87,50 +87,208 @@ Out of scope for this feature
...
 
-The mechanism to enable enforcement of a child policy would be to add
-an annotation to the child policy which enables enforcement
+```
+apiVersion: policy.open-cluster-management.io/v1
+kind: PlacementBinding
+...
+remediationActionOverride:
+    remediationAction: [null|enforce]
+    subFilter: [fale|true]
```
jc-rh: Does this field belong to the binding itself? I thought this was for the binding part, not so much for the action override?

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -87,50 +87,208 @@ Out of scope for this feature
...
+```
+
+The `subFilter` field is true in this example so only clusters in
+placementrules with `subFilter` that is false are considered. In this
+case that is only the placementrule-initial with clusters A, B, C, and
+D. The PlacementRule in this CR is `placementrule-extended` which
+includes A, B, E, and F.
+
+The set of clusters set to enforce mode is the intersection of these
+two sets. Clusters A and B are set to enforce mode.
```
jc-rh, line 272: So E and F will still be bound without action override? Not sure in what use case this kind of mixed  binding behavior would be useful. For our main use case from TALM, it would make more sense to keep the subfilter at the binding level so that the new binding only does action override to already bound clusters and no other clusters get impacted.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,624 @@
...
```
jc-rh, line 272: So E and F will still be bound without action override? Not sure in what use case this kind of mixed binding behavior would be useful. For our main use case from TALM, it would make more sense to keep the subfilter at the binding level so that the new binding only does action override to already bound clusters and no other clusters get impacted.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,624 @@
...
+The proposal extends the API with `remediationActionOverride` and
+includes two configurable elements as shown in this sample:
+
+```
+apiVersion: policy.open-cluster-management.io/v1
+kind: PlacementBinding
+...
+remediationActionOverride:
+    remediationAction: [null|enforce]
+    subFilter: [false|true]
```
jc-rh, line 105: Does this field belong to the binding itself? I thought this was for the binding part, not so much for the action override?

imiller0: I think it stays here with the override. We don't want to change the general semantics of the binding or of multiple bindings. The current behavior of multiple bindings is that they associate a policy with clusters and multiple bindings are a "union". For this feature, and our use case, the subFilter is a mechanism to select a subset of the already bound clusters and do the override to them.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
imiller0: Yes, E and F are bound only in this override placement binding which is created to apply the policy to a subset of the original bound clusters. When the user sets subFilter to "true" they are saying "I don't want this binding/override to add any new clusters to the set that the policy applies to, I'm just trying to enforce it to some of the originally bound ones". The subFilter is a safety mechanism to avoid accidentally adding more clusters into the set.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
mprahl: Actually, I think it might make sense to have it at the root level since it would allow us to add additional overrides in the future but have a `subFilter` value per `PlacementBinding`.

imiller0: I've updated the proposal to move the subFilter to the root level so that it is a control on how the binding itself works. As @jc-rh recommended this makes the override control simpler (it applies to this whole binding) and allows future enhancements to benefit from subFiltering. Thanks!

mprahl (approved):
Nice!

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,639 @@
...
+
+## Motivation
+
+Selective Policy Enforcement allows fine grained control over the
+timing and application of a policy to clusters. It is intended to
+support this level of control natively within the policy framework. By
+supporting this as a feature users, or higher level
+controllers/orchestrators, do not have to implement a procedural
+pattern of copying policies and manipulating bindings/labels to
+progressivly enforce the policy on clusters.
```
JustinKuli, line 48: Nit
```suggestion
progressively enforce the policy on clusters.
```

jc-rh (approved):

enhancements/sig-policy/28-selective-policy-enforcment/README.md
```diff
@@ -0,0 +1,639 @@
...
+The subFilter option is specified to allow a user to ensure that a
+single PlacementBinding can be used to define the set of potentially
+affected clusters. Without the subFilter option, it is possible to
+erroneously select additional clusters in the "override"
+PlacementBinding and accidentally enforce the policy to those
+additional clusters. The `subFilter` option, when set true, ensures
+that a second PlacementBinding used to enforce the policy to a smaller
+number of clusters will respect the intent of the initial binding.
+
+For a given Policy, when there are multiple PlacementBindings
```
JustinKuli, line 189: I wonder if we should add a diagram for this, I would be happy to draft something (I did it on paper and it certainly helped me understand what this is saying)

JustinKuli (approved):
/lgtm

Very well-written! I feel good about this.

enhancements/sig-policy/28-selective-policy-enforcment/README.md
imiller0: I think that would be really helpful. Thanks!

JustinKuli: Thoughts? Easy to change:

override-diagram

imiller0: Looks great. Thank you! If you want to post a PR to this branch I'll merge it, or I can directly add the diagram.

JustinKuli: You can add it - no need for an extra commit

imiller0: Added. Thank you!

mprahl (approved):
Nice!

PR #58, opened 2022-06-08
addon configuration enhancement

skeeey:
Signed-off-by: Wei Liu <liuweixa@redhat.com>

### DIFF
--- a/enhancements/sig-architecture/58-addon-configuration/README.md
+++ b/enhancements/sig-architecture/58-addon-configuration/README.md
@@ -1,0 +1,617 @@
+# Add-on Configuration
+
+## Release Signoff Checklist
+
+- [] Enhancement is `implemented`
+- [] Design details are appropriately documented from clear requirements
+- [] Test plan is defined
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal includes:
+
+1. The add-on configuration reference in the add-on APIs to help add-on developers to locate their add-on configuration.
+2. An add-on deployment configuration API, the add-on developers can choose to use this API to configure their add-on
+deployment.
+
+## Motivation
+
+For some add-ons, they need to run with configuration, we want to provide a way to locate their configuration with
+add-on APIs.
+
+## Use cases
+
+1. As a user, I want to specify a default nodeSelector/tolerations configuration for all my add-ons.
+2. As a user, I want to specify nodeSelector/tolerations configuration for some of my add-ons.
+
+## Proposal
+
+We want to refactor the definition of `ConfigCoordinates` in the `ClusterManagementAddOn` API and introduce a way to
+specify the configuration namespace and name in the `ManagedClusterAddOn` API.
+
+### Design Details
+
+Currently, we have a `ConfigCoordinates` in the `ClusterManagementAddOn` API. It helps to locate the add-on
+configuration information.
+
+```golang
+// ClusterManagementAddOnSpec provides information for the add-on.
+type ClusterManagementAddOnSpec struct {
+    // addOnMeta is a reference to the metadata information for the add-on.
+    // +optional
+    AddOnMeta AddOnMeta `json:"addOnMeta"`
+
+    // addOnConfiguration is a reference to configuration information for the add-on.
+    // +optional
+    AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration"`
+}
+
+// ConfigCoordinates represents the information for locating the CRD and CR that configures the add-on.
+type ConfigCoordinates struct {
+    // crdName is the name of the CRD used to configure instances of the managed add-on.
+    // This field should be configured if the add-on have a CRD that controls the configuration of the add-on.
+    // +optional
+    CRDName string `json:"crdName"`
+
+    // crName is the name of the CR used to configure instances of the managed add-on.
+    // This field should be configured if add-on CR have a consistent name across the all of the ManagedCluster
+    // instaces.
+    // +optional
+    CRName string `json:"crName"`
+
+    // lastObservedGeneration is the observed generation of the custom resource for the configuration of the add-on.
+    // +optional
+    LastObservedGeneration int64 `json:"lastObservedGeneration,omitempty"`
+}
+```
+
+But the `ConfigCoordinates` only supports to use `CustomResourceDefinition` as the add-on configuration, and it's lacks
+the namespace/name for a configuration.
+
+To make the `ConfigCoordinates` more flexible, we want to refactor the `ConfigCoordinates` to
+
+``` golang
+// ConfigCoordinates represents the information for locating the add-on configuration.
+type ConfigCoordinates struct {
+    // Deprecated: Use configGroupResource filed instead
+    // crdName is the name of the CRD used to configure instances of the managed add-on.
+    // This field should be configured if the add-on have a CRD that controls the configuration of the add-on.
+    // +optional
+    CRDName string `json:"crdName,omitempty"`
+
+    // Deprecated: Use configGroupResource filed instead
+    // crName is the name of the CR used to configure instances of the managed add-on.
+    // This field should be configured if add-on CR have a consistent name across the all of the ManagedCluster
+    // instaces.
+    // +optional
+    CRName string `json:"crName,omitempty"`
+
+    // Deprecated: This will be removed
+    // lastObservedGeneration is the observed generation of the custom resource for the configuration of the addon.
+    // +optional
+    LastObservedGeneration int64 `json:"lastObservedGeneration,omitempty"`
+
+    // ConfigGroupResource represents the GroupResource of the add-on configuration.
+    // +optional
+    ConfigGroupResource ConfigGroupResource `json:"configGroupResource,omitempty"`
+    
+    // defaultConfig represents the namespace and name of the default add-on configuration.
+    // In scenario where all add-ons have a same configuration.
+    // +optional
+    DefaultConfig ConfigReferent `json:"defaultConfig,omitempty"`
+}
+
+// ConfigGroupResource represents the GroupResource of the add-on configuration
+type ConfigGroupResource struct {
+    // group of the add-on configuration.
+    Group string `json:"group"`
+
+    // resource of the add-on configuration.
+    Resource string `json:"resource"`
+}
+
+// ConfigReferent represents the namespace and name for an add-on configuration.
+type ConfigReferent struct {
+    // namespace of the add-on configuration.
+    // If this field is not set, the configuration is in the cluster scope.
+    // +optional
+    Namespace string `json:"namespace,omitempty"`
+
+    // name of the add-on configuration.
+    Name string `json:"name"`
+}
+```
+
+When the `ConfigCoordinates` is defined in the `ClusterManagementAddOnSpec`, the addon-framework will sync the
+information of `ConfigCoordinates` to the `ManagedClusterAddOn` status.
+
+In the `ManagedClusterAddOn` spec, we will add a `config` field, it is used to specify the add-on configuration
+namespace and name, and introduce a `ConfigReference` instead of the `ConfigCoordinates` in the
+`ManagedClusterAddOn` status.
+
+We will introduce a `Configured` condition for the `ManagedClusterAddOn` to report the condition when we configure
+an addon with its configuration.
+
+```golang
+const (
+    // ManagedClusterAddOnCondtionConfigured represents that the addon agent is configured with its configuration
+    ManagedClusterAddOnCondtionConfigured string = "Configured"
+)
+
+// ManagedClusterAddOnSpec defines the install configuration of an add-on agent on managed cluster.
+type ManagedClusterAddOnSpec struct {
+    // installNamespace is the namespace on the managed cluster to install the add-on agent.
+    // If it is not set, open-cluster-management-agent-addon namespace is used to install the add-on agent.
+    // +optional
+    // +kubebuilder:validation:MaxLength=63
+    // +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+    InstallNamespace string `json:"installNamespace,omitempty"`
+
+    // config represents the configuration namespace and name for the current add-on.
+    // In scenario where the current add-on has its own configuration.
+    // +optional
+    Config ConfigReferent `json:"config,omitempty"`
+}
+
+// ManagedClusterAddOnStatus provides information about the status of the operator.
+// +k8s:deepcopy-gen=true
+type ManagedClusterAddOnStatus struct {
+    // Deprecated: Use configReference instead
+    // addOnConfiguration is a reference to configuration information for the add-on.
+    // This resource is use to locate the configuration resource for the add-on.
+    // +optional
+    AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration",omitempty`
+
+    // configReference is a reference to the add-on configuration.
+    // +optional
+    ConfigReference ConfigReference `json:"configReference,omitempty"`
+}
+
+// ConfigReference is a reference to the current add-on configuration.
+// This resource is used to locate the configuration resource for the current add-on.
+type ConfigReference struct {
+    // This field is synced from ClusterManagementAddOn configGroupResource field.
+    ConfigGroupResource ConfigGroupResource `json:",inline"`
+
+    // This field is synced from ClusterManagementAddOn defaultConfig and ManagedClusterAddOn config fields.
+    // If both them are specified, the ManagedClusterAddOn config will overwrite the ClusterManagementAddOn
+    // defaultConfig.
+    Config ConfigReferent `json:",inline"`
+
+    // version of the add-on configuration.
+    Version string `json:"version"`
+
+    // lastObservedGeneration is the observed generation of the add-on configuration.
+    LastObservedGeneration int64 `json:"lastObservedGeneration"`
+}
+```
+
+The add-on framework will watches the `ManagedClusterAddOn` to get the add-on configuration reference from the 
+`ManagedClusterAddOn` status.
+
+If the `ManagedClusterAddOn` has the configuration reference, the add-on framework use the `configGroupResource` and
+`config` to locate the configuration resource and watch the configuration resource, once the configuration resource
+changes, the add-on famework updates the `ManagedClusterAddOn` status (`lastObservedGeneration`), this will trigger
+the [`Manifests`](https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/agent/inteface.go#L31)
+interface which add-on implement.
+
+As a specific case, we want to introduce a `AddOnDeploymentConfig` API in cluster scope to configure the add-on
+deployment.
+
+This API is optional for add-on developers, add-on developers can choose to use this API to configure their add-on
+deployment or use their own API for configuration.
+
+```golang
+// AddOnDeploymentConfig represents a deployment configuration for an add-on.
+// AddOnDeploymentConfig is a cluster-scoped resource.
+type AddOnDeploymentConfig struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // spec represents a desired configuration for an add-on.
+    // +required
+    Spec AddOnDeploymentConfigSpec `json:"spec"`
+}
+
+const ManagedClusterAddOnConditionInvalidDeploy
+
+type AddOnDeploymentConfigSpec struct {
+    // CustomizedVariables is a list of name-value variables for the current add-on deployment.
+    // The add-on implementation can use these variables to render the add-on deployment.
+    // For example, using this to define the image information of one add-on agent:
+    // `"customizedVariables": [
+    //    {"name":"Image","value":"myimg:latest"},
+    //    {"name":"ImagePullPolicy","value":"Always"}]`,
+    // or using this to define a http proxy for one add-on agent:
+    // `"customizedVariables": [
+    //    {"name":"HTTP_PROXY","value":"http://proxy.example.com:3128"},
+    //    {"name":"HTTPS_PROXY","value":"http://proxy.example.com:3128"},
+    //    {"name": "NO_PROXY", "value":"svc,local"}]` 
+    // The default is an empty list.
+    // +optional
+    CustomizedVariables []CustomizedVariable `json:"customizedVariables,omitempty"`
+
+    // NodePlacement enables explicit control over the scheduling of the add-on.
+    // +optional
+    NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+}
+
+// CustomizedVariable represents a customized variable for add-on deployment.
+type CustomizedVariable struct {
+    // Name of this variable. Must be a C_IDENTIFIER.
+    Name string `json:"name"`
+
+    // Value of this variable.
+    Value string `json:"value"`
+}
+
+// NodePlacement describes node scheduling configuration for the pods.
+type NodePlacement struct {
+    // NodeSelector defines which Nodes the Pods are scheduled on. The default is an empty list.
+    // +optional
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+    // Tolerations is attached by pods to tolerate any taint that matches
+    // the triple <key,value,effect> using the matching operator <operator>.
+    // The default is an empty list.
+    // +optional
+    Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+}
+```
+
+Examples:
+
+1. Using `AddOnDeploymentConfig` to configure add-on
+
+    User have a default add-on deployment configuration on the hub
+
+    ```yaml
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: AddOnDeploymentConfig
+    metadata:
+      name: addon-default-placement
+    spec:
+      nodeSelector:
+        "kubernetes.io/arch": "amd64"
+      tolerations:
+      - key: "node.kubernetes.io/unreachable",
+        operator: "Exists",
+        effect: "NoSchedule"
+    ```
+
+    One add-on operator creates the `ClusterManagementAddOn` with the configuration referance on the hub to share this
+    configuration to its add-ons
+
+    ```yaml
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: ClusterManagementAddOn
+    metadata:
+      name: myaddon
+    spec:
+      addOnConfiguration:
+        configGroupResource:
+          group: addon.open-cluster-management.io
+          resource: AddOnDeploymentConfig
+        defaultConfig:
+          name: addon-default-placement
+    ```
+
+    For each managed cluster add-on, the configuration can be located with `configReference` in the
+    `ManagedClusterAddOnStatus`
+
+    ```yaml
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: ManagedClusterAddOn
+    metadata:
+      name: myaddon
+      namespace: cluster1
+    spec:
+      installNamespace: open-cluster-management-agent-addon
+    status:
+      configReference:
+        group: addon.open-cluster-management.io
+        version: v1alpha1
+        resource: AddOnDeploymentConfig
+        name: addon-default-placement
+        lastObservedGeneration: 1
+
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: ManagedClusterAddOn
+    metadata:
+      name: myaddon
+      namespace: cluster2
+    spec:
+      installNamespace: open-cluster-management-agent-addon
+    status:
+      configReference:
+        group: addon.open-cluster-management.io
+        version: v1alpha1
+        resource: AddOnDeploymentConfig
+        name: addon-default-placement
+        lastObservedGeneration: 1
+    ```
+
+
+    If user need to configure some add-ons specially, the user can define a specifc configuration and specify the
+    configuration name in the `ManagedClusterAddOn`
+
+    ```yaml
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: AddOnDeploymentConfig
+    metadata:
+      name: addon-arm-placement
+    spec:
+      nodeSelector:
+        "kubernetes.io/arch": "arm64"
+      tolerations:
+      - key: "node.kubernetes.io/unreachable",
+        operator: "Exists",
+        effect: "NoSchedule"
+
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: ManagedClusterAddOn
+    metadata:
+      name: myaddon
+      namespace: cluster3
+    spec:
+      installNamespace: open-cluster-management-agent-addon
+      config:
+        name: addon-arm-placement
+    ```
+
+    then the configuration reference in this add-on status will be
+
+    ```yaml
+    apiVersion: addon.open-cluster-management.io/v1alpha1
+    kind: ManagedClusterAddOn
+    metadata:
+      name: myaddon
+      namespace: cluster3
+    spec:
+      installNamespace: open-cluster-management-agent-addon
+      config:
+        name: addon-arm-placement
+    status:
+      configReference:
+        group: addon.open-cluster-management.io
+        version: v1alpha1
+        resource: AddOnDeploymentConfig
+        name:  addon-arm-placement
+        lastObservedGeneration: 1
+    ```
+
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,62 @@
...
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+The proposal introduces two annotations to help user to specify node selector and tolerations for their addons.
+
+## Motivation
+
+Sometimes user want to place their addons on specific nodes, so we need a way to set the node selector and tolerations for a addon.
```
qiujian16, line 17: a --> an

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,62 @@
...
+### Test Plan
+- test add placement annotation to a addon which has no values annotation 
+- test add placement annotation to a addon which has values annotation 
+- test delete placement annotation from a addon
+
+### Graduation Criteria
+N/A
+
+### Upgrade Strategy
+N/A
```
qiujian16, line 57: any concern on the upgrade?

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: fixed

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```yaml
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ManagedClusterAddOn
+metadata:
+  name: work-manager
+  namespace: cluster1
+  annotations:
+    addon.open-cluster-management.io/nodeSelector: "{\"kubernetes.io/os\":\"linux\"}"
+    addon.open-cluster-management.io/tolerations: "[{\"key\":\"node.kubernetes.io/unreachable\",\"operator\":\"Exists\",\"effect\":\"NoSchedule\"}]"
```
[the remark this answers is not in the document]
qiujian16: I think values is reserved for other usage? what if a user or another controller set the value in advance. I think it will be hard to merge them.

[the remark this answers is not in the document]
zhiweiyin318: I just use values as an example, we could find a new name if using one annotation is accepted.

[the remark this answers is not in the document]
skeeey: I think use two independent annotations is more clear and more easier to maintain

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: add a probable upgrade case, and expecting other reviewers give more comments

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,65 @@
+# The Placement Annotation of Addon
```
mikeshng, line 1: Minor nit. Can we use a different word instead of `Placement`? We have the Placement API and this enhancement is talking about something that is not related to that.

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,65 @@
...
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+The proposal introduces two annotations to help user to specify node selector and tolerations for their addons.
+
+## Motivation
+
+Sometimes user want to place their addons on specific nodes, so we need a way to set the node selector and tolerations
```
mikeshng, line 17:
```suggestion
Sometimes user want to assign their addons on specific nodes, so we need a way to set the node selector and toleration
```

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,65 @@
...
+for an addon.
+
+## Proposal
+
+We introduce two annotations to set node selector and tolerations for an addon
```
mikeshng, line 22:
```suggestion
We introduce two annotations to set node selector and toleration for an addon:
```

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. Node selector annotation, the key of this annotation is: `addon.open-cluster-management.io/nodeSelector`, the value of this annotation represents a node selector with json format
+2. Tolerations annotation, the key of this annotation is: `addon.open-cluster-management.io/tolerations`, the value of this annotation represents a toleration array with json format
+
+There is an example
```
mikeshng, line 27:
```suggestion
Example:
```

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,65 @@
...
+N/A
+
+### Upgrade Strategy
+
+If there is an existing addon CR that already used the addon values annotation to specify the node selector and tolerations and the addon implementation used the `GetValuesFromAddonAnnotation` function to get them. The addon user can choose to keep them in addon values annotation or use new addon placement annotation to define them and remove them from the addon values annotation.
+
+### Version Skew Strategy
+N/A
+
+## Alternatives
```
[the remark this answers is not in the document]
skeeey: yeah, we have considered this, if we add these to addon spec, it may not meet all needs, e.g. if one addon need to deploy more than one agent, and these agents have different node selector/tolerations, this will be hard to use

to keep the addon flexibility, we proposed this, providing an optional way for the addon developers, they can choose use this way or use their own way to handle the node selector/tolerations for their addons, e.g. the developers may have own addon configuration cr, and use the configuration to render their addon agents

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: change to `The Addon Node Selector and Toleration Annotation`

enhancements/sig-architecture/58-addon-placement-annotation/README.md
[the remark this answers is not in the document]
qiujian16: I think we cannot ensure the agent will use this even if adding this to spec. installNamespace is different, since it is needed for the registration controller to put secret for the addon to communicate to hub. @skeeey let's add spec as an atlernative in the proposal, and clarify the reason.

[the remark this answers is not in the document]
skeeey: added

qiujian16:
/approve
/hold

let's hold for another 72 hours for additional reviews.

enhancements/sig-architecture/58-addon-placement-annotation/README.md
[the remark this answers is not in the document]
mprahl: I had the same thought. I think the additional flexibility is a negative in this case. We want a consistent interface for how to configure addons from a user's perspective. If this is free-form JSON strings that are different per addon, I think this becomes hard to use and confusing. It's especially confusing when a user puts in an invalid JSON string since there is no immediate feedback from API validation. It also makes API compatibility hard since it becomes up to each OCM addon how to interpret that JSON strin.

I also dislike that an annotation has a JSON string as a value since these are not user-friendly to read or write and suggest that a new item in the spec is more appropriate.

My suggestion is to make the spec flexible enough for all current OCM addons and make it optional. Any complex configurations beyond this can be addon specific in their own CRDs.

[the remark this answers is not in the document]
qiujian16: but nodeSelect/toleration is not a common property for every addon. Having nodeSelect/toleration in the spec gives an impression that the the addon will honor this field, but it is not true. If we want a more strict validation, I think we should have an Object ref field (@jnpacker mentioned this also) instead of putting nodeselector in the addon spec.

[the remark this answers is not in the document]
jnpacker: What about an objectRef or list of objectRef's.  It does mean that each addon may not configure the same set of values or even resources, but if as @mprahl  suggests, you used a CRD for your addon, you could have basic validation, defaults, etc...

Also a ConfigMap or secret could be leveraged to try things out, and at least the format is well understood.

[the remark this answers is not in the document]
skeeey: I will add the objectRef to the enhancement as an optional proposal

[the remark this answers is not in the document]
mprahl: @qiujian16 and @jnpacker this sounds interesting. I hadn't heard about `objectRef` before and couldn't find anything concrete online. Could you please explain what this concept is?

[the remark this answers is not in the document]
mikeshng: @mprahl probably this https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5622-L5639

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,295 @@
...
+
+The proposal introduces some ways to help user to set up configuration for their addons.
+
+## Motivation
+
+We want to provide a way to help user to set up configuration for their addons with current addon APIs
+
+## Use cases
+
+1. As a user, I have a configuration for my all managed cluster addons, the addons share the same configuration.
```
qiujian16, line 21: I think we should have some details here.

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,295 @@
...
+configuration, and for addon developers, it's not easy to locate the configuration object in their controllers. we want
+to refactor the `ConfigCoordinates` to
+
+``` golang
+// ConfigCoordinates contains enough information to let one addon identify its configuration object.
+// A config object must be in the same namespace with its addon, or be cluster-scoped, so there is no
+// namespace field.
+type ConfigCoordinates struct {
+	// API version of the referent.
+	APIVersion string `json:"apiVersion"`
```
qiujian16, line 75: let's use gvr here. and also default namespace/name??

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,295 @@
...
+	// installNamespace is the namespace on the managed cluster to install the addon agent.
+	// If it is not set, open-cluster-management-agent-addon namespace is used to install the addon agent.
+	// +optional
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	InstallNamespace string `json:"installNamespace,omitempty"`
+
+	// addOnConfiguration is a configuration object reference that is used by current addon.
+	// +optional
+	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration,omitempty"`
```
qiujian16, line 104: we should only have namespace/name here. Also we need update status field

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,295 @@
...
+}
+```
+
+As a specific case, sometimes user want to assign their addons on specific nodes, we want to introduce a `AddOnDeployementConfig`
+to support this.
+
+```golang
+// AddOnDeployementConfig represents a deployment configuration for an add-on.
+// Currently, user can use this to configure nodeSelector, toleration and variables.
+type AddOnDeployementConfig struct {
```
qiujian16, line 114: namespace/cluster scoped?

enhancements/sig-architecture/58-addon-placement-annotation/README.md
```diff
@@ -0,0 +1,295 @@
...
+	// CustomizedVariables is a list of name-value variables for the current add-on.
+	// add-on can use these variables to render its deployment.
+	// The default is an empty list.
+	// +optional
+	CustomizedVariables []CustomizedVariable `json:"customizedVariable,omitempty"`
+
+	// NodeSelector defines which Nodes the Pods are scheduled on.
+	// The default is an empty list.
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
qiujian16, line 133: I think we should put it in a structure working as a duck type maybe. Some thing like nodeSchedulable

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: use the specific use case

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: use the gvr and add the default ns and name

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: yes, we need update status

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: it is a cluster scope

enhancements/sig-architecture/58-addon-placement-annotation/README.md
skeeey: separate to a struct

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,439 @@
...
+	// version of the add-on configuration.
+	Version string `json:"version"`
+
+	// resource of the add-on configuration.
+	Resource string `json:"resource"`
+
+	// namespace of the default add-on configuration.
+	// In scenario where all add-ons have a same configuration.
+	// +optional
+	DefaultNamespace string `json:"namespace,omitempty"`
```
qiujian16, line 101: describe what if it is empty

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,439 @@
...
+	// configuration represents the configuration namespace and name for the current add-on.
+	// In scenario where the current add-on has its own configuration.
+	// +optional
+	Configuration ConfigurationReferent `json:"configuration,omitempty"`
+}
+
+// ConfigurationReferent represents the namespace and name for the current add-on.
+type ConfigurationReferent struct {
+	// namespace of the current add-on configuration.
+	// +optional
```
qiujian16, line 136: we should clear state if it is not set, the related namespace/name in clustermanagementaddon is used if any?

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,439 @@
...
+type ManagedClusterAddOnStatus struct {
+	// Deprecated: Use configReference instead
+	// addOnConfiguration is a reference to configuration information for the add-on.
+	// This resource is use to locate the configuration resource for the add-on.
+	// +optional
+	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration",omitempty`
+
+	// configReference is a reference to the add-on configuration.
+	// +optional
+	ConfigReference ConfigReference `json:"configReference",omitempty`
```
qiujian16, line 154: I think we could put it in the ConfigCoordinates?

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,439 @@
...
+kind: ManagedClusterAddOn
+metadata:
+  name: myaddon
+  namespace: cluster1
+spec:
+  installNamespace: open-cluster-management-agent-addon
+status:
+  configReference:
+    group: addon.open-cluster-management.io
+	version: v1alpha1
```
qiujian16, line 286: format issue?

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,439 @@
...
+  namespace: cluster3
+spec:
+  installNamespace: open-cluster-management-agent-addon
+  configuration:
+    name: addon-arm-placement
+status:
+  configReference:
+    group: addon.open-cluster-management.io
+	version: v1alpha1
+    resource: ManagedClusterAddOn
```
qiujian16, line 351: the resource should be AddOnDeployementConfig

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,439 @@
...
+	// default name.
+	Name string `json:"name"`
+
+	// lastObservedGeneration is the observed generation of the add-on configuration.
+	LastObservedGeneration int64 `json:"lastObservedGeneration"`
+}
+```
+
+As a specific case, we want to introduce a `AddOnDeployementConfig` API in cluster scope to configure the add-on
+deployment.
```
qiujian16, line 191: please clarify that user can also use their own crd. I think an example of customized CRD will be helpful also.

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: corrected

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: corrected

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: I will think the example

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: added

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: I clear this in the status part

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,593 @@
...
+### Graduation Criteria
+1. The new APIs is reviewed and accepted;
+2. Implementation is completed to support the functionalities;
+3. Develop test cases to demonstrate this proposal work correctly;
+
+### Upgrade Strategy
+- We will deprecate the elder fields of `ConfigCoordinates` in `ClusterManagementAddOn`, the add-on developers need
+adopt the new API if they used the `ConfigCoordinates` in `ClusterManagementAddOn`
+- We will deprecate the values annotation for `ManagedClusterAddOn`, the add-on developers need adopt the new API if
+they used the values annotation to configure their add-ons
```
JustinKuli, line 523: Is it necessary to deprecate the values annotation? Or can it be left to individual add-ons to determine how to get configuration?

I like the approach in this proposal. I like that add-ons can choose to implement their own CRD for configuration, or they can use the proposed `AddOnDeployementConfig`, or they could even just use a ConfigMap. That flexibility is very nice.

But I think that *requiring* a separate configuration resource in all cases is unnecessary for all configuration - annotations on the ManagedClusterAddon provide a simple way for basic configuration to be applied. For example, the policy addon currently supports a `log-level` annotation which I think is a nice user experience.

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,590 @@
...
+current add-on APIs.
+
+## Use cases
+
+1. As a user, I want to specify a default nodeSelector/tolerations configuration for all my add-ons.
+2. As a user, I want to specify nodeSelector/tolerations configuration for some of my add-ons.
+
+## Proposal
+
+We want to refacotor the definition of `ConfigCoordinates` in the `ClusterManagementAddOn` API and introduce a way to
```
mprahl, line 31:
```suggestion
We want to refactor the definition of `ConfigCoordinates` in the `ClusterManagementAddOn` API and introduce a way to
```

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,590 @@
...
+    // If both them are defined, the ManagedClusterAddOn config will overwrite the ClusterManagementAddOn
+    // defaultConfig.
+    Config ConfigReferent `json:"config"`
+
+    // lastObservedGeneration is the observed generation of the add-on configuration.
+    LastObservedGeneration int64 `json:"lastObservedGeneration"`
+}
+```
+
+As a specific case, we want to introduce a `AddOnDeployementConfig` API in cluster scope to configure the add-on
```
mprahl, line 180:
```suggestion
As a specific case, we want to introduce a `AddOnDeploymentConfig` API in cluster scope to configure the add-on
```

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,590 @@
...
+deployment.
+
+The add-on developers can choose to use this API to configure their add-on deployment or use their own API for
+configuration.
+
+```golang
+// AddOnDeployementConfig represents a deployment configuration for an add-on.
+// AddOnDeployementConfig is a cluster-scoped resource.
+type AddOnDeployementConfig struct {
```
mprahl, line 189:
```suggestion
type AddOnDeploymentConfig struct {
```

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,590 @@
...
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // spec represents a desired configuration for an add-on.
+    // +required
+    Spec AddOnDeployementConfigSpec `json:"spec"`
+}
+
+type AddOnDeployementConfigSpec struct {
```
mprahl, line 198:
```suggestion
type AddOnDeploymentConfigSpec struct {
```

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,590 @@
...
+    spec:
+      nodeSelector:
+        "kubernetes.io/arch": "amd64"
+      tolerations:
+      - key: "node.kubernetes.io/unreachable",
+        operator: "Exists",
+        effect: "NoSchedule"
+    ```
+
+    User set up the configuration referance in the `ClusterManagementAddOn` to share this configuration to their add-ons
```
mprahl, line 253: How does the user know that the addon supports this configuration type?

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: done

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: good catch, I think the ClusterManagementAddOn should be applied along with addon hub controller deployed, so users should not need to configure this, I will make this clear

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,592 @@
...
+type AddOnDeploymentConfig struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // spec represents a desired configuration for an add-on.
+    // +required
+    Spec AddOnDeployementConfigSpec `json:"spec"`
+}
+
+type AddOnDeployementConfigSpec struct {
```
mprahl, line 198:
```suggestion
type AddOnDeploymentConfigSpec struct {
```

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,592 @@
...
+    // If this field is not set, the configuration is in the cluster scope.
+    // +optional
+    Namespace string `json:"namespace,omitempty"`
+
+    // name of the add-on configuration.
+    Name string `json:"name"`
+}
+```
+
+When the `ConfigCoordinates` is defined in the `ClusterManagementAddOnSpec`, the addon-framework will sync the
```
mprahl, line 125: How will the addon know when the configuration has been updated?

qiujian16: in my mind, framework will watch the configuration resource and update addon status (lastObservedGeneration), which then trigger the manifests interface which addon implement. 

The addon then needs to read this configuration itself and render the agent's manifests. @skeeey i am not sure if framework's higher level interface would look like. This might also need to be added in this proposal.

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: fixed

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,606 @@
...
+type ManagedClusterAddOnStatus struct {
+    // Deprecated: Use configReference instead
+    // addOnConfiguration is a reference to configuration information for the add-on.
+    // This resource is use to locate the configuration resource for the add-on.
+    // +optional
+    AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration",omitempty`
+
+    // configReference is a reference to the add-on configuration.
+    // +optional
+    ConfigReference ConfigReference `json:"configReference,omitempty"`
```
deads2k, line 164: only the status will have the LastObservedGeneration, right?

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,606 @@
...
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    // spec represents a desired configuration for an add-on.
+    // +required
+    Spec AddOnDeploymentConfigSpec `json:"spec"`
+}
+
+type AddOnDeploymentConfigSpec struct {
+    // CustomizedVariables is a list of name-value variables for the current add-on deployment.
+    // The add-on implementation can use these variables to render the add-on deployment.
```
deads2k, line 214: I worry about this being used in lieu of defining a schema.  Is there a way to describe what variables are available?

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: yes, it's right

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: sometimes, we need configure some env variables or other fields for deployment, e.g. the http proxy, the imagePullPolicy, etc. we want  to use this to configure them

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,606 @@
...
+    DefaultConfig ConfigReferent `json:"defaultConfig,omitempty"`
+}
+
+// ConfigGVR represents the GroupVersionResource of the add-on configuration
+type ConfigGVR struct {
+    // group of the add-on configuration.
+    Group string `json:"group"`
+
+    // version of the add-on configuration.
+    Version string `json:"version"`
```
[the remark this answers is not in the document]
skeeey: yeah, the version is not necessary in the spec, I will update this part

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: @deads2k do you have any suggestion for this part? is it better to use a generic configuration (like current name-value variables) or use a specific configuration, we define the image, imagePullPolicy and http proxy in this spec?

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,589 @@
...
+    CRName string `json:"crName,omitempty"`
+
+    // Deprecated: This will be removed
+    // lastObservedGeneration is the observed generation of the custom resource for the configuration of the addon.
+    // +optional
+    LastObservedGeneration int64 `json:"lastObservedGeneration,omitempty"`
+
+    // ConfigGR represents the GroupResource of the add-on configuration.
+    // +optional
+    ConfigGR ConfigGR `json:"configGR,omitempty"`
```
deads2k, line 98: no abbreviations in the API

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,589 @@
...
+- We do not need to change current API definitions.
+- Addon developers can choose their add-on configuration implementation freely.
+
+Cons:
+- It's hard to validate the json string that is in the annotations.
+
+### Adding the node selector and toleration to the `ManagedClusterAddOn` spec
+
+As an alternative, we may add the node selector and toleration to the add-on spec, but this may not meet all needs, e.g.
+if one add-on need to deploy more than one agent, and these agents have different node selector/toleration, and the
```
deads2k, line 585: If an add-on needs more than one node selector, how will that be achieved with the API proposed here?

enhancements/sig-architecture/58-addon-configuration/README.md
deads2k: I gave this some thought over the weekend and I cannot think of a compact way to avoid this problem.  I think we could limit the problems with free-form fields by suggesting that the managedclusteraddon report a condition if the CustomizedVariables included an unrecognized name or if the rendering fails.

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,606 @@
...
+    // +required
+    Spec AddOnDeploymentConfigSpec `json:"spec"`
+}
+
+type AddOnDeploymentConfigSpec struct {
+    // CustomizedVariables is a list of name-value variables for the current add-on deployment.
+    // The add-on implementation can use these variables to render the add-on deployment.
+    // The default is an empty list.
+    // +optional
+    CustomizedVariables []CustomizedVariable `json:"customizedVariable,omitempty"`
```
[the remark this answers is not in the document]
deads2k: > I don't see this value being used in the examples below. Typo somewhere?

Can you provide an example of this below?

deads2k:
The idea seems worth trying, I think we need a way to present failures that the managedclusteraddon observes.  I am interested in how the problem presented in the alternatives is solved by this API.

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: thank you, I think report the condition is a good suggestion, I will add this in the enhancement

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: for this case, the `AddOnDeploymentConfig` is not suitable, user can choose to define their own config api, then refer to the config api in their managedclusteraddon, so there is the reason why we don't want to add the config information to the managedclusteraddon api directly.

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: corrected

enhancements/sig-architecture/58-addon-configuration/README.md
skeeey: I add a new condition `Configured` to the managedclusteraddon, I think we can use this condition to report the condition when we configure an addon with its configuration

enhancements/sig-architecture/58-addon-configuration/README.md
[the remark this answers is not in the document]
skeeey: added

deads2k:
The design looks worth trying in alpha to me. I suggest trying to write some of these before promoting to beta.

enhancements/sig-architecture/58-addon-configuration/README.md
```diff
@@ -0,0 +1,608 @@
...
+        namesapce: cluster2
+        lastObservedGeneration: 1
+    ```
+
+### Test Plan
+- All implementations will have thorough coverage of unit tests.
+- Integration tests will cover the user cases
+
+### Graduation Criteria
+1. The new APIs is reviewed and accepted;
```
qiujian16, line 533: let's define alpha/beta stage here. I think we need some criteria for beta.
e.g.
1. at least 2-3 consumers of this API
2. at lease 2-3 customized API using this mechanism.

skeeey: added

Issue #62, opened 2022-06-14
Implement LabelSelector for policy NamespaceSelector


dhaiducek:
Enhancement related to:
- https://github.com/open-cluster-management-io/community/issues/155

PR #63, opened 2022-06-15
`LabelSelector` to select namespaces for `ConfigurationPolicy` objects

dhaiducek:
- #62 
- https://github.com/stolostron/backlog/issues/23483

### DIFF
--- a/enhancements/sig-policy/62-namespace-labelselector/README.md
+++ b/enhancements/sig-policy/62-namespace-labelselector/README.md
@@ -1,0 +1,252 @@
+# Namespace `LabelSelector` for `ConfigurationPolicies`
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Implement `MatchLabels` and `MatchExpressions` (collectively known as `LabelSelector`) for
+`ConfigurationPolicies` in order to further filter the list of namespaces to which an object should
+be applied.
+
+## Motivation
+
+Currently the `ConfigurationPolicy` has a `spec.namespaceSelector` that includes `include` and
+`exclude` arrays of strings that optionally accept a leading and/or trailing wildcard to filter by
+the name of the namespace. In order to more closely align with Kubernetes use cases, the proposal is
+to implement a `LabelSelector` along with these arrays in order to filter namespaces by their label.
+The `LabelSelector` struct includes `MatchLabels` and `MatchExpressions`. In addition to aligning
+with Kubernetes practices, this will cut down on processing the namespace list for clusters with a
+large number of namespaces since a shorter list would be returned from the Kubernetes API.
+
+Futhermore, an additional proposal is to use the `filepath` Go library rather than a custom wildcard
+implementation for the `include` and `exclude` lists. This will bring maintainability and additional
+power to the namespace filtering. This will not be a breaking change since the current behavior is
+to only support leading and trailing wildcards, and a pattern like `foo*bar` is treated as a literal
+string, which is not a valid namespace name.
+
+### Goals
+
+1. Filter the namespace list by `LabelSelector` for the `NamespaceSelector`
+2. Allow `LabelSelector` and `include`/`exclude` arrays to operate together when both are provided
+   or separately when provided individually
+3. Replace custom wildcard code with `filepath` (See
+   [filepath](https://pkg.go.dev/path/filepath#Match) for reference)
+
+## Proposal
+
+Implement `MatchLabels` and `MatchExpressions` alongside the `include` and `exclude` arrays:
+
+```yaml
+namespaceSelector:
+  include: []
+  exclude: []
+  matchLabels: {}
+  matchExpressions: []
+```
+
+These can easily be used to construct a `LabelSelector` for the Kubernetes API to consume.
+
+The `MatchLabels` and `MatchExpressions` are AND-ed by default, so the implementation will carry
+this forward to find the intersection of the `LabelSelector` and, if provided, the `include` and
+`exclude` arrays. Either the `include` and `exclude` arrays or the `LabelSelector` can be provided.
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/62-namespace-labelselector/README.md
```diff
@@ -0,0 +1,199 @@
...
+  matchExpressions: []
+```
+
+These can easily be used to construct a `LabelSelector` for the Kubernetes API to consume.
+
+The `MatchLabels` and `MatchExpressions` are AND-ed by default, so the implementation will carry
+this forward to find the intersection of the `LabelSelector` and, if provided, the `include` and
+`exclude` arrays. Either the `include` and `exclude` arrays or the `LabelSelector` can be provided.
+If the `LabelSelector` is not provided, all namespaces are retrieved and filtered with the `include`
+and `exclude` arrays.
```
JustinKuli, line 59: Maybe just to clarify, this could link to https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector, where the "usual" behavior of `MatchLabels` and `MatchExpressions` is documented.

Also, in that doc it notes:
> An empty label selector matches all objects. A null label selector matches no objects.

Would we preserve that behavior?

enhancements/sig-policy/62-namespace-labelselector/README.md
```diff
@@ -0,0 +1,199 @@
...
+power to the namespace filtering. This will not be a breaking change since the current behavior is
+to only support leading and trailing wildcards, and a pattern like `foo*bar` is treated as a literal
+string, which is not a valid namespace name.
+
+### Goals
+
+1. Filter the namespace list by `LabelSelector` for the `NamespaceSelector`
+2. Allow `LabelSelector` and `include`/`exclude` arrays to operate together when both are provided
+   or separately when provided individually
+3. Replace custom wildcard code with `filepath`
```
JustinKuli, line 39: To clarify, this could link to https://pkg.go.dev/path/filepath#Match, which documents the rules the wildcards would follow.

enhancements/sig-policy/62-namespace-labelselector/README.md
```diff
@@ -0,0 +1,199 @@
...
+objects by the namespace label and then filter out by name using filepath patterns.
+
+```yaml
+namespaceSelector:
+  exclude: ["test[2-3]"]
+  matchLabels:
+    all: namespaces
+```
+
+**Result**: `test2`, `test3`
```
JustinKuli, line 108: I think you mean `include` here, or the result should be `test1`, `test4`, `test5`? Or I'm missing something

enhancements/sig-policy/62-namespace-labelselector/README.md
```diff
@@ -0,0 +1,199 @@
...
+namespaceSelector:
+  matchLabels:
+    name: test2
+  matchExpressions:
+    key: name
+    operator: In
+    values: ["test1", "test2"]
+```
+
+**Result**: `test2`
```
JustinKuli, line 81: From this example, it looks like if `include` is not provided, it will act like `include: ["*"]` - is that right? And if `exclude` is not provided, it acts like an empty list.

I'm not positive if that matches with the current behavior. I like the (potentially new) behavior in this example. I think we just need to clarify how this will be handled in a backwards compatible way. Eg, will `include` default to `["*"]` only if `matchLabels` or `matchExpressions` are specified?

JustinKuli (commented):
Great examples!

enhancements/sig-policy/62-namespace-labelselector/README.md
dhaiducek: Sort of, yes. If `namespaceSelector: {}` is provided, no namespaces will be passed in. But `include`/`exclude` will continue to function even if `MatchLabels`/`MatchExpressions` is not provided (this is the exception to where the implementation differs from the `LabelSelector` behavior documented).

enhancements/sig-policy/62-namespace-labelselector/README.md
dhaiducek: Good catch! Here I was thinking I was being tricky...and apparently I was. 😅

enhancements/sig-policy/62-namespace-labelselector/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+```yaml
+namespaceSelector:
+  include: []
+  exclude: []
+  matchLabels: {}
+  matchExpressions: []
+```
+
+These can easily be used to construct a `LabelSelector` for the Kubernetes API to consume.
```
JustinKuli, line 160: In this section, instead of repeating the YAML format, can we show what the struct will look like in Go, since that is more detailed. Possibly also how the `LabelSelector` will be created from that struct.

The final working implementation doesn't need to be here, I just think it could help prevent surprises.

dhaiducek: I'll add it to the implementation details since the `struct` is a bit more technical in nature.

dhaiducek: Oh, actually, I didn't even realize that wasn't even the section we're in! Well shoot. Maybe I'll take this section out? Either that or move content from Implementation Details...

dhaiducek: Okay, it's gone. 😄

enhancements/sig-policy/62-namespace-labelselector/README.md
dhaiducek: Explanations added!

mprahl (approved):
Nice job!

/hold for other discussion

mprahl (approved):

enhancements/sig-policy/62-namespace-labelselector/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Summary
+
+Implement `MatchLabels` and `MatchExpressions` (collectively known as `LabelSelector`) for
+`ConfigurationPolicies` in order to further filter the list of namespaces to which an object should
+be applied.
+
+## Motivation
+
+Currently the `ConfigurationPolicy` has a `spec.namespaceSelector` that includes `include` and
```
gparvin, line 20: Note that this shouldn't really be limited to `ConfigurationPolicy` since this feature is also needed for the certificate policies too.  If we move to a common library for our controllers it may fix this anyway, but wanted to make sure we aim for some consistency here since this is really needed for cert policies too.

PR #65, opened 2022-06-21
Running deployments of addons agents outside managed cluster

zhujian7:
Signed-off-by: zhujian <jiazhu@redhat.com>


enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Motivation
+
+Currently, all addons agents should be deployed on the managed cluster, it is mandatory to expose the hub cluster to 
+the managed cluster even if the managed cluster resides in an insecure environment.
+
+The proposal [hosted-deploy-mode](./../33-hosted-deploy-mode/README.md) introduced a new deploy mode `Hosted` to deploy 
+klusterlet outside of the managed cluster, but keeps the addons agents remaining on the managed cluster, this proposal 
+makes the addon framework support running addons agents outside of the managed cluster, which is defined as `Hosted` 
+mode deploy for addons agents.
```
qiujian16, line 24: we should clarify what hosted mode is. hosted cluster is a managedcluster of the hub, which is to host the agent to manage another managedcluster, is this correct?

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+### For addon developers
+
+We propose to add a label `addon.open-cluster-management.io/manifest-type` for the addon agent manifests to mark which 
+manifests should be deployed where in Hosted mode:
+- No matter in Default mode or Hosted mode, all manifests will be deployed on the managed cluster
+
+- When the label does not exist or its value is `hosted-mode-managed`: the manifest will be deployed on the managed 
+cluster in Hosted mode
+- When the label's value is `hosted-mode-hosting`: the manifest will be deployed on the hosting cluster in Hosted mode
```
qiujian16, line 81: this could be confusing, we should clarify what is hosting cluster and what is hosted cluster.

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+</table>
+
+### For addon deployers
+
+Once the managed cluster is imported to the hub in `Hosted` mode, we propose to add an annotation 
+`addon.open-cluster-management.io/hosting-cluster-name` for `ManagedClusterAddon` to 
+indicate the addon should be deployed in Hosted mode, and the value of the annotation is the hosting cluster which 
+should:
+- be a managed cluster of the hub as well
+- be the same cluster where the managed cluster klusterlet(registration-agent & work-agent) runs
```
qiujian16, line 131: what does this mean? I think you have a pre-requirement that the addon in hosted mode MUST run in a managedcluster where klusterlet is deployed using hosted mode. Also the hosting cluster of the klusterlet and addon must be the same, right?

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### How does the addon agent connect to the managed cluster?
+
+Addon agents deployments will be deployed in the `{ManagedClusterAddon.Spec.InstallNamespace}` namespeace, and the 
+addon deployer is required to provide a secret, which name is `{addon-name}-managed-kubeconfig` and contains the key
```
qiujian16, line 136: define what is addon deployer.

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Guidance for transforming Default mode addon to hosted mode
+
+Here are some guidelines for transforming addons, which are built with addon-framework in Default mode, to Hosted mode. 
+After transforming, the addon should be deployed successfully either in Default mode or Hosted mode.
+
+1. Consider carefully whether the addon can run in Hosted mode. Will addon functions be affected when addon agent is 
+running outside the managed cluster? If the answer is no, stop.
+2. Refactor the code of the addon agent to accept a flag like `managed-kubeconfig`, and use the kubeconfig received 
+with the flag to build the managed cluster kube client, and use this client to process resources on the managed cluster.
```
qiujian16, line 208: this should be different from the `local-kubeconfig` or `in-cluster kubeconfig` which is mainly used for leader election, and lease update. right?

qiujian16 (commented):
It needs a explanation on the procedure of how to deploy and addon on hosted mode. To my understanding.

1. user creates a managedcluster addon with 'hosting cluster' annotation
2. there needs to be some check on whether the hosting cluster is valid and reflect on addon conditions
3. addon-framework reads the annotation where is hosting cluster
4. addon-framework read manifests and decide which should be put on hosting and hosted cluster.

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+runs
+- After the addon is deployed, the deployment mode cannot be switched
+
+## Design Details
+
+### For addon developers
+
+We propose to add a label `addon.open-cluster-management.io/manifest-type` for the addon agent manifests to mark which 
+manifests should be deployed where in Hosted mode:
+- No matter in Default mode or Hosted mode, all manifests will be deployed on the managed cluster
```
JustinKuli, line 77: From the table below, I think you mean "in Default mode, all manifests will be deployed on the managed cluster" (which preserves the way it currently works?)

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+    <td class="tg-c3ow"></td>
+    <td class="tg-c3ow"></td>
+    <td class="tg-c3ow"></td>
+  </tr>
+</tbody>
+</table>
+
+### For addon deployers
+
+Once the managed cluster is imported to the hub in `Hosted` mode, we propose to add an annotation
```
JustinKuli, line 126: Is the annotation something that would be added automatically when the managedcluster is in Hosted mode? Or would this be specified by the user / another process when creating the ManagedClusterAddOn resource?

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Graduation Criteria
+
+Alpha:
+* Hosted mode addon is implemented and works well on KinD cluster for at least one addon
+* E2E tests exists
+
+## Alternatives
+
+1. Change the ManagedClusterAddon API to add a field `DeployOption` to describe the deploy option(mode) of the addon.
```
JustinKuli, line 260: I think the field would go well in the spec, it seems similar to the `installNamespace` that is already there.

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+Some potential ways to provide the secret:
+- The addon deployer asks the managed cluster administrator for a specific permission kubeconfig, then create it on the 
+hosting cluster. Because the kubeconfig is generated by the managed cluster administrator, the addon developer does not 
+need to deploy any service-account, role, rolebinding on the managed cluster, labels these resources 
+`addon.open-cluster-management.io/manifest-type: hosted-mode-none`
+- The addon developer deploys service-account, role, rolebinding on the managed cluster, labels these resources 
+`addon.open-cluster-management.io/manifest-type: hosted-mode-managed`(or does not add this label, keep this label 
+non-exist), then develops a component to get the service account on the remote managed cluster, and retrieve the token 
+back to the hosting cluster to generate a kubeconfig.
```
JustinKuli, line 148: It seems like it would be very useful if this was done by the addon-framework (or it had functions to help with the process at least)

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+       |       | | clusterClaim        |   | role,rolebinding,sa |  etc.    |                                  
+       |       | | appliedmanifestwork |   +----------^----------+          |                                  
+       |       | +---------------------+              |                     |                                  
+       |       +--------------------------------------|---------------------+                                  
+       |                                              |                                                        
+       |                                              |                                                        
+       -----------------------------------------------|                                                        
+```
+
+### Guidance for transforming Default mode addon to hosted mode
```
JustinKuli, line 200: This example/guide is very helpful. I think we had done some initial tests with the policy addon using this kind of volumeMount and kubeconfig arg, and it worked reasonably well.

enhancements/sig-architecture/63-hosted-addon/README.md
JustinKuli: Great point! If I remember right, the addon-framework usually uses a lease for addon health on the managed cluster. In Hosted mode, I think that would be on the hosting cluster? This is another place where the footprint on the managed cluster is reduced 👍 

But also, there is a backup way to put the lease on the hub cluster, in case the managed cluster doesn't support that API (eg openshift 3.11). Would that need to be considered here as well, for a potential 3 kubeconfigs? Or can we require the hosting cluster to support that API?

enhancements/sig-architecture/63-hosted-addon/README.md
zhujian7: yes, will add more explanation on the hosting cluster Term, and we do not use`hosted cluster` as a term in OCM to describe the hosted mode, instead we use `managed cluster`.

enhancements/sig-architecture/63-hosted-addon/README.md
zhujian7: The annotation will not be added automatically, it should be specified by the user / another process when creating the ManagedClusterAddOn resource. since even the managedcluster is in Hosted mode, the addon still can be in Default mode.

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+manifests should be deployed where in Hosted mode:
+- No matter in Default mode or Hosted mode, all manifests will be deployed on the managed cluster
+
+- When the label does not exist or its value is `hosted-mode-managed`: the manifest will be deployed on the managed 
+cluster in Hosted mode
+- When the label's value is `hosted-mode-hosting`: the manifest will be deployed on the hosting cluster in Hosted mode
+- When the label's value is `hosted-mode-none`: the manifest will not be deployed in Hosted mode
+
+
+<table class="tg">
```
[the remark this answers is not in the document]
zhujian7: But it is not easy to describe with markdown

enhancements/sig-architecture/63-hosted-addon/README.md
zhujian7: yes, Great point. I think the lease should be on the hosting cluster.

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,262 @@
...
+- The hosting cluster must be a managed cluster of the hub, hosting cluster should be managed by the hub
+- The hosting cluster must be the same cluster where the managed cluster klusterlet(registration-agent & work-agent) 
+runs
+- After the addon is deployed, the deployment mode cannot be switched
+
+## Design Details
+
+### For addon developers
+
+We propose to add a label `addon.open-cluster-management.io/manifest-type` for the addon agent manifests to mark which
```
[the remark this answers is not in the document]
zhujian7: `addon.open-cluster-management.io/hosted-manifest-location` looks good, @qiujian16 What is your point?

enhancements/sig-architecture/63-hosted-addon/README.md
zhujian7: Currently, I don't think there is a suitable component for this. This requires the component can connect to the managed cluster naturally and has access to get the service account secret back to the hosting cluster. It looks like only registration-agent and work-agent can meet the condition, but the feature is clearly not what they were designed for, the registration-agent is for registry a cluster/an addon to the HUB, and the work-agent is for applying manifestwork to a managed cluster.

enhancements/sig-architecture/63-hosted-addon/README.md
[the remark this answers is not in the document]
qiujian16: i think you only want to know whether the resource should be put on hosting or hosted. In that case, why do we need to care the value? You only need to care about the annotation key existence, right?

qiujian16 (commented):
only one thing that needs to be addresses. Can use mutate the annotation on adoon resource, for instance, remove the annotation or change to another hosting cluster?

zhujian7:
@JustinKuli PTAL

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Design Details
+
+### For addon developers
+
+We propose to add a label `addon.open-cluster-management.io/hosted-manifest-location` for the addon agent manifests to 
+mark which manifests should be deployed where in Hosted mode:
+- No matter what the label's value is, all manifests will be deployed on the managed cluster in Default mode.
+
+- When the label does not exist: the manifest will be deployed on the managed cluster in Hosted mode
```
JustinKuli, line 81: Could we also have a specific value like `managed` that does this? I think a developer could want to specify that they are deliberately placing something on the managed cluster. If the only way is to not specify a label, then it's not clear if the behavior is on purpose or not.

JustinKuli (commented):
Seems good in general

enhancements/sig-architecture/63-hosted-addon/README.md
zhujian7: Great point, I agree. @qiujian16 WDYT?

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,274 @@
...
+### User Stories
+
+#### Story 1
+
+As a managed cluster administrator, I hope as few as possible workloads running in my cluster while the addon 
+functionality is not affected.
+
+#### Story 2
+
+As a hub cluster administrator, on the premise that the addon functionality is not affected, I do not want the addon
```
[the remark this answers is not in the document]
zhujian7: fixed

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+Addon agents deployments will be deployed in the `{ManagedClusterAddon.Spec.InstallNamespace}` namespeace, and the 
+addon deployer is required to provide a secret, which name is `{addon-name}-managed-kubeconfig` and contains the key 
+`kubeconfig` to connect to the managed cluster, in the `{ManagedClusterAddon.Spec.InstallNamespace}` namespace on the 
+hosting cluster.
+
+Some potential ways to provide the secret:
+- The addon deployer asks the managed cluster administrator for a specific permission kubeconfig, then create it on the 
+hosting cluster. Because the kubeconfig is generated by the managed cluster administrator, the addon developer does not 
+need to deploy any service-account, role, rolebinding on the managed cluster, labels these resources
```
mprahl, line 145: Did you mean this?
```suggestion
need to deploy any service-account, role, or rolebinding on the managed cluster. These resources would be labeled with 
```

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,276 @@
...
+- After the addon is deployed, the deployment mode cannot be switched
+
+## Design Details
+
+### For addon developers
+
+We propose to add a label `addon.open-cluster-management.io/hosted-manifest-location` for the addon agent manifests to 
+mark which manifests should be deployed where in Hosted mode:
+- No matter what the label's value is, all manifests will be deployed on the managed cluster in Default mode.
+
```
xuezhaojun, line 79: The following content is related with PR: https://github.com/open-cluster-management-io/registration-operator/pull/253  @elgnay @zhujian7 @qiujian16 

---

Developers should at least add a namespace resource with the label `"addon.open-cluster-management.io/namespace": "true"` for addons' hosted mode manifests.

Because manifests in hosted mode could be installed in any namespace specified by `{ManagedClusterAddon.Spec.InstallNamespace}` which may not be created or synced with image pull secret.

Here is an example:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: {{ .Release.Namespace }}
  labels:
    addon.open-cluster-management.io/namespace: true

Comment (zhujian7) [reply enhancements/sig-architecture/63-hosted-addon/README.md:79]: Developers are required to ensure the existence of the addon agent namespace. if there is the label `"addon.open-cluster-management.io/namespace": "true"` on the namespace, the addon-framework will copy the image pull secret to the namespace, otherwise, developers need to provide the image pull secret themselves.

File: enhancements/sig-architecture/63-hosted-addon/README.md [hunk 17]
```diff
@@ -0,0 +1,276 @@
...
+    <td class="tg-c3ow"></td>
+  </tr>
+</tbody>
+</table>
+
+### For addon deployers
+
+Once the managed cluster is imported to the hub in `Hosted` mode, the addon deployers can choose to run addon in Hosted
+mode or Default mode, By default the addon agent will run on the managed cluster(Default mode), the addon deployers can 
+add an annotation `addon.open-cluster-management.io/hosting-cluster-name` for `ManagedClusterAddon` so that the addon
```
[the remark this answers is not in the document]
zhujian7: please see https://github.com/open-cluster-management-io/enhancements/pull/65#discussion_r908216507

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,276 @@
...
+- The managed cluster(Klusterlet) must be imported to hub in `Hosted` mode
+- The hosting cluster must be a managed cluster of the hub
+- The hosting cluster of the addon must be the same hosting cluster of the klusterlet
+- After the addon is deployed, the deployment mode cannot be switched
+
+## Design Details
+
+### For addon developers
+
+We propose to add a label `addon.open-cluster-management.io/hosted-manifest-location` for the addon agent manifests to
```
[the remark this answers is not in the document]
zhujian7: the value of each manifest if determined by the addon developer

enhancements/sig-architecture/63-hosted-addon/README.md
xuezhaojun: @zhujian7  Agree， clear and simple!

enhancements/sig-architecture/63-hosted-addon/README.md
```diff
@@ -0,0 +1,276 @@
...
+addon deployer is required to provide a secret, which name is `{addon-name}-managed-kubeconfig` and contains the key 
+`kubeconfig` to connect to the managed cluster, in the `{ManagedClusterAddon.Spec.InstallNamespace}` namespace on the 
+hosting cluster.
+
+Some potential ways to provide the secret:
+- The addon deployer asks the managed cluster administrator for a specific permission kubeconfig, then create it on the 
+hosting cluster. Because the kubeconfig is generated by the managed cluster administrator, the addon developer does not 
+need to deploy any service-account, role, rolebinding on the managed cluster, labels these resources 
+`addon.open-cluster-management.io/hosted-manifest-location: none`
+- The addon developer deploys service-account, role, rolebinding on the managed cluster,labels these resources
```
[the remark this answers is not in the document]
zhujian7: Sorry, this may be confusing. I mean once the addon developer adds the label addon.open-cluster-management.io/hosted-manifest-location: managed or keeps the label non-exist, the addon-framework will deploy the resource to the managed cluster.

enhancements/sig-architecture/63-hosted-addon/README.md
xuezhaojun: Maybe we can addon some context before?

---

If manifests will be deployed on the hosting cluster in Hosted mode,  then Developers are required to ensure the existence of the addon agent namespace(for example, add a namespace YAML file into manifests). 

And developers also should add the label "addon.open-cluster-management.io/namespace": "true" on the namespace, otherwise, addon-framework will not copy the image pull secret to the namespace and developers need to provide the image pull secret by themselves.

enhancements/sig-architecture/63-hosted-addon/README.md
zhujian7: @qiujian16 should we change the `{addon-name}-managed-kubeconfig` to `{addon-name}-spoke-kubeconfig`?

qiujian16:
/approve
/hold 

hold for 24 hours for any further comment

PR #69, opened 2022-07-11
Initial pass at stories for token registration

dangorst1066:
Hey @mikeshng @skeeey @qiujian16 - as per discussion on Slack, firing over a PR covering the admin developer experience of joining/accepting clusters using token registration - just to get the conversation started. Technical details/impl notes/testing etc still need to be added to the doc

### DIFF
--- a/enhancements/sig-architecture/68-token-registration/README.md
+++ b/enhancements/sig-architecture/68-token-registration/README.md
@@ -22,7 +22,7 @@ As such, OCM needs to support an alternative registration mechanism to CSRs that
 
 ### Goals
 
-- The OCM hub cluster can run on EKS, and in other environments where CSR registration cannot be used.
+- The OCM hub cluster can run on EKS and other environments where CSR registration cannot be used.
 - `clusteradm` tooling provides options to select token based registration when joining a worker to the hub.
 - Suitable warnings/error messages/conditions are added to OCM such that it is easily identifiable when CSRs cannot be used.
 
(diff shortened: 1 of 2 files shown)


qiujian16:
Thanks!

enhancements/sig-architecture/68-token-registration/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+There needs to be a procedure to follow covering scenarios where the spoke cluster is unable to refresh its service account token, and has lost its ability to authenticate with the hub.
+It should be possible for administrators to restore functionality to the spoke cluster.
+
+Some example scenarios:
+
+1. OCM agents in the spoke cluster were offline (e.g. due to an outage) during which its service account token expired
+2. A network outage resulted in the spoke cluster loosing connectivity to the hub api server for an extended period of time
+3. A service account was intentionally deleted (e.g. the associated token was compromised) and replaced.
+
```
qiujian16, line 100: we should have another story about addons. I think in token base approach, addons also need to use token to talk to the hub.

qiujian16:
alternatively, we could build a front-proxy for kube-apiserver, the front-proxy can use mtls with a customized signer.

enhancements/sig-architecture/68-token-registration/README.md
dgorst: Added 👍

enhancements/sig-architecture/68-token-registration/README.md
```diff
@@ -0,0 +1,157 @@
+## Release Signoff Checklist
```
deads2k, line 1: Please add https://github.com/open-cluster-management-io/api/blob/main/docs/clusterjoinprocess.md as a see-also.  It may be worth it to move the doc into this repo.

That doc outlines the current process and a few key features of the flow are visible there:
1. the subject has a group associated with the managedcluster and the name has an agent associated with the process on that managedcluster
2. an agent on the managedcluster is able to get new agent credentials by using shared (non-agent) credentials to create a request and the hub admin can approve or reject.
3. an agent on the managedcluster is able renew credentials to identify itself.  This also allows for non-shared credential flows.
4. different agents on the managed cluster are members of the same group and have different names, with the agent clearly identified
5. the hub admin gets a choice about whether or not to allow a particular managecluster to get a valid credential
6. the hub admin can control expiry/removal of a particular agent.  In this case, by rotating the signing credential.  (Different distros are better or worse at doing this.)

Any new flow that is developed should allow the same capabilities and should use the standard group and usernames to ease adoption.  Capability 2 may be negotiable (not required), but it would be limiting because the hub would need to communicate unique credentials to each managed cluster.

dgorst: 👍 Added to implementation details section

PR #70, opened 2022-08-03
Spread policy in Placement

huaouo:


enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,274 @@
...
+    order: 2
+```
+
+### Implementation
+Two types of spread constraints require different logics. Hence, they will be implemented separately. Affinity constraints are implemented by a `Prioritizer` to calculate the scores of the clusters using the information conveyed via `topologyWeights`. However, it seems that the scheduler's current design of the plugin framework makes it difficult to express the logic of even spreads. Therefore, an extra `Selector` plugin type is proposed.
+
+**Implementation of (Cluster) Affinity/Anti-affinity.** An `AffinitySpreadPriorititizer` will be implemented as follows:
+
+1. for each cluster, score it according to `topologyWeights` in the cluster affinity rules
+2. normalize the cluster affinity scores to [0, 100]
```
haoqing0110, line 240: How will this score work together with other Prioritizers? For example:
```
spreadConstraints:
  - type: Affinity
    topologyKey: region
    topologyKeyType: Label
    topologyWeights:
    - weight: 20
      operator: In
      values:
      - us-east-1
      - cn-hangzhou
    - weight: 10
      operator: In
      values:
      - us-west-1
 prioritizerPolicy:
    mode: Exact
    configurations:
      - scoreCoordinate:
          builtIn: ResourceAllocatableMemory
```
Will cluster in us-west-1 with much more ResourceAllocatableMemory has a higher score than cluster in us-east-1?

huaouo: I consider not to include the Affinity and Even constraints in Additive mode. In Exact mode, users may need to specify the weight of them manually. The above example that only contains affinity constraints will be like:

```yaml
spreadConstraints:
  - type: Affinity
    topologyKey: region
    topologyKeyType: Label
    topologyWeights:
    - weight: 20
      operator: In
      values:
      - us-east-1
      - cn-hangzhou
    - weight: 10
      operator: In
      values:
      - us-west-1
 prioritizerPolicy:
    mode: Exact
    configurations:
      - scoreCoordinate:
          builtIn: ResourceAllocatableMemory
      - scoreCoordinate:
          builtIn: ClusterAffinitySpread
        weight: 2
```

In this scenario, if we only have one cluster in each region mentioned above (i.e., cluster A in us-west-1 and cluster B in us-est-1), then the affinity prioritizer will score cluster A to be 100 and cluster B to be 50 after normalization. If the memory resource score of cluster A and cluster B are α and β, the final score of them will be α+200 (for cluster A) and β+100 (for cluster B).

The proposal will be updated soon to contain examples about the weight of Affinity and Even.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,301 @@
...
+	// +required
+	TopologyKey string `json:"topologyKey"`
+
+	// TopologyKeyType indicates the type of TopologyKey. It could be Label or Claim.
+	// +required
+	TopologyKeyType string `json:"topologyKeyType"`
+
+	// TopologyWeights indicates the optional weights of topologies, valid when Type
+	// is Affinity.
+	TopologyWeights []TopologyWeightTerm `json:"topologyWeights,omitempty"`
```
qiujian16, line 121: I think this should be handled differently. I think what you want is to put as many as possible decisions in the certain set of clusters. Mixing this with spreading is confusing.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,301 @@
...
+  prioritizerPolicy:
+    mode: Exact
+    configurations:
+      - scoreCoordinate:
+          builtIn: ResourceAllocatableMemory
+      - scoreCoordinate:
+          builtIn: ClusterAffinitySpread
+        weight: 2
+      - scoreCoordinate:
+          builtIn: ClusterEvenSpread
```
qiujian16, line 222: hrm, if this is in the prioritizer, what is the relationship of this and the weight in topologyWeight

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+	// MaxSkew describes the degree to which the workload may be unevenly distributed,
+	// valid when Type is Even. If omitted, the scheduler treat the even constraint a soft
+	// constraint.
+	MaxSkew *int32 `json:"maxSkew,omitempty"`
+
+	// Order indicates the order of even constraints. It is valid when Type is Even and
+	// required only when multiple even constraints are used. To spread the workload,
+	// the scheduler takes the even constraints with a smaller order value into prior consideration.
+	Order int32 `json:"order,omitempty"`
```
qiujian16, line 131: order seems strange. If we finally calculate this as a prioritizer, should this be weight?

qiujian16 (commented):
After reading the whole doc, I think we should put affinity in scope, or at least uses a different field. Mixing spread and affinity could be confusing to users IMO.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: The initial rationale of mixing affinity and even spreading is to force users to think and write the Placement API following the TopologyKey. If we separating affinity and even spreading, a user might submit an object like:

```
kind: Placement
metadata:
  name: demo-placement
spec:
  # ...
  affinity:
    clusterAffinity:
      labelSelector:
        matchExpressions:
        - key: provider
          operator: In
          values:
          - AWS
  spreadConstraints:
  - topologyKey: provider
    topologyKeyType: Label
  # ...
```
If we have both AWS and GCP clusters, it is confusing to both the scheduler and the user himself/herself, as the spread constraint and the affinity rule specify opposite semantics towards the same topologyKey.

By mixing affinity and even spreading, the SpreadConstraintsTerms in the SpreadConstraints array will be ensured to have different topologyKeys using api validation. As each SpreadConstraintsTerm can be of either Even OR Affinity type, the problem may be solved.

I'm aware that such a design seems complex and differs with K8s's API design. If my motivation is unreal, affinity and even spreading can be certainly separated. And This proposal might focus on even spreading and put affinity out of scope.

qiujian16: I think we should separate them. I feel that the API validation will be similar and it is complicated when we need to treat matchExpression and even more complicated scenario (e.g. including predicate). This might be surfaced by events or condition in the placement instead of using validation.

huaouo: OK, I'll edit the proposal and take your suggestions.

enhancements/sig-architecture/70-spread-policy/metadata.yaml
```diff
@@ -0,0 +1,14 @@
...
+reviewers:
+  - "@qiujian16"
+  - "@elgnay"
+  - "@haoqing0110"
+approvers:
+  - "@qiujian16"
+  - "@elgnay"
+  - "@haoqing0110"
+creation-date: 2022-08-03
+last-updated: 2022-08-03
```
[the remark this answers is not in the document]
huaouo: Done.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,301 @@
+Spread Policy across Failure-domains in Placement APIs
+---
+
+## Summary
+This document proposes relevant API and code changes to apply for the OCM project [Topology-aware scheduling in cloud-native multi-cluster/multi-cloud environments](https://summer-ospp.ac.cn/#/org/prodetail/221390086) in OSPP 2022. This work is based on [previous community discussions](https://github.com/open-cluster-management-io/community/issues/49#issuecomment-857743027) on spread policy with several modifications.
```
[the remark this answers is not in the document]
huaouo: Done.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,301 @@
...
+### Use Case 3: Max Skew
+
+When the even spread is unsatisfiable to a certain degree, we may want the scheduler not to schedule the workload.
+
+**Example.** We want a even spread across the zones in us-east-1 (i.e., us-east-1a and us-east-1b) and four clusters in total. If us-east-1a can provide three clusters and us-east-1b can only provide one cluster, then the even spread cannot be satisfied strictly.
+
+Following K8s's PodTopologySpread, we define the skew of a topoloy to be selected_clusters_in_current_topology - **min** selected_clusters_in_a_topology. For the above example, max skew is 3 - 1 = 2. If we want the max skew to be smaller than 2, then the scheduler should report an error.
+
+### Use Case 4: (Cluster) Affinity/Anti-affinity Spread
+
```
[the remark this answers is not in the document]
huaouo: `maxSkew` is a hard constraint if not omitted. If the skew of the final scheduling result is higher than `maxSkew`, the scheduler is expected not to schedule the workload.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,301 @@
...
+## Proposal
+
+### API Spec
+```go
+type PlacementSpec struct {
+	// ...
+
+	// SpreadConstraints defines how placement decision should be distributed among a
+	// set of clusters.
+	SpreadConstraints []SpreadConstraintsTerm `json:"spreadConstraints,omitempty"`
```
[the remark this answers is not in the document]
huaouo: Done.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: I have changed this field to `weight`. But the value of this field is designed to represent the hierarchical structure of topology keys, and is not used directly for calculating scores.

For example, we have three topology keys: `provider`, `region` and `zone`. Their weight are a, b, and c, respectively. Then any combination of three numbers that ensures a > b > c is considered appropriate.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: `topologyWeight` is used for cluster affinity which has been excluded from this proposal.

enhancements/sig-architecture/70-spread-policy/README.md
qiujian16: yeah, your description indicates that is could either be weight or priority.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+1. The result cluster set = {}
+2. Call `selectOne` n times, in each time insert the returned cluster into the result cluster set
+3. Return the result cluster set
+
+`selectOne`:
+
+1. For each cluster, calculate a score (even_score) according the benefit of selection of it to balancing the skewness. If the selection of a cluster would cause the violation to `maxSkew`, exclude it from the later selecting phases. If all clusters are excluded, report a error.
+2. For each cluster, calculate its final score = u*even_score + its_score_from_prioritizers, where u is a parameter.
+3. Sort the clusters by final scores and return the one with the highest final score.
```
qiujian16, line 226: let's define the graduation criteria here.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,226 @@
...
+    # regions. 
+  - topologyKey: provider
+    topologyKeyType: Label
+    maxSkew: 2
+    weight: 20
+    # maxSkew is omitted as this term is a soft constraint
+  - topologyKey: region
+    topologyKeyType: Label
+    weight: 10
+  # I plan to exclude the spread policy in Additive mode.
```
qiujian16, line 173: what is the reason for this?

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,226 @@
...
+  predicates:
+  - # ...
+  # SpreadPolicy contain two terms, which work
+  # independently and jointly.
+  spreadPolicy:
+    # When multiple terms are used, the user should specify their weights.
+    # The scheduler will consider to first do the spread according to the
+    # terms with higher weight.
+    # In this example, the scheduler will first spread the workload
+    # by providers.
```
qiujian16, line 162: could you explain what happens when there is a higher weight of soft constraint and lower weight of hard constraint? It seems that hard constraint will always have the highest weight?

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,250 @@
...
+    maxSkew: 2
+    weight: 20
+    # maxSkew is omitted as this term is a soft constraint
+  - topologyKey: region
+    topologyKeyType: Label
+    weight: 10
+  # I plan to exclude the spread policy in Additive mode.
+  # In Exact mode, users need to specify the weight of them manually.
+  # The spread policy only work when there is a SpreadPolicy coordinate
+  # in prioritizerPolicy.
```
haoqing0110, line 176: The prioritizers are some soft constraints, while SpreadPolicy could be both hard&soft. 
If user doesn't define `builtIn: SpreadPolicy`, should spreadPolicy with hard constraints work?

huaouo: The hard decision logic of the spread policy is applied to the final score of a cluster, which is a weighted sum of the scores from all the  prioritizers and the score from the proposed spread policy selector. So, if we want a spread policy to work properly as either a hard or a soft constraint, a weight of it is needed.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: Hmmm, in my opinion it is rather a advanced feature. If we include it into Additive mode, we need to provide a default weight, which may be hard to set.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: Actually, a spread policy with `maxSkew` is both a hard and a soft constraint. The hard part prevent some illegal set of clusters to be scheduled. The soft part then do the best-effort spread. For the soft part, we still need a weight.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: A spread policy term with `maxSkew` can be considered as both a soft and a hard constraint. The soft part does the *best-effort* spread.  The hard part prevents some extreme cases that violate `maxSkew`. The hard part *does* have the highest weight semantically. But here the `weight` field is given weight to the soft parts and terms without `maxSkew`.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: Actually, a spread policy term with `maxSkew` can be considered as both a soft and a hard constraint. The soft part does the *best-effort* spread. The hard part prevents some extreme cases that violate `maxSkew`. So, we still need weights for the soft parts.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: Done.

enhancements/sig-architecture/70-spread-policy/README.md
qiujian16: By default, the plugin could be disabled, The reason of using additive is if we do not need to add other plugin (balance or steady) when setting this plugin

enhancements/sig-architecture/70-spread-policy/README.md
qiujian16: I am not quite sure about the meaning of soft and hard part. When a policy is set, it must be either hard or soft constraint, right? a policy cannot be both hard and soft. If maxSkew is not set, doest it mean it is a hard constraint, and setting weight for this policy is meaningless?

huaouo: In this proposal, when `maxSkew` is not set, it means `maxSkew` is infinity. If users want to set a hard limit to an spread policy term, they need to explicitly set  `maxSkew` like `maxSkew: 1`. This design is welcome for comments.

Now let's consider a spread policy term with a hard limit like `maxSkew: 10`. If we have two regions `ra`, `rb`, and each of them provides enough cl;usters. We want to select 11 clusters in total for our workload. The hard limit prevents a placement like `ra: 11 clusters, rb: 0 cluster`, but it doesn't try the best to give an even spread like `ra: 6 clusters, rb: 5 clusters`. The placement `ra: 10 clusters, rb: 1 cluster` is also valid for the hard limit. To give an even spread, we still need to *score* the clusters. Therefore, there still needs a weight for the even spread scores when summing them with other scores from prioritizers .

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+	// +kubebuilder:validation:MaxLength=316
+	TopologyKey string `json:"topologyKey"`
+
+	// TopologyKeyType indicates the type of TopologyKey. It could be Label or Claim.
+	// +required
+	// +kubebuilder:validation:Required
+	TopologyKeyType TopologyKeyType `json:"topologyKeyType"`
+
+	// MaxSkew describes the degree to which the workload may be unevenly distributed.
+	// If not omitted, the scheduler would not schedule if MaxSkew cannot be satisfied.
```
deads2k, line 144: If present, the scheduler will not schedule if the MaxSkew cannot be satisfied.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+      - scoreCoordinate:
+          builtIn: SpreadPolicy
+        weight: 2
+```
+
+### Implementation
+The current scheduling framework contains two types of plugins, i.e., `Filter`s and `Prioritizer`s. 
+`Filter`s are invoked *before* `Prioritizer`s to handle hard constrains, filtering the clusters. 
+Then, the `Prioritizer`s score the clusters.
+To implement `maxSkew`, we need to first score the clusters (i.e., do the spread), then ensure the `maxSkew` is not exceeded by some filter-like operations for hard constraints.
```
deads2k, line 220: What filter are you thinking of?  The total number of filtered results is not applied until after prioritization.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+This cannot be implemented in the current framework.
+Hence, a new plugin type, `Selector`, is proposed to implement hard constrains *after* `Prioritizer`s.
```
deads2k, line 222: why wouldn't the spread identify all groups of targets and then use the prioritizer to order inside that group?  For instance,

1. there are 10 clusters in each aws zone
2. the filters reduce the set to 7 clusters in each aws zone
3. the spread finds 5 equivalent clusters to choose
4. the prioritizer orders those 5

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: Placement
+metadata:
+  name: demo-placement
+spec:
+  # numberOfClusters is required if spreadConstraints contains
+  # a constraint of Even type.
+  numberOfClusters: 4
```
deads2k, line 179: is number of clusters logically required to satisfy the skew?  I don't think it is, but it makes picking the values to put there slightly more difficult.

enhancements/sig-architecture/70-spread-policy/README.md
deads2k: Ok, I think I see.  If the prioritizer runs before the Selector, then the selector can rely on "is earlier in the list" to decide among equivalent clusters.  After prioritization, the clusters will be interleaved, but it doesn't matter because the selector can choose properly.

Running after prioritization helps keep the layers clean.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+// PluginSelectResult contains the details of a select plugin result.
+type PluginSelectResult struct {
+	// Selected contains the selected ManagedCluster.
+	Selected []*clusterapiv1.ManagedCluster
+	// Err contains the select plugin error message.
+	Err error
+}
+```
+
+Then, a `EvenSpreadSelector` will be implemented as follows to select n clusters.
```
deads2k, line 243: Is there a PR showing this algorithm.  It seems plausible in broad strokes, but the details are a little hard in psuedo code.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+⇥5.3 iterate over the sorted spread policies, for each one as `curPolicy`:  
+⇥⇥5.3.1 `curSelectedCount := selectedCount[curPolicy.topologyKey]`  
+⇥⇥5.3.2 `minSelectedCount := min{ values in curSelectedCount }`  
+⇥⇥5.3.3 iterate over the available clusters, for each one as `curCluster`:  
+⇥⇥⇥5.3.3.1 `clusterSpreadScores[curCluster] <<= 1`  
+⇥⇥⇥5.3.3.2 `if curSelectedCount[curCluster.labelOrClaims[curPolicy.topologyKey]] == minSelectedCount: clusterSpreadScores[curCluster] += 1`  
+⇥5.4 normalize the scores in `clusterSpreadScores` to [0, 100]  
+
+// **STEP 3: sum the spread scores and the scores from Prioritizers, and select the cluster with highest final score**  
+⇥5.5 calculate final score of each `curCluster` by `clusterSpreadScores[curCluster]` * the weight of spread policy in prioritizerPolicy + the score of `curCluster` from the Prioritizer phase, and pick the cluster (as `pickedCluster`) with the highest final score
```
deads2k, line 272: prioritizePolicy or the weight from the spreadconstraintsterm?

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: If numberOfClusters is not specified, the behavior of the scheduler may be trying to select all clusters which meet the placement requirements, and check the skew. I agree that it could be optional.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: The code is expected to be available in 1~2 weeks.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: In STEP 3, we need a final weighted sum of the scores from all prioritizers and the score calculated by spread policy, so the weight is the one defined in prioritizerPolicy.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: Examples for filters here is Predicate and TaintToleration.

enhancements/sig-architecture/70-spread-policy/README.md
huaouo: Done :)

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+6 return `resultClusterSet`  
+
+### Graduation Criteria
+
+#### Alpha
+1. Demonstrate proposed features at the community meeting (last week in Sept 2022);
+2. The new APIs are reviewed and accepted.
+
+#### Beta
```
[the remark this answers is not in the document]
huaouo: The simulator is added into graduation criteria.

enhancements/sig-architecture/70-spread-policy/README.md
deads2k: > The code is expected to be available in 1~2 weeks.

checking back to see if there's a WIP/draft branch.

huaouo: > > The code is expected to be available in 1~2 weeks.
> 
> checking back to see if there's a WIP/draft branch.

Sorry for late reply, the draft implementation is available at https://github.com/huaouo/ocm-placement, and there is a demo video at https://www.youtube.com/watch?v=j7ycK5zaaXA.

enhancements/sig-architecture/70-spread-policy/README.md
```diff
@@ -0,0 +1,261 @@
...
+	// +kubebuilder:validation:Minimum=1
+	MaxSkew *int32 `json:"maxSkew,omitempty"`
+
+	// Weight indicates the weight of the SpreadPolicyTerm. It is required only when multiple
+	// terms are used. To spread the workload, the scheduler takes the even constraints with
+	// higher weight into prior consideration.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default=1
+	Weight int32 `json:"weight,omitempty"`
```
haoqing0110, line 156: In the current design, the spreadPolicy has 2 weight, one is the “prioritizer” weight, another is the “SpreadConstraints” [weight](https://github.com/open-cluster-management-io/enhancements/pull/70/files#diff-03ac0f150ce968ff17c6967fd1616808dea99017fe744703bc4aba9c1f97f931R156) .

Since the “SpreadConstraints” weight only impact the order/priority when selecting clusters in the [implementation](https://github.com/open-cluster-management-io/placement/compare/main...huaouo:ocm-placement:main#diff-51801281ce54e097a0a393a3b8ca7b65121c521de25bc6a70bdc725264dc944dR49) . I’m thinking if we can remove this “SpreadConstraints” weight, just use the nature order of SpreadConstraints array to select clusters. It’s easy to understand and avoid bringing too much “weight” in placement.

deads2k:
> Since the “SpreadConstraints” weight only impact the order/priority when selecting clusters in the [implementation](https://github.com/open-cluster-management-io/placement/compare/main...huaouo:ocm-placement:main#diff-51801281ce54e097a0a393a3b8ca7b65121c521de25bc6a70bdc725264dc944dR49) . I’m thinking if we can remove this “SpreadConstraints” weight, just use the nature order of SpreadConstraints array to select clusters. It’s easy to understand and avoid bringing too much “weight” in placement.

Perhaps it would be helpful to include an examples section that includes
1. if `numberOfClusters` is not specified, but the `maxSkew` is specified, what happens.  I think the spreadPolicy ends up limiting the number of clusters returned to honor the `maxSkew`.  Is there any way for a user to know this happened?
2. if the motivation layout had another provider "weirdProvider" where "region-1" and "region-2" had zone names that overlapped (region-1,zone-A, region1,zone-B, and region2,zone-A), would the there be a way to spread on zones per region?  I think this design requires every key to have unique values without a way to describe a tuple.  I'm ok with this since compound keys could be built via a controller, but I'd like to be explicit about the limitation.
3. it's not clear to me if the spread weight is the prioritizer.weight*spread.weight.  An example using the cluster layout from the motivation could clarify this and indicate how it will function.  The implementation looks to me as though it would weight based on order of weights, not based on how different the weight is. For instance, weights 5 and 6 would have the same effective weighting in the spread calculation as 20, 10.

Overall I like the API.

huaouo:
> > Since the “SpreadConstraints” weight only impact the order/priority when selecting clusters in the [implementation](https://github.com/open-cluster-management-io/placement/compare/main...huaouo:ocm-placement:main#diff-51801281ce54e097a0a393a3b8ca7b65121c521de25bc6a70bdc725264dc944dR49) . I’m thinking if we can remove this “SpreadConstraints” weight, just use the nature order of SpreadConstraints array to select clusters. It’s easy to understand and avoid bringing too much “weight” in placement.
> 
> Perhaps it would be helpful to include an examples section that includes
> 
> 1. if `numberOfClusters` is not specified, but the `maxSkew` is specified, what happens.  I think the spreadPolicy ends up limiting the number of clusters returned to honor the `maxSkew`.  Is there any way for a user to know this happened?
> 2. if the motivation layout had another provider "weirdProvider" where "region-1" and "region-2" had zone names that overlapped (region-1,zone-A, region1,zone-B, and region2,zone-A), would the there be a way to spread on zones per region?  I think this design requires every key to have unique values without a way to describe a tuple.  I'm ok with this since compound keys could be built via a controller, but I'd like to be explicit about the limitation.
> 3. it's not clear to me if the spread weight is the prioritizer.weight*spread.weight.  An example using the cluster layout from the motivation could clarify this and indicate how it will function.  The implementation looks to me as though it would weight based on order of weights, not based on how different the weight is. For instance, weights 5 and 6 would have the same effective weighting in the spread calculation as 20, 10.
> 
> Overall I like the API.

For 1, the current implementation will try to select all the clusters if `numberOfClusters` is not specified. Then, if `maxSkew` is specified but cannot be satisfied by the decision including all the clusters, the scheduling will just fail. We can change this behavior to selecting as many as possible clusters, but it cause an inconsistency with the behavior when `numberOfClusters` is specified. The current implementation with specified `numberOfClusters` will also make the scheduling fail if `maxSkew` is specified but cannot be satisfied. Or, we may add a field following [k8s](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#spread-constraint-definition) to let users make their decision about the behavior. The `maxSkew` may contain a field `whenUnsatisfiable`,  which can be:
- `DoNotSchedule`: Try to select the expected number of clusters. If `maxSkew` cannot be satisfied for the expected number of clusters, fail. If `numberOfClusters` is not specified, the expected number of clusters means all the clusters.
- `ScheduleGreedily`: Select as many as possible clusters if `maxSkew` cannot be satisfied for the expected number of clusters. 
- `ScheduleAnyway`: Select the expected number of clusters anyway if `maxSkew` cannot be satisfied but still try to reduce the skew. (I'm doubt if this option is necessary. If users want the `ScheduleAnyway` behavior, why don't them just omit the `maxSkew` field?) 

For 2, for zones with the same name inside different regions (i.e., overlapping), they are considered different zones in the current implementation, and the spread will work correctly. Only with labels on clusters, we cannot describe the tuples. But in the `spreadConstraints` array, the constraints on different topology keys are ordered (by either the `weight`, or the natural order in the array).  So, the implementation can construct the tuples (region-1, zone-A) and (region-2, zone-A), and recognizes that they are different.

For 3, as I described in the "For 2" paragraph, the `weight` of the `spreadConstraints` are used only to construct the tuples, making sure (region-1, zone-A) but not (zone-A, region-1). The scheduler should first do the spread according to the regions and then zones. And yes, the implementation treats weight based on order of weights, not based on how different the weight is. Should we follow the comments from @haoqing0110 to use the natural order instead of the `weight` field?

deads2k:
> For 3, as I described in the "For 2" paragraph, the `weight` of the `spreadConstraints` are used only to construct the tuples, making sure (region-1, zone-A) but not (zone-A, region-1). The scheduler should first do the spread according to the regions and then zones. And yes, the implementation treats weight based on order of weights, not based on how different the weight is. Should we follow the comments from @haoqing0110 to use the natural order instead of the `weight` field?

Weight in the current prioritizer is a coefficient to multiply the priority by in order to make some priorities more important than others.  It's not just about ordering.  If I read this shift operation properly (https://github.com/huaouo/ocm-placement/blob/5946042d775a8338c7c90fa7ca4e4e778191ddc4/pkg/plugins/spread/spread.go#L118), weight in the proposed spread API is not a coefficient indicating how preferred a particular label is, but rather simply an ordering choice.  However, the shift is such a large multiplication that I think it will drown out the other prioritizers.

deads2k:
https://github.com/open-cluster-management-io/enhancements/pull/70#issuecomment-1284993432

Can you provide those examples in the KEP so we'll have them for future reference.  I think the actual prioritizer values may help as we try to describe what happens with the bit shift in the current implementation and its impact on other prioritizers.

huaouo:
> [#70 (comment)](https://github.com/open-cluster-management-io/enhancements/pull/70#issuecomment-1284993432)
> 
> Can you provide those examples in the KEP so we'll have them for future reference. I think the actual prioritizer values may help as we try to describe what happens with the bit shift in the current implementation and its impact on other prioritizers.



> > For 3, as I described in the "For 2" paragraph, the `weight` of the `spreadConstraints` are used only to construct the tuples, making sure (region-1, zone-A) but not (zone-A, region-1). The scheduler should first do the spread according to the regions and then zones. And yes, the implementation treats weight based on order of weights, not based on how different the weight is. Should we follow the comments from @haoqing0110 to use the natural order instead of the `weight` field?
> 
> Weight in the current prioritizer is a coefficient to multiply the priority by in order to make some priorities more important than others. It's not just about ordering. If I read this shift operation properly (https://github.com/huaouo/ocm-placement/blob/5946042d775a8338c7c90fa7ca4e4e778191ddc4/pkg/plugins/spread/spread.go#L118), weight in the proposed spread API is not a coefficient indicating how preferred a particular label is, but rather simply an ordering choice. However, the shift is such a large multiplication that I think it will drown out the other prioritizers.

Actually, the spread score whose calculation involves shifts will be [normalized](https://github.com/huaouo/ocm-placement/blob/5946042d775a8338c7c90fa7ca4e4e778191ddc4/pkg/plugins/spread/spread.go#L149) to `[-100, 100]` before summing it with the scores from prioritizers. This may not be a problem.

huaouo:
Hi @deads2k,

After discussed with @qiujian16, this proposal now has two major changes in the API:
1. The weight in each spread constraint has been removed. Now, the natural order of the spread constraints in the `spreadConstraints` array indicates their relative importance.
2. A new `whenUnsatisfiable` field has been added into each spread constraint to present the behavior of the scheduler when `maxSkew` cannot be satisfied. It can be either `ScheduleAnyway` or `DoNotSchedle`. `ScheduleAnyway` means to still schedule the expected number of clusters when `maxSkew` cannot be satisfied. `DoNotSchedule` means to stop scheduling more clusters when `maxSkew` cannot be satisfied. The final placement decision may contain less clusters but will satisfy `maxSkew`.

Also, the [API proposal](https://github.com/open-cluster-management-io/api/pull/181) has been updated.

qiujian16:
/approve
/lgtm
/hold

I am still not quite sure about natural order of importance for each spread policy. I think we should note this is the current implementation but not necessarily to surface it on API. But this could be a followup.

hold in case you would need more review, otherwise feel free to release.

haoqing0110:
/unhold

Thanks for reviewing @qiujian16 and @deads2k. More comments and discussions are still welcomed.

deads2k:
> 2\. A new `whenUnsatisfiable` field has been added into each spread constraint to present the behavior of the scheduler when `maxSkew` cannot be satisfied. It can be either `ScheduleAnyway` or `DoNotSchedle`. `ScheduleAnyway` means to still schedule the expected number of clusters when `maxSkew` cannot be satisfied. `DoNotSchedule` means to stop scheduling more clusters when `maxSkew` cannot be satisfied. The final placement decision may contain less clusters but will satisfy `maxSkew`.

Thanks, that helps quite a bit.

deads2k:
> An example about the implementation is now available in the proposal in the [Implementation Example section](https://github.com/open-cluster-management-io/enhancements/blob/83475f70ad1e0b7e90a20741a6c5a52dd6612e81/enhancements/sig-architecture/70-spread-policy/README.md#implementation-example).

The example is great!  Extremely h elpful.

Issue #71, opened 2022-08-31
Support automatic Policy reconciliation when a Hub policy template's referenced object updates


mprahl:
Policies support dynamic values through policy templating. One such option is Hub policy templating, which are templates that are executed on the Hub before the policy is distributed to the managed clusters. Hub policy templating supports Kubernetes API queries such as getting a field from a ConfigMap.

The current limitation is that if a referenced object in the Hub policy template is updated, it does not immediately cause the policy to be reconciled with this update. This does happen [every ten hours](https://github.com/kubernetes-sigs/controller-runtime/blob/02dc46444cfe0e8bbe61bb3d4c3a12a0faadcd24/pkg/manager/manager.go#L109-L134) or every time the policy is reconciled such as when the policy status is updated or the policy applies to a new cluster. A workaround exists where the `policy.open-cluster-management.io/trigger-update` annotation can be updated on the policy to cause the policy to be reconciled immediately.

The current options that trigger the policy to be reconciled are insufficient and it should happen automatically and immediately when the referenced object is updated.

PR #72, opened 2022-08-31
Automatic Reconciliation When Hub Template Referred Object is Updated

mprahl:
Relates:
https://github.com/open-cluster-management-io/enhancements/issues/71

Signed-off-by: mprahl <mprahl@users.noreply.github.com>

### DIFF
--- a/enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
+++ b/enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
@@ -1,0 +1,232 @@
+# Automatic Reconciliation When Hub Template Referred Object is Updated
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Policies support dynamic values through policy templating. One such option is Hub policy templating, which are templates
+that are executed on the Hub before the policy is distributed to the managed clusters. Hub policy templating supports
+Kubernetes API queries such as getting a field from a ConfigMap. For example,
+`{{hub fromConfigMap "" "my-configmap" "my-field" hub}}`.
+
+The current limitation is that if a referenced object in the Hub policy template is updated, it does not immediately
+cause the policy to be reconciled with this update. This does happen every ten hours or every time the policy is
+reconciled such as when the policy status is updated or the policy applies to a new cluster. A workaround exists where
+the `policy.open-cluster-management.io/trigger-update` annotation can be updated on the policy to cause the policy to be
+reconciled immediately.
+
+The current options that trigger the policy to be reconciled are insufficient and it should happen automatically and
+immediately when the referenced object is updated.
+
+## Motivation
+
+A common use-case for Hub policy templates is to copy a Secret from the Hub to a managed cluster. If the Secret on the
+Hub is updated, it is desired for the Secret to be immediately updated on the managed cluster. This isn’t possible
+without the `policy.open-cluster-management.io/trigger-update` annotation being manually updated or waiting for up to
+[ten hours](https://github.com/kubernetes-sigs/controller-runtime/blob/02dc46444cfe0e8bbe61bb3d4c3a12a0faadcd24/pkg/manager/manager.go#L109-L134).
+
+### Goals
+
+1. A policy is automatically reconciled when it has a Hub policy template that references an object that was updated.
+1. The implementation efficiently uses Kubernetes API watches to achieve this.
+
+### Non-Goals
+
+N/A
+
+## Proposal
+
+The Policy Propagator, which is the controller on the Hub that executes the Hub policy templates, should watch all
+individual objects that are referenced in Hub policy templates. Whenever a referenced object changes, the policy should
+be reconciled.
+
+### User Stories
+
+#### Story 1
+
+As a Policy addon user, I want my policies to automatically reconcile immediately when a Hub policy template references
+an object that was updated.
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### New Client Library
+
+A new library should be written to dynamically watch objects and notify the consumer through a
+[client-go rate limiting queue](https://pkg.go.dev/k8s.io/client-go@v0.23.5/util/workqueue?utm_source=gopls#NewNamedRateLimitingQueue),
+which is the same queue used by the controller-runtime library. Note that the reason that this is a library is so that
+this functionality can be used in other parts of the Policy addon such as the Configuration Policy controller if it
+becomes event-driven. It could also be useful for the open-source community.
+
+The queue in this new library would output ObjectIdentifier objects (shown below) that reference the watcher object
+(e.g. Policy) that is watching dependent objects (e.g. Secret or ConfigMap) whenever a dependent object changes. The
+[client-go rate limiting queue](https://pkg.go.dev/k8s.io/client-go@v0.23.5/util/workqueue?utm_source=gopls#NewNamedRateLimitingQueue)
+is well suited for this use-case because it only maintains a queue of unique objects, so if a watched object updates
+several times in a short period of time, this will greatly reduce the number of queue entries and therefore, greatly
+reduce the work on the consumer side.
+
+As shown below, the
+[client-go rate limiting queue](https://pkg.go.dev/k8s.io/client-go@v0.23.5/util/workqueue?utm_source=gopls#NewNamedRateLimitingQueue)
+will be abstracted away and the consumer of the library would implement a struct using the Reconciler interface. The
+Reconcile method would behave exactly as the controller-runtime Reconcile method, but it instead receives an
+ObjectIdentifier object rather than just the namespace and name. The consumer of the library can easily add or update
+watches by calling the AddOrUpdateWatcher method and can remove watches with the RemoveWatcher method.
+
+```go
+import (
+   "context"
+
+   "k8s.io/client-go/tools/cache"
+   "sigs.k8s.io/controller-runtime/pkg/ratelimiter"
+   "sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Reconciler interface {
+   Reconcile(ctx context.Context, watcher ObjectIdentifier) (reconcile.Result, error)
+}
+
+
+// Options are the arguments for creating a new DynamicWatcher.
+type Options struct {
+   // Reconciler reconciles the watched object.
+   Reconciler Reconciler
+
+   // RateLimiter is used to limit how frequently requests may be queued.
+   // Defaults to controller-runtime's MaxOfRateLimiter which has both overall and per-item rate limiting.
+   // The overall is a token bucket and the per-item is exponential.
+   RateLimiter ratelimiter.RateLimiter
+}
+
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+policy template is updated, the root Policy would be reconciled. This would have the same logic as the first event loop,
+and thus this new Reconcile method would just wrap the Reconcile method from the first event loop. Note that since these
+event loops will be run as different Goroutines, locking should be added so that only one root Policy object is
+reconciled at a time.
+
+### Risks and Mitigation
+
+One risk is that creating too many Kubernetes watch API requests could be expensive since we would need to create a
+watch request per object to avoid client-side filtering. Measuring the CPU usage of the Kubernetes API once this is
+implemented is important. If the CPU usage is too high, we could measure the performance of client-side filtering to see
```
gparvin, line 169: If a customer determines this feature is causing an undesirable amount of CPU usage, will there be a way to disable the automatic reconcile and rely only on the current annotation?

mprahl: I think it's a good idea. I'll see about making this configurable but have the default be the new behavior.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+## Motivation
+
+A common use-case for Hub policy templates is to copy a Secret from the Hub to a managed cluster. If the Secret on the
+Hub is updated, it is desired for the Secret to be immediately updated on the managed cluster. This isn’t possible
+without the `policy.open-cluster-management.io/trigger-update` annotation being manually updated or waiting for up to
+[ten hours](https://github.com/kubernetes-sigs/controller-runtime/blob/02dc46444cfe0e8bbe61bb3d4c3a12a0faadcd24/pkg/manager/manager.go#L109-L134).
+
+### Goals
+
+1. A policy is automatically reconciled when it has a Hub policy template that references an object that was updated.
```
yahliu, line 36: For managed side template policy,  is it supported ?

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+The Policy Propagator, which is the controller on the Hub that executes the Hub policy templates, should watch all
+individual objects that are referenced in Hub policy templates. Whenever a referenced object changes, the policy should
+be reconciled.
+
+### User Stories
+
+#### Story 1
+
+As a Policy addon user, I want my policies to automatically reconcile immediately when a Hub policy template references
+an object that was updated.
```
yahliu, line 54: References objects are various. If the references object is an image and it's updated, can policies automatically reconcile immediately with new image ? ( e.g  delete the original pod and create a pod with new image)

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+this, the [go-template-utils](https://github.com/stolostron/go-template-utils) library would need to be updated to
+return the objects it queried for on the API. The reconcile loop would then collect references to all objects that were
+queried for using Hub policy templates for that root Policy object and call AddOrUpdateWatcher method from the library.
+When the root Policy object is deleted, the RemoveWatcher method from the library is called.
+
+The second (new) event loop would be the event loop from the new library, where anytime an object referenced in a Hub
+policy template is updated, the root Policy would be reconciled. This would have the same logic as the first event loop,
+and thus this new Reconcile method would just wrap the Reconcile method from the first event loop. Note that since these
+event loops will be run as different Goroutines, locking should be added so that only one root Policy object is
+reconciled at a time.
```
yahliu, line 163: Since used to  `policy.open-cluster-management.io/trigger-update` annotation in previous release, seems the status message of the policy no updated, until user manually check the resources on the managed clusters.  So in this function,  after reconciled,  will there be a detailed status message shown in the UI results page, that can tell user the policy already be applied new objects ?

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
mprahl: It wouldn't be since on the managed side, policies are usually evaluated every 10-20 seconds, so the extra overhead isn't worth it in my opinion.

If a user did have a longer evaluation interval on their policy than the default, then I think it would be a nice feature, but I'd rather defer this for now.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
mprahl: That's a good question. By quickly looking at the code, it seems that if a policy was previously compliant and needed to perform an update to remain compliant, the message doesn't indicate that.

Is that what you are seeing? I do think there's value in updating the status message in these cases but I think it's out of scope for this feature.

yahliu: So in large clusters,  It doesn't make sense that user manually check the resources in the managed clusters after the referred objects are updated and there is no clear sign,  they don't know when should they check after one minute or 5 minutes ?   for example, the resources still not be used new objects after 10 mins, how and where they should check ?  
“immediately updated” is generic statement.  Does it specific time for this ? or user can customize the time ？

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
yahliu: I think this should be doc'd in the 2.7 doc.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+No need to outline all of the test cases, just the general strategy. Anything that would count as tricky in the
+implementation and anything particularly challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage expectations).
+
+All code is expected to have sufficient e2e tests.
+
+### Graduation Criteria
+
+It would be GA in the release after implementation.
```
qiujian16, line 197: we should consider performance test before it is GAed.

drkho: Hi, QE can also plan for testing the specific objects commonly used by customers. Can more info be provided here? re: "...Anything that would count as tricky in the
implementation and anything particularly challenging to test should be called out."?

Also, what is the jira item or git issue for this feature?

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
JustinKuli, line 163: Why is this locking necessary? I would understand needing to lock whatever structures are used for mapping policies <-> objects in templates, but I don't see why the whole reconcile needs to be locked.

For example, if a secret is used in two policies, I don't see why those would need to be reconciled sequentially when the secret is changed.

mprahl: It's not time based. It's event-driven. So as soon as the referenced object is updated, that event triggers the policy to be reconciled.

I'll check with the squad on why there is no status update message when an object is updated when it was previously compliant. Like I said before, this is out of scope for this epic, but we should still think about this.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+#### New Client Library
+
+A new library should be written to dynamically watch objects and notify the consumer through a
+[client-go rate limiting queue](https://pkg.go.dev/k8s.io/client-go@v0.23.5/util/workqueue?utm_source=gopls#NewNamedRateLimitingQueue),
+which is the same queue used by the controller-runtime library. Note that the reason that this is a library is so that
+this functionality can be used in other parts of the Policy addon such as the Configuration Policy controller if it
+becomes event-driven. It could also be useful for the open-source community.
+
+The queue in this new library would output ObjectIdentifier objects (shown below) that reference the watcher object
```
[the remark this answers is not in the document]
mprahl: It's only the objects referenced by Hub policy templates (e.g. `{{hub fromSecret "" "my-secret" "my-field" hub}}`) that would be watched. In most cases, the same object is used for a single policy but different fields are used for each cluster.

So far, from what I've seen in production deployments, it's been well below 100 objects, but I think it's worth testing the load on the Hub Kubernetes API when there are 1000 objects being watched.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
mprahl: I agree that performance testing should be done before it can be GA. I think the backup plan of if it's too performance intensive is to make this opt-in.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
mprahl: The lock would be at the root policy level and not based on the referenced object.

What I want to try to avoid is two goroutines updating the same policy and causing one's update to fail because a replicated policy was updated by the other goroutine since the replicated policy was pulled from the cache. This would then cause unnecessary log noise and additional execution of code that isn't necessary.

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
mprahl: > Hi, QE can also plan for testing the specific objects commonly used by customers. Can more info be provided here? re: "...Anything that would count as tricky in the implementation and anything particularly challenging to test should be called out."?
> 
> Also, what is the jira item or git issue for this feature? Is this for https://issues.redhat.com/browse/ACM-1208?

That is the correct associated Jira.

I would say the most common scenarios are a user copying a secret from the Hub to a managed cluster using the `fromSecret` Hub template function.

Another common scenario is a `ConfigMap` that contains variable configuration (e.g. an operator channel) per managed cluster. For example:

```yaml
apiVersion: v1
kind: ConfigMap
....
data:
  cluster1-gitops-channel: gitops-1.6
  cluster2-gitops-channel: gitops-1.5
  cluster3-gitops-channel: stable
```

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy. Anything that would count as tricky in the
+implementation and anything particularly challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage expectations).
+
+All code is expected to have sufficient e2e tests.
```
dhaiducek, line 193: Should this boilerplate text be replaced?

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
mprahl: I added this to the "Risks and Mitigation" section.

mprahl:
Could you all please take another look?

enhancements/sig-policy/71-hub-templates-auto-reconcile/README.md
drkho: fyi, @yahliu ⬆️

gparvin:
/hold
Hold for more approvals

PR #74, opened 2022-09-23
design ordering policy execution

willkutler:
https://github.com/stolostron/backlog/issues/25611

this is a draft, feel free to add some review! want to wait until @JustinKuli gets back next week to finalize anything.


enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,209 @@
...
+in a desired order and guarantee that this order will be stable
+2. Opt-in: any extra fields that are used to specify policy ordering should be optional,
+and policies that do not contain these fields should function as they did before in order
+to avoid issues with older policies
+
+### Non-Goals
+
+1. Status message matching: we plan to only allow users to conditionally process policy
+templates based on compliance state, so users will not be able to specify that they want a
+policy to be processed if another policy
```
gparvin, line 44: You left me hanging here.  Could it be a NonCompliance status message like a CRD lookup failure would not be treated differently than a NonCompliance for a missing CR -- those could mean very different things but are treated both as NonCompliant.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,209 @@
...
+4. As an author of many ConfigurationPolicies, I have several policies that
+depend on another policy. I want to be able to reference completely separate
+policies (outside of the Policy “wrapper”) as required before a configuration
+is applied. (Maybe allow a policy to reference the status of a policy that runs
+before it (success or failure) before it tries to process it. Allowing you to chain
+them together)
+5. As a cluster administrator, I want to do a declarative action on a managed
+cluster when a policy becomes non-compliant. So I want to “activate” a
+ConfigurationPolicy based on the state of another policy in the cluster.
+
```
gparvin, line 77: Can we call out a user story for the generator so we don't forget to add support there?

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,209 @@
...
+are in a pending state, so that if a dependency is no longer satisfied, any objects created
+on the cluster as a result of that dependency will be deleted.
+
+### Open Questions
+
+How will the "pending" status for policy templates that have not had their conditions met work?
+We know we want to set the compliance of the policy template object to "pending", but different
+policy templates have differently arranged status fields, so there is not a standard field
+to add a message saying why the evaluation of the policy is pending. We also need to investigate
+if this status will have any issues showing up in the hub cluster templates, and how it will look
```
gparvin, line 160: Will the status sync be able to flow the pending status to the hub?  Or is there no status for pending policies which today shows up as the yellow triangle.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,209 @@
...
+as `pending`.
+
+```
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: my-complicated-policy
+spec:
+  dependencies: //for all policy templates
+    - kind: Policy
```
gparvin, line 117: Can this be a PolicySet?

gparvin (commented):
This looks really good.  Added a few basic comments

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,209 @@
...
+## Design Details
+
+Here is what the new spec field will look like in a policy. The top-level
+`dependencies` field applies to all policy templates in the policy, while
+the template-level `extraDependencies` field can be used to set up
+dependencies for a specific policy template. Like the policy namespace
+selector, the top-level `dependencies` will take priority over the 
+template-level `extraDependencies`, so policy templates that have
+`extraDependencies` satisfied but not `dependencies` will still be marked
+as `pending`.
```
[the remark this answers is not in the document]
willkutler: I believe we used to have a field like this a while ago and it has been removed, but I had thought some older policies still used it. I'll just remove the analogy

enhancements/sig-policy/74-order-policy-execution/README.md
willkutler: @gparvin I think the pending status would make it to the hub, since we will be setting it on the replicated policies in the managed cluster. However, I think it might be best to still show the yellow triangle - that icon means the policy hasn't been processed on a cluster yet, which is pretty much true if some of its templates are in a pending state. @JustinKuli what do you think?

JustinKuli: I agree. It shouldn't be a problem for status sync to tell the hub that the policy is pending, and it should provide details on what it's waiting on.

I also agree that the yellow triangle icon would fit here, and we will need to add UI logic to ensure that is what's used while the policy is pending.

willkutler:
@gparvin @JustinKuli I think I should be done with the changes, anything else I should change/add?

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+ConfigurationPolicy based on the state of another policy in the cluster.
+6. As a user of the policy generator, I want to be able to generate policies that
+will have their templates be processed in a certain order.
+
+### Implementation Details/Notes/Constraints 
+
+We will leverage the new dynamic watcher library (enhancement 71) to watch all
+policy template dependencies. When a policy is created with a dependency, we
+can start to watch the policy object that it is dependent on. If the compliance
+of the dependency matches up with what the policy wants, we will add process it
```
mprahl, line 88: Did you mean this?
```suggestion
of the dependency matches up with what the policy wants, we will process it
```

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+policy to be processed if another policy has a specific status message
+
+## Proposal
+
+We would add an optional new `dependencies` field to the spec on the configuration 
+policy level as well as an `extraDependencies` field on the template level to
+allow users to specify conditions for policy processing on both the policy and
+template level. When these fields are set and their conditions are not met, 
+the policy templates will not be applied on the managed cluster and the affected
+policy will enter a "pending" state. If this happens, any policy sets that contain
```
mprahl, line 55: How will this new state be reflected in the `policy_governance_info` metric exposed by the Policy Propagator? If I recall correctly, the compliance state is represented by a gauge.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+can start to watch the policy object that it is dependent on. If the compliance
+of the dependency matches up with what the policy wants, we will add process it
+normally in the template sync. If not, we will assign a `Pending` state to the
+policy compliance, and the watcher library will notify us when the dependency
+changes, so we can re-evaluate the decision. This logic will be added into the
+new combined governance-policy-framework-addon repository.
+
+These dependencies will be able to accept any object that has a
+`status.complianceState` field - policies, policy sets, all current policy
+types, and anything that might use the policy nucleus in the future. If an
```
mprahl, line 96: I think a reference (e.g. design spec) to the policy nucleus would be nice. Alternatively, this mention could be removed.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+object in `dependencies` or `extraDependencies` does not contain that status
```
mprahl, line 97: I'm not sure I like this because the Policy API has `omitempty` set on the `status.complianceState` field. This could lead to the weird situation where it goes from non-compliant to pending.

I would prefer to just keep it in pending and have a different status message based on if the field is set.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+(500+ configuration policies) it could take over a minute for the dependent policy
+status to update correctly
+
+## Design Details
+
+Here is what the new spec field will look like in a policy. The top-level
+`dependencies` field applies to all policy templates in the policy, while
+the template-level `extraDependencies` field can be used to set up
+dependencies for a specific policy template. The top-level `dependencies`
+will take priority over the template-level `extraDependencies`, so policy
```
mprahl, line 123: Do you mean that the arrays will be merged?

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+afterward the state of the cluster changes and those dependencies are no longer met, we do
+not want the template to be active anymore. Utilizing the watch library should allow us to
+follow the dependencies for each template and put the template back into a pending state
+when this happens. When the template is put into a pending state, the configuration policy
+will be deleted, so if a dependency is no longer satisfied, any objects created
+on the cluster as a result of that dependency will be deleted.
+
+### Open Questions
+
+Is setting an `apiGroup` field in `dependencies/extraDependencies` necessary? Currently we should
```
mprahl, line 172: What if we add a new alpha or beta policy type? Is the plan to use [restmapper](https://pkg.go.dev/k8s.io/client-go/restmapper)?

enhancements/sig-policy/74-order-policy-execution/README.md
willkutler: @mprahl my thinking was we would ignore pending policies in the metric as they are essentially disabled until their dependencies are satisfied

enhancements/sig-policy/74-order-policy-execution/README.md
willkutler: I think the best way to handle this would be if we want to add a new alpha or beta policy type, we will also add the `apiGroup` field here

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,219 @@
...
+- Update: We do not plan to add `apiGroup` now as it can be added in a backwards-compatible way
+in the future if we feel it is needed.
+
+### Test Plan
+
+- Unit tests can be implemented to determine whether the logic that enables/disables policy
+templates based on their dependencies is working properly
+- E2E/integration tests will be necessary to ensure that synced objects are being managed
+properly and policy status is what it should be when dependencies are met/not met.
+
```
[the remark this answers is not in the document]
JustinKuli: I think that detecting when there is a circular dependency, and giving a special status message for that situation is more complex than necessary. I think that the "pending" messages will be helpful enough that a user can figure it out on their own...

Maybe we need an open question: if we have several compliant policies, and then we edit them to become circular dependencies, what should happen? If we're supposed to detect the situation, then they would all become pending. If we don't actively detect it, they would remain compliant until one changes state, and then I think they would all become pending. Is either of these situations better than the other?

I can't imagine a useful circular dependency as-is, so I think we could detect the issue and give a special status.

**_If_** dependencies weren't just on one status (eg if I could define a policy to be active if another policy was *not* Compliant - meaning Pending, NonCompliant, Unknown, or Terminating) then maybe having a circle could be meaningful? Then the whole circle of policies might not necessarily go into a Pending status all at once and get stuck.

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: Good question. I agree with Will, I think that when the policy is Pending, we would treat it as disabled for the purposes of that metric. Likely we should improve the metric's logic to better handle "unusual" statuses - I don't think it would handle an Unknown state very well either (not sure how common that is though). For reference: https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/controllers/policymetrics/policymetrics_controller.go#L128

I think we're considering using a warning icon in the UI when a policy is Pending, and I think "disabling" the metric in this situation would match the behavior in other situations where that icon is used. So the UI and this metric would match, which would be good.

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: @mprahl what do you mean by new alpha or beta policy type?

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,227 @@
...
+## Proposal
+
+We would add an optional new `dependencies` field to the spec on the configuration 
+policy level as well as an `extraDependencies` field on the template level to
+allow users to specify conditions for policy processing on both the policy and
+template level. When these fields are set and their conditions are not met, 
+the policy templates will not be applied on the managed cluster and the affected
+policy will enter a "pending" state. If this happens, any policy sets that contain
+the pending policy will also be marked as pending. When the conditions are met or
```
JustinKuli, line 56: I had an extra thought around when to put the whole policy into a Pending state: since it could be that only one policy template is Pending, but others are Compliant and NonCompliant, which state gets the highest priority? As-is, it sounds like the Pending status would override any others, but I wonder if NonCompliant should be the most important...

willkutler: thinking about it more, you are probably right - I think users would want to know if any policies are noncompliant at a glance even if parts of them are still pending. but I think we should prioritize pending over compliant, since we don't want to report compliant if some templates have not been processed yet

enhancements/sig-policy/74-order-policy-execution/README.md
mprahl: This sounds good to me.

enhancements/sig-policy/74-order-policy-execution/README.md
mprahl: @JustinKuli say I create a new policy with the kind `OperatorPolicy` and it's in the `policy.open-cluster-management.io` `apiGroup` but it's a `beta1` version. How will this be handled by the dependency resolution? When reading this doc, I got the impression that `v1` was always implied.

enhancements/sig-policy/74-order-policy-execution/README.md
mprahl: I agree with what @willkutler said.

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: Ah, ok. Good point: since we want to support PolicySet, and that's only `v1beta1` right now, we probably already need an apiGroup field here.

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: One more extra thought (sorry): if I make a policy that I want to "activate" when another thing is NonCompliant (or something), does it make sense to treat that as "Pending" in the same way as the others?

I imagined a policy like that would be used to do some additional automatic remediation on something, in which case it's "normal" state would probably be non-active. So should it be preventing the parent Policy from becoming Compliant?

What if (optionally) a pending template could be ignored in the Policy's compliance? Like:
```
apiVersion: policy.open-cluster-management.io/v1
kind: Policy
metadata:
  name: my-complicated-policy
spec:
  policy-templates:
    - extraDependencies:
        - kind: ConfigurationPolicy
          name: bar
          compliance: NonCompliant
      ignorePending: true // <===== New <=====
      objectDefinition:
        apiVersion: policy.open-cluster-management.io/v1
        kind: ConfigurationPolicy
        metadata:
          name: fix-bar-in-foo
        spec:
          remediationAction: enforce
          ...
```

enhancements/sig-policy/74-order-policy-execution/README.md
mprahl: Maybe the field should be `apiVersion` to match what you specify in a Kubernetes manifest?

enhancements/sig-policy/74-order-policy-execution/README.md
willkutler: yeah, I like that idea @JustinKuli . I definitely think there are some scenarios where we basically want to ignore a policy template unless something bad happens, and in those cases we wouldn't want it messing up the status.

willkutler:
Thanks to everyone who contributed in the sync call! Feel free to add any other comments. We discussed adding the following changes to the doc:
- should the compliance field be made optional? we decided to keep it
- should policy sets be ordered instead of basing ordering around policy dependencies? we decided to keep the current approach as it allows for more flexibility
- should a namespace field be added in the dependencies map? yes, we will add a namespace field and the namespace logic will handle replicated policy names so users do not have to include the namespace in the policy name
- create a simple dependency example that Christian can show to customers for feedback

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: In the call today I said `PolicySet` would work... but those aren't copied to managed clusters, are they? So they probably can't be used as a dependency.

JustinKuli: @mprahl shared the idea that the propagator could see a PolicySet as a dependency, and replace it with the list of Policies in that set. We might need to be careful to keep that replacement updated if the PolicySet changes, but I think overall it would work.

I was concerned it might be confusing if the replicated policy and the root policy had different specs, but we already have that happen with hub templates, so it's really not that bad. And it would be more clear what exactly the policy is waiting on (a specific policy, not a PolicySet that the user would need to look up)

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -158,12 +158,14 @@ spec:
...
       kind: Policy
       name: namespace-foo-setup-policy
+      namespace: policies
       compliance: Compliant
   policy-templates:
     - extraDependencies: //for this policy template only
         - apiVersion: policy.open-cluster-management.io/v1
           kind: ConfigurationPolicy
           name: bar
+          namespace: local-cluster
```
mprahl, line 168: Let's not allow or ignore namespaces to be set for policy templates.

mprahl: I'm not sure about allowing `<namespace>.<policy-name>` for `Policy`. My thinking was any dependency on an object defined on the Hub would require a namespace (Policy/PolicySet) and anything only ever created on the managed cluster (ConfigurationPolicy/IAMPolicy/etc) would not have a namespace.

dhaiducek (commented):
Late to the party, but this looks great to me, and I appreciate the thorough descriptions on the variety of scenarios this would encounter/affect (like how compliance should most appropriately propagate). Well done!

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,250 @@
...
+### Non-Goals
+
+1. Status message matching: we plan to only allow users to conditionally process policy
+templates based on compliance state, so users will not be able to specify that they want a
+policy to be processed if another policy has a specific status message
+
+## Proposal
+
+We would add an optional new `dependencies` field to the spec on the configuration 
+policy level as well as an `extraDependencies` field on the template level to
```
yahliu, line 51: 1. New `dependencies` field is only for configuration policy ? or all types policy ?  Please clarify.
2. Will be implemented these two new fields `dependencies` and `extraDependencies` on UI policy creation wizard ?  Just like 2.6  deletion objects feature (Prune Object Behavior)

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,250 @@
...
+allow users to specify conditions for policy processing on both the policy and
+template level. When these fields are set and their conditions are not met, 
+the policy templates will not be applied on the managed cluster and the affected
+policy will enter a "pending" state. If this happens, any policy sets that contain
```
yahliu, line 55: How is "Pending" status showing on UI ? warning icon? or Is there detailed message to be shown when move the icon ?

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,250 @@
...
+policies (outside of the Policy “wrapper”) as required before a configuration
+is applied. (Maybe allow a policy to reference the status of a policy that runs
+before it (success or failure) before it tries to process it. Allowing you to chain
+them together)
+5. As a cluster administrator, I want to do a declarative action on a managed
+cluster when a policy becomes non-compliant. So I want to “activate” a
+ConfigurationPolicy based on the state of another policy in the cluster.
+6. As a user of the policy generator, I want to be able to generate policies that
+will have their templates be processed in a certain order.
+
```
yahliu, line 82: For mixed policy, how is the feature to handle ?  Two cases: 
1. Two or three types templates in a policy, for example, cert and configuration templates 
2. One cert type policy A is depend on another configuration type policy B.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,250 @@
...
+1. As an author of a ConfigurationPolicy, I don’t want resources to be wasted when
+I know one object depends on another. If an object cannot be cleanly applied before
+another is created, I want to specify this in my Policy, and not waste
+resources (CPU, k8s API requests) in the policy controllers.
+2. As an author of a ConfigurationPolicy, I want to specify the order resources are
+applied in a policy when the API is not declarative. For some APIs (eg OLM), if
+resources are applied in the wrong order they can end up in an unhealthy state
+that I want to avoid.
+3. As a user of PolicySets, I want policies inside a set to be applied in a
```
yahliu, line 69: If one policy is pending state, what's status of the policyset ?

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: 1. It will be on the Policy type, not any of the specific policy types (like ConfigurationPolicy, CertificatePolicy). Thanks for catching that this wasn't clear here.
2. Yes, we plan to add it to the UI

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: The handling of the dependencies will be done by the framework, not by the individual policy controllers. When a dependency is not met, the template will not be created on the managed cluster, and the framework will specially handle populating the status of that template in the Policy. So mixing types in a Policy is no problem, the framework will handle them all the same.

If a cert policy "A" depends on a config policy "B" to be compliant, then the framework will not create "A" on the managed cluster until "B" is compliant. And until then, the framework will mark "A" as `Pending`, and add a status message mentioning that it is waiting on "B".

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: I think it should follow the same rules as the Policies, unless there's a reason to handle it differently? So if any policies in the policyset are NonCompliant, the status will be NonCompliant (regardless of whether any are pending). If the policyset only has Compliant and Pending policies, then it will be Pending. And only if all the policies are Compliant will it be Compliant.

yahliu: OK. I think the priority of states is reasonable.   Can you add the warning message or tip message to prompt in the policyset card if any policy is pending in this policyset?

enhancements/sig-policy/74-order-policy-execution/README.md
JustinKuli: This seems like the last tricky thing to figure out. What if the namespace field is (generally) optional in dependencies, and the propagator cleans up / verifies some things before replicating the policy? So:
- it removes the namespace from ConfigurationPolicy, CertificatePolicy, and IAMPolicy dependencies
- for Policy dependencies without a namespace, it tries to extract the namespace from the name, or it gives an error status
- for Policy dependencies with a namespace, the dependency is already in the "right" format
- for PolicySet dependencies, a namespace is required, or it gives an error status. If the namespace is given, it can expand the PolicySet dependency into a list of Policy dependencies (with the namespace on each)
- for other/unknown types, it just keeps them as defined.

Then on the managed cluster, when checking the dependencies, if the namespace is not defined but the resource is namespaced, then it uses the cluster namespace.

enhancements/sig-policy/74-order-policy-execution/README.md
```diff
@@ -0,0 +1,250 @@
...
+      kind: Policy
+      name: namespace-foo-setup-policy
+      namespace: policies
+      compliance: Compliant
+  policy-templates:
+    - extraDependencies: //for this policy template only
+        - apiVersion: policy.open-cluster-management.io/v1
+          kind: ConfigurationPolicy
+          name: bar
+          namespace: local-cluster
```
serngawy, line 168: I believe it will be better UX , if we simplify the `extraDependencies` by considering the order of the objectDefinations under the `policy-templates->[]objectDefinitions. `  for ex;

```
policy-templates:
-  objectDefinitions:
     apiVersion: policy.open-cluster-management.io/v1
     kind: ConfigurationPolicy
     metadata:
        name: policy1
      spec:
          .....
- ignorePending: false
  compliance: compliant
  objectDefinitions:
     apiVersion: policy.open-cluster-management.io/v1
     kind: ConfigurationPolicy
     metadata:
        name: policy2
      spec:
          .....

```
This way we avoid re-write the configurationPolicy info each time AND it make sense to have a constrain for having configurationPolicies with dependencies must be defined under the same Policy-Templates.

OR

Another way of considering the order is by define order with numbers. Ex; the policy that creates the NS with order value 0 while all other policies with order value 1. so the order of the []objectDefinations doesn't matter (to avoid back compatibility issue ) something like below 

```
policy-templates:
- policyOrder: 0 (default is 0 so all policies have the same order to apply unless user specify)
  objectDefinitions:
     apiVersion: policy.open-cluster-management.io/v1
     kind: ConfigurationPolicy
     metadata:
        name: policy1
      spec:
          .....
- policyOrder: 1
  ignorePending: false
  compliance: compliant
  objectDefinitions:
     apiVersion: policy.open-cluster-management.io/v1
     kind: ConfigurationPolicy
     metadata:
        name: policy2
      spec:
          .....
- policyOrder: 1
  ignorePending: true
  compliance: compliant
  objectDefinitions:
     apiVersion: policy.open-cluster-management.io/v1
     kind: ConfigurationPolicy
     metadata:
        name: policy3
      spec:
          .....
- policyOrder: 2
  ignorePending: true
  compliance: compliant
  objectDefinitions:
     apiVersion: policy.open-cluster-management.io/v1
     kind: ConfigurationPolicy
     metadata:
        name: policy4
      spec:
          .....
```

JustinKuli: These suggestions could simplify the case where templates inside a Policy naturally chain together, but I think they have disadvantages in other cases. I would rather have these kinds of options in the policy generator, and have that tool deal with the tedious "wiring" of the details, if possible.

An example of something I'd like to see, but I don't think would work under either of these suggestions:
- template A has no dependencies
- template B should activate when template A is compliant
- template C should activate when template B is compliant
- template D should activate then template A is terminating
```
A -> B -> C
 `-> D
```

I would also rather only have one way of declaring dependencies: these would remove `extraDependencies` inside each template, but would still have the top-level `dependencies` field.

serngawy: It make sense for Policy dependencies to be defined as mentioned in the example as below ;
```
  dependencies: //for all policy templates
    - apiVersion: policy.open-cluster-management.io/v1
      kind: Policy
      name: namespace-foo-setup-policy
      namespace: policies
      compliance: Compliant
```
however for policy-templates->(configurationPolicies) I still believe it can be simplified similar to my previous comment. 
Just to clarify the case as endUser; if i have ConfigPolicy1 defined in Policy-template1 that will be an extraDependency for  configPolicy2 defined in policy-template2. I would level up policy-template1 to be dependency for policy-template2.

AND agree policyGenerator will definitely help a lot on simplifying the policy creation.

PR #78, opened 2022-10-24
Create a policy-collection repository

gparvin:
Community of resources to help you adopt the policy framework and create:
- sample and best practices policies and policy sets
- generator projects of policies and policy sets
- learning resources
- deployment resources

Signed-off-by: Gus Parvin <gparvin@redhat.com>

### DIFF
--- a/enhancements/sig-policy/78-policy-collection/README.md
+++ b/enhancements/sig-policy/78-policy-collection/README.md
@@ -1,0 +1,75 @@
+# Policy Collection
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+The goal of this feature is to automate and simplify deployment of policies, policy sets and
+policy generator projects by creating (transferring) a repository that provides best practices, samples, blogs
+and deployment resources.  The repository for this policy collection will be
+[policy-collection](https://github.com/open-cluster-management-io/policy-collection/).  The content will be transferred 
+from [existing policy-collection](https://github.com/stolostron/policy-collection/).
+
+## Motivation
+
+The governance policy framework is currently provided without any samples or guidance on
+how to create policies other than by detailing the architecture of the 
+[policy framework](https://open-cluster-management.io/getting-started/integration/policy-framework/).
+While this is functional, it does not lend itself to easy adoption by new community users.
+
+### Goals
+
+- Transfer the policy collection community that provides samples, best practices and encourages quicker
+  adoption of the policy framework.
+- Make sure the transferred community is easy to consume by first time users of ocm policy.
+
+### Non-Goals
+
+- Because policies can be intended to work in many different environments with prerequisites that
+  may not be available, validation of policies is focused on syntax initially and additional validation
+  can be considered in the future.
+
+## Proposal
+
+### User Stories
+
+1. As a k8s admin, I want to deploy policies from the community with an experience that fosters:
+  - Best practice community policies and policy sets to achieve governance goals for common k8s platforms
+  - Donations of policies from users that have forked the community repository and added extra value
+    to existing or created new policies
+  - Quick adoption of the policy framework through blogs and documentation that re-enforces concepts
+    documented through the policy framework
+  - Community adoption.  Validate documentation is centered around the open cluster management 
+    community and not product specific.
+
+### Architecture
+
+The policy-collection does not add any architectural changes to the Policy Framework.  It provides
+samples and resources that help community members adopt the Policy Framework and take advantage of the
+features that are provided.  By showing the existing features, we also hope to make it easier for
+community members to see where new features can be added or where existing enhancements can be made.
+
+### Test Plan
+
+- Basic policy validation will be added
+- Policies can be tested by applying them to a cluster
+- New users should be able to understand how to get started
+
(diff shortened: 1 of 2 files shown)


mprahl (approved):

qiujian16 (commented):
could we consider to put the policy in https://github.com/open-cluster-management-io/governance-policy-nucleus repo as an alternative? We were trying to avoid maintaining many repos. overnance-policy-nucleus is the repo that policy addon will define APIs and libraries. Putting a set of collections of policies in it sounds reasonable to me.

gparvin:
> could we consider to put the policy in https://github.com/open-cluster-management-io/governance-policy-nucleus repo as an alternative? We were trying to avoid maintaining many repos. overnance-policy-nucleus is the repo that policy addon will define APIs and libraries. Putting a set of collections of policies in it sounds reasonable to me.

I think if we start a community around policy controllers then the governance-policy-nucleus repository may be a good fit.  The policy collection community should be separate from that and we are really looking to move our existing repository and fork it back to the current location.

qiujian16:
Is the plan to transfer the repo or create a new repo and copy the files? 

If it is a transfer, I think there are some openshift specific policies whic you may want to put in the openshift folder so user can understand what can be used with/without openshift. we have to update the readme to avoid mentioning product. We could have a sub folder indicating it is ACM specific policies and can only be used in ACM or openshift.

if it is latter option, since you would lose all PRs history anyway, I feel having it in the governance-policy-nucleus is fine? So that there are API definition in it, together with the real policies. 

I am also fine if we have a policy-collection dir in OCM repo. Any thought @mikeshng ?

Another question about https://github.com/stolostron/policy-collection/tree/main/policygenerator. Does it also rely on app addon?

gparvin:
> Is the plan to transfer the repo or create a new repo and copy the files?
> 

Transfer the repo is the plan.

> If it is a transfer, I think there are some openshift specific policies whic you may want to put in the openshift folder so user can understand what can be used with/without openshift. we have to update the readme to avoid mentioning product. We could have a sub folder indicating it is ACM specific policies and can only be used in ACM or openshift.
> 

Yes, we can update the readmes to make it not ACM specific but for ocm.

> if it is latter option, since you would lose all PRs history anyway, I feel having it in the governance-policy-nucleus is fine? So that there are API definition in it, together with the real policies.
> 
> I am also fine if we have a policy-collection dir in OCM repo. Any thought @mikeshng ?
> 
> Another question about https://github.com/stolostron/policy-collection/tree/main/policygenerator. Does it also rely on app addon?

The policy generator works with just the kustomize command.  It is also integrated with the app addon, but that is not a requirement for taking advantage of the policy generator.
Thanks!

gparvin:
@qiujian16 Are you ok with the transfer of policy-collection as long as we make updates for open-cluster-management and not keep it ACM specific?

qiujian16:
yes, I think we need to clarify in the doc it is a transfer rather than create. I would also suggest we update the repo before the transfer, e.g. reorganize the directory, update readme.

I am a bit concerned about no validation tests. I think a minimum API validation test is necessary per PR, otherwise we do not even known if a the API format is correct.

gparvin:
@qiujian16 @mprahl I have updated the PR to add the suggestions that were made.  Take another look when you get a chance and thanks for the reviews!

enhancements/sig-policy/78-policy-collection/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Goals
+
+- Transfer the policy collection community that provides samples, best practices and encourages quicker
+  adoption of the policy framework.
+- Make sure the transferred community is easy to consume by first time users of ocm policy.
+
+### Non-Goals
+
+- Beecause policies can be intended to work in many different environments with prerequisites that
```
qiujian16, line 34:
```suggestion
- Because policies can be intended to work in many different environments with prerequisites that
```

enhancements/sig-policy/78-policy-collection/README.md
```diff
@@ -0,0 +1,77 @@
...
+  may not be available, validation of policies is focused on syntax initially and additional validation
+  can be considered in the future.
+- At this time a procedure to provide support for donated policies by moving them to a stable
```
qiujian16, line 37: not quite clear about what is the criteria to address this.

enhancements/sig-policy/78-policy-collection/README.md
```diff
@@ -0,0 +1,77 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+The goal of this feature is to automate and simplify deployment of policies, policy sets and
+policy generator projects by creating (transferring) a community that provides best practices, samples, blogs
```
qiujian16, line 15: need some clarification on the meaning of community here. does it mean to be ocm community?

enhancements/sig-policy/78-policy-collection/README.md
gparvin: I am removing that item since I agree that it is confusing to provide as a non goal.

enhancements/sig-policy/78-policy-collection/README.md
gparvin: Fixed

enhancements/sig-policy/78-policy-collection/README.md
gparvin: I tried to clarify this.  Please review again.

qiujian16:
thanks, it looks fine to me

mprahl (approved):

Issue #81, opened 2022-11-18
Addon lifecycle enhancements


haoqing0110:
This is the issue to support addon rolling upgrade on API level.

PR #77, opened 2022-10-17
Addon install strategy

qiujian16:
Signed-off-by: Jian Qiu <jqiu@redhat.com>

https://github.com/open-cluster-management-io/enhancements/issues/76

PR #82, opened 2022-11-18
addon lifecycle enhancement

haoqing0110:
Ref: https://github.com/open-cluster-management-io/enhancements/issues/81
https://github.com/open-cluster-management-io/OCM/issues/114
Signed-off-by: haoqing0110 <qhao@redhat.com>


enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+
+### Goals
+
+- Enhance the existing `ClusterManagementAddon` and `ManagedClusterAddOn` API, to support version-based addon installation 
+on a group of clusters. 
+- Enhance the existing `ClusterManagementAddon` and `ManagedClusterAddOn` API, to reflect the addon installation, upgrade 
+and rollback progress, including the addon version, progress, and completed time.
+
+### Non-Goals
+How to enhance the existing addon framework and addon manager to support multiple version addon installtion is not this
```
deads2k, line 30: At a practical level in openshift, we found a need to tolerate multiple versions of various components at the same time.  Given a long rollout, potential canary failure, and then a roll-forward or roll-back, I suspect you'll face similar.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+```
+
+```golang
+// ManagedClusterAddOnSpec defines the install configuration of
+// an addon agent on managed cluster.
+type ManagedClusterAddOnSpec struct {
+  // …
+  InstallNamespace string `json:"installNamespace,omitempty"`
+	
+  // installVersion represents the desired addon version to install.
```
deads2k, line 80: I think I can tell from here, but which bit of code is responsible for determining what a version means and how to move from the current version to the desired version?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+  // Valid installVersion are listed in ClusterManagementAddOn status.supportedVersion.
```
deads2k, line 81: Since there is no transaction or foreign key constraints in a kube-apiserver, this is reasonably likely to contain invalid values.  How are invalid values represented?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+  // If installVersion is not specified, the ClusterManagementAddOn status.defaultVersion 
+  // wil be used.
+  // +optional
+  InstallVersion string `json:"installVersion,omitempty"`
```
deads2k, line 85: minor: I prefer desiredVersion

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+#### Beta
+1. Need to revisit the API shape before upgrading to beta based on user feedback.
+
+### Upgrade / Downgrade Strategy
+TBD
+
+### Version Skew Strategy
+N/A
+
+## Alternatives
```
deads2k, line 668: Can you compare this against an approach that uses `customizedVariables`?  I can appreciate version as a special thing, but the general idea of a rolling deployment is generic to any configuration change.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+spec:
+  installNamespace: open-cluster-management-agent-addon
+Status:
+…
+  currentVersion: v1
+  conditions:
+  - lastTransitionTime: "2022-11-02T05:48:12Z"
+    message: install completed with no errors.
+    reason: Succeed
+    status: "Flase"
```
deads2k, line 291:
```suggestion
    status: "False"
```

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+- When the addon in installing/upgrading/rollingback, the status will be true, 
+  reason will be Installing/Upgrading/Rollingback.
+- When the addon is installed successfully/failed, the status will be false, 
+  reason will be Succeed/Failed.
+
+```yaml
+  - lastTransitionTime: "2022-11-02T05:48:12Z"
+    message: install completed with no errors.
+    reason: Succeed
+    status: "Flase"
```
qiujian16, line 124: False

enhancements/sig-architecture/81-addon-lifecycle/README.md
qiujian16: this is a great point :)

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,669 @@
...
+  installVersion: v2
+Status:
+…
+  lastVersion: v1
+  currentVersion: v2
+  conditions:
+  - lastTransitionTime: "2022-11-02T05:48:12Z"
+    message: install completed with no errors.
+    reason: Succeed
+    status: "Flase"
```
zhiweiyin318, line 181: typo

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: Thanks for the comments, add more description to the desiredVersion. The upgrade/rollback should be triggerd if the current version listed in the status does not match the desired version.

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: Add more description to the desiredVersion. If invalid desiredVersion is specified, that may cause the upgrade or rollback to fail. And the result will be reported in the conditions.

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: Thanks for this comment, would provide an alternative solution.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,674 @@
...
+        configs:
+        - group: addon.open-cluster-management.io
+          resource: addondeploymentconfigs
+          name: canary-config
+status:
+  supportedVersions:
+  - v1
+  - v2
+  defaultVersion:
+  - v2
```
skeeey, line 167: it's a string, right?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,674 @@
...
+      addonTemplate:
+        configs:
+        - group: addon.open-cluster-management.io
+          resource: addondeploymentconfigs
+          name: global-config
+status:
+  supportedVersions:
+  - v1
+  defaultVersion:
+  - v1
```
skeeey, line 275: a string

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,674 @@
...
+        configs:
+        - group: addon.open-cluster-management.io
+          resource: addondeploymentconfigs
+          name: global-config
+status:
+  supportedVersions:
+  - v1
+  - v2
+  defaultVersion:
+  -v2
```
skeeey, line 338: one version

enhancements/sig-architecture/81-addon-lifecycle/metadata.yaml
```diff
@@ -0,0 +1,19 @@
...
+approvers:
+  - "@qiujian16"
+  - "@deads2k"
+  - "@elgnay"
+creation-date: 2022-11-18
+last-updated: 2022-11-18
+status: provisional
+see-also:
+  - "/enhancements/sig-architecture/58-addon-configuration"
+  - "/enhancements/sig-architecture/76-addon-install-strategy"
```
skeeey, line 19: empty line

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: yes, fixed.

haoqing0110:
Hello @deads2k @qiujian16 @skeeey I append an [alternatives](https://github.com/open-cluster-management-io/enhancements/blob/3819ef0d47193454efd1bedb25f1ea52457fe5d5/enhancements/sig-architecture/81-addon-lifecycle/README.md#alternatives) ,to treat the version as a variable of config. There are some gaps and concerns listed in the doc, want to know your ideas on that, and any other better suggestion on that?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,867 @@
...
+The design is based on [Addon install strategy #77](https://github.com/open-cluster-management-io/enhancements/pull/77/files). 
+
+### API changes
+
+```golang
+// ClusterManagementAddOnStatus represents the current status of cluster management add-on.
+type ClusterManagementAddOnStatus struct {
+  // SupportedVersions lists all the valid addon versions. 
+  // +optional
+  SupportedVersions []string `json:"supportedVersions,omitempty"`
```
deads2k, line 65: This is going to be tricky.  I suggest using this as a hint, possibly to create warnings, but at a practical level I think that enumerating all possible versions is difficult because different clusters may support different levels.

An example from openshift, we may recognize 4.10.8 and 4.10.9.  But 4.10.8 is bugged on GCP and 4.10.9 is bugged on vsphere.  If a given hub has spokes on GCP and on vsphere, the individual spokes will reject an invalid level and refuse to move to it..

Having the list of supported is good for a hint, but I suggest not enforcing based on it.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,867 @@
...
+          resource: addondeploymentconfigs
+          name: global-config
+    - name: canary-placement
+      namespace: default
+      addonTemplate:
+        desiredVersion: v2
+        configs:
+        - group: addon.open-cluster-management.io
+          resource: addondeploymentconfigs
+          name: canary-config
```
deads2k, line 465: I'm going to experiment with kustomize, but here's where my head is at.  If we consider that configuration changes are going to be treated the same and that often we even want to choose both, I'm wondering if using a gitops flow and having desiredVersion on the addondeploymentconfigs type and then using the "add-sha-suffix" option in kustomize and then substituting that in this name.  

That might conveniently solve version updates (with arbitrary skew), configuration changes, gitops flows, and gitops reverts, and manual management.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,867 @@
...
+      addonTemplate:
+        desiredVersion: v1
+        configs:
+        - group: addon.open-cluster-management.io
+          resource: addondeploymentconfigs
+          name: global-config
+    - name: canary-placement
+      namespace: default
+      addonTemplate:
+        desiredVersion: v2
```
deads2k, line 158: As I've been playing with this to see how the CLI flow would work, I hit a snag.  What happens in the case where a managedcluster matches both global-placement and canary-placement?  The problem exists for anything setting the addonTemplate.

Also, as I've been messing with it, I've wondered what the field experience with addonDeploymentConfigs have been.  Some configuration options are definitely per-managed-cluster configuration (the enhancement explained those well), but in the field do we also see a class of "these values should be set for this addon on many clusters" and if that class exists, would we embed it here or reference it here or something else?

I'm going to push a branch that extends this PR with the kustomize example I was working with.  I don't think it's the only way, but it exposed these as issues and could drive goals, non-goals, and alternatives.

deads2k: > I'm going to push a branch that extends this PR with the kustomize example I was working with. I don't think it's the only way, but it exposed these as issues and could drive goals, non-goals, and alternatives.

I opened https://github.com/open-cluster-management-io/enhancements/pull/84 and dropped a few comment and READMEs in there to seed discussion about a possible alternative.

qiujian16: we do have received many feedbacks on addonDeploymentConfigs api yet. I think currently it is mostly use as "these values should be set for this addon on many clusters". But not per cluster configuration.

haoqing0110:
Sorry for taking so long to refresh the proposal, the new proposal is trying to combine rollout strategy and config based on https://github.com/open-cluster-management-io/enhancements/pull/84 . 

It indeed takes some effort to consider, simplify and cover all the cases, now the new API is relatively clear. 

@deads2k @qiujian16 @skeeey Looking forward to your comments and feedback. 
/cc @zhiweiyin318 @zhujian7

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -37,26 +38,787 @@ proposal's scope.
...
+
+  // Rolling update with placement config params. Present only if the type is RollingUpdateWithPlacement.
+  // +optional
+  RollingUpdateWithPlacement *RollingUpdateWithPlacement `json:"rollingUpdateWithPlacement,omitempty"`
+}
+
+// RollingUpdateWithPlacement represents the placement and behavior to rolling update add-on configurations
+// on the selected clusters.
+type RollingUpdateWithPlacement struct {
+  // name of the placement
```
deads2k, line 141: How will this intersect with install strategy?  The PR in https://github.com/open-cluster-management-io/enhancements/pull/77 looked pretty good to me and allowed targeting multiple placements in a way that seemed useful.

Having an overall set of placements for the `ClusterManagementAddOn` and a way to specify a different placement (list of placements?) for a canary would match your description.  Is that the intent of this placement?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -37,26 +38,787 @@ proposal's scope.
...
+`ManagedClusterAddOnStatus` adds `SupportedConfigs` to list the configuration types which are allowed to override the 
+the configuration defined in `ClusterManagementAddOn`.
+
+`ManagedClusterAddOnStatus` remove `LastObservedGeneration` and use `DesiredConfigSpecHash` and `LastAppliedConfigSpecHash` 
+to record the configuration spec hash.
+
+```golang
+// ManagedClusterAddOnStatus provides information about the status of the operator.
+// +k8s:deepcopy-gen=true
+type ManagedClusterAddOnStatus struct {
```
deads2k, line 172: When performing a canary rollout, you want new clusters that are added to the placement to get the old configuration, correct?  If so, you need to track that old configuration somehow.  It could be done in status or it could be done in a different resource.  I suspect status here will be easier, but I don't know for certain.

I think the configuration is logically the combination of *all* the ConfigReferences, not the individual parts.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -37,26 +38,787 @@ proposal's scope.
...
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ClusterManagementAddOn
+metadata:
+  name: helloworld
+spec:
+  configuration:
+    configs:
+    - group: addon.open-cluster-management.io
+      resource: addondeploymentconfigs
+      name: deploy-config-xxx
```
deads2k, line 297: note to self, I think these could linked to the names of AddOnHubConfig in kustomize, but an example proving it is probably important.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -37,26 +38,787 @@ proposal's scope.
...
+    type: Placements
+    placements:
+    - name: aws-placement
+      namespace: default
+      addonSpec:
+        installNamespace: open-cluster-management-agent-addon
+        configs:
+          - group: addon.open-cluster-management.io
+            resource: addondeploymentconfigs
+            name: aws-config
```
deads2k, line 318: Is someone going to ask if we can have a canary for changing this as well?  Would we do it by having an update strategy per `.spec.installStrategy.placement`?  That may be practical.  Perhaps my rolloutStrategy suggestion should have been under `.spec.installStrategy.placement`?

Can we try to set up a synchronous meeting to discuss this nesting aspect?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -37,26 +38,787 @@ proposal's scope.
...
+    desiredConfigSpecHash: <aws-config-spec-hash>
+    lastAppliedConfigSpecHash: <aws-config-spec-hash>
+  - group: addon.open-cluster-management.io
+    resource: addonhubconfigs
+    name: hub-config-yyy
+    desiredConfigSpecHash: <hub-config-yyy-spec-hash>
+    lastAppliedConfigSpecHash: <hub-config-xxx-spec-hash>
+  conditions:
+  - lastTransitionTime: "2022-11-02T05:48:12Z"
+    message: upgrading...
```
deads2k, line 670: We should be able to indicate how many are at sha/X, sha/Y, and sha/Z, right?

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,1407 @@
...
+    configs:
+    - group: addon.open-cluster-management.io
+      resource: addondeploymentconfigs
+      name: deploy-config-xxx
+      namespace: open-cluster-management
+    - group: addon.open-cluster-management.io
+      resource: addonhubconfigs
+      name: hub-config-yyy
+    rolloutStrategy:
+      type: RollingUpdateWithPlacement
```
deads2k, line 303: 1. RollingUpdateWithCanary -  RollingUpdate PLUS you can choose which clusters update first.
2. RollingUpdate - this would let you choose maxConcurrentlyUpdating and would block until addons on cluster were healthy
3. UpdateAll - updates everything all at once.

I think the default being RollingUpdate is probably best.

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: Yes, a nested rollout strategy makes more sense, have updated the proposal with our latest ideas.

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: Considering adding `installProgression` to the `ClusterManagementAddOn` status, more details should be fulfilled there.

enhancements/sig-architecture/81-addon-lifecycle/README.md
haoqing0110: Some corner cases consideration is added to the proposal, including the newly added clusters.

deads2k (commented):
Minor questions.  The examples in this proposal are excellent.  When the docs are written, I think the focus should be on how to make good use of kustomize (or similar) with the `AddOnHubConfig` to achieve rollingupdate, canaryupdate, and rollback.

I know this has been long in process, but I think this API is making it possible (and fairly easy) for clients to successfully manage their addons.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,1259 @@
...
+	// - Placements: install to clusters selected by placements.
+	// +kubebuilder:validation:Enum=Manual;Placements
+	// +kubebuilder:default:=Manual
+	// +optional
+	Type string `json:"type"`
+
+	// ClusterLabelSelector is a label selector honored when install strategy type is
+	// ClusterLabelSelector
+	// +optional
+	ClusterLabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
```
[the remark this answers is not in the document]
haoqing0110: Yes, removed that.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,1259 @@
...
+
+`ClusterManagementAddOnStatus` adds `ConfigReferences` to represent the current configuration references and `DesiredConfigSpecHash`.
+
+```golang
+// ClusterManagementAddOnStatus represents the current status of the cluster management add-on.
+type ClusterManagementAddOnStatus struct {
+	// configReferences is a list of current add-on configuration references.
+	// The configs defined in ClusterManagementAddOn defaultConfigs and installStrategy configs fields will be listed here.
+	// +optional
+	ConfigReferences []DesiredConfigReference `json:"configReferences,omitempty"`
```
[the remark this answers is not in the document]
haoqing0110: Yes, modified the `InstallProgression` to show the progression per placement. Inside it has a field named `LastKnownGoodConfigSpecHash` to representing the last levels that have passed the canary placement.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,1259 @@
...
+A new annotation `configsSpecHash` is added to the `ManifestWork` to record the corresponding config and spec hash.
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata:
+  labels:
+    open-cluster-management.io/addon-name: application-manager
+  annotations:
+    configsSpecHash: {"<resource>.<group>/<namespace>/<name>":"<config-spec-hash>",...}
```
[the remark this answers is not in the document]
haoqing0110: Needs more discussion with @qiujian16 for this.

enhancements/sig-architecture/81-addon-lifecycle/README.md
```diff
@@ -0,0 +1,1259 @@
...
+- if there's already `ManagedClusterAddOn` on that cluster, it should wait for the canary upgrade finish, then follow the 
+aws rollout strategy to upgrade the addon to the desired status.
+
+If the new added cluster belongs to both of them, it should first follow the strategy in canary-placement, and then follow 
+the strategy in aws-placement.
+
+2. Should the changes in canary frequently trigger the aws rollout?
+
+The case is configs or clusters change after the canary upgrade is complete and before the aws upgrade finish. Should this 
+stops the aws rollout and wait for the canary to complete again and retrigger the rollout?
```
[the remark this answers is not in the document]
haoqing0110: Introduce last-known-good is a good suggestion, have updated the proposal with this idea.

deads2k:
I like the idea and it still lgtm.  I think it composed pretty well with https://github.com/open-cluster-management-io/enhancements/pull/82 .

It will be interesting to learn how many placements are added in practice, whether it is just one or multiple.  That might inform the idea of allowing defining the "install here" separately in the future. But given how well it composed, this looks like a good place to start.

enhancements/sig-architecture/81-addon-lifecycle/README.md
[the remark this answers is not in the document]
qiujian16: @deads2k yes it represents the concept of "contributingObjectHashes". Will we have such general concept in k8s?

haoqing0110:
Try to add kustomize examples in the latest [commits](https://github.com/open-cluster-management-io/enhancements/pull/82/commits/e56405f53b6f553a3587f47178e298bd6b1ab8be) . 

@deads2k see you used `HashTransformer` in https://github.com/open-cluster-management-io/enhancements/pull/84 , I try to search it, but don't find an example of it in kustomize website. I'm wondering what's it used for? Can it be used in this proposal?

haoqing0110:
@qiujian16 plz help tag.

PR #85, opened 2023-01-27
Add the Gatekeeper integration for OCM Policy design

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-2746

### DIFF
--- a/enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
+++ b/enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
@@ -1,0 +1,268 @@
+# Native Support for Gatekeeper Constraints in Policies
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Enhance the Open Cluster Management (OCM) policy framework to allow Gatekeeper `ConstraintTemplate` and constraint
+manifests to be directly embedded in OCM `Policy` object's `policy-templates` array for multicluster distribution and
+status aggregation.
+
+## Motivation
+
+The Open Cluster Management (OCM) Policy addon does not have native support for an admission controller. Gatekeeper is
+currently favored by the OCM Policy SIG for its flexibility and performance. Though the installation of Gatekeeper will
+not be simplified by this enhancement, there is a desire to provide native OCM Policy support for Gatekeeper
+`ConstraintTemplate` and constraint objects. This will allow users to skip using `ConfigurationPolicy` objects that wrap
+Gatekeeper constraint manifests.
+
+This also allows the audit messages from Gatekeeper constraints to be aggregated in the policy status on the OCM hub,
+which is not possible by the wrapping approach. The wrapping approach involves an inform mode `ConfigurationPolicy` that
+checks that the Gatekeeper constraint's `status.totalViolations` field is set to `0`. If it's not, a violation message
+of `violation - k8srequiredlabels not found: [ns-must-have-gk] found but not as specified` is returned. This vague
+message does not indicate what the constraint violations are, which is not comparable to the status messages of native
+OCM Policy addon controllers.
+
+### Goals
+
+1. Gatekeeper `ConstraintTemplate` and constraint manifests can be embedded in OCM `Policy` object's `policy-templates`
+   array.
+2. Status of Gatekeeper constraints are aggregated in the OCM `Policy` objects on the OCM hub.
+
+### Non-Goals
+
+1. The installation of Gatekeeper will not be simplified in this enhancement.
+2. When Gatekeeper denies a request, the OCM `Policy` status will be unaffected since the constraint is working as
+   expected.
+
+## Proposal
+
+### Design
+
+#### Policy Syntax
+
+This proposes adding native support to the OCM Policy addon for Gatekeeper constraints. As shown in the example below, a
+user would be able to directly embed Gatekeeper `ConstraintTemplate` (v1beta1 and v1 should both be supported) and
+constraint manifests in the OCM `Policy` object's `policy-templates` array. Note that the `ConstraintTemplate` is not
+required to be in the OCM `Policy`. It can be deployed in another way such as with a `ConfigurationPolicy` or with
+ArgoCD.
+
+```yaml
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: gk-policy
+spec:
+  remediationAction: inform
+  disabled: false
+  policy-templates:
+    - objectDefinition:
+        apiVersion: templates.gatekeeper.sh/v1beta1
+        kind: ConstraintTemplate
+        metadata:
+          name: k8srequiredlabels
+        spec:
+          crd:
+            spec:
+              names:
+                kind: K8sRequiredLabels
+              validation:
+                openAPIV3Schema:
+                  properties:
+                    labels:
+                      type: array
+                      items: string
+          targets:
+            - target: admission.k8s.gatekeeper.sh
+              rego: |
+                package k8srequiredlabels
+
+                violation[{"msg": msg, "details": {"missing_labels": missing}}] {
+                  provided := {label | input.review.object.metadata.labels[label]}
+                  required := {label | label := input.parameters.labels[_]}
+                  missing := required - provided
+                  count(missing) > 0
+                  msg := sprintf("you must provide labels: %v", [missing])
+                }
+    - objectDefinition:
+        apiVersion: constraints.gatekeeper.sh/v1beta1
+        kind: K8sRequiredLabels
+        metadata:
+          name: ns-must-have-gk
+        spec:
+          enforcementAction: dryrun
+          match:
+            kinds:
+              - apiGroups: [""]
+                kinds: ["Namespace"]
+          parameters:
+            labels: ["gatekeeper"]
+```
+
+When the `Policy` has the `spec.remediationAction` field set to `inform`, it would override the constraint's
+`spec.enforcementAction` with `warn`. This will not deny requests but instead, it will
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+The Template Sync controller would not put owner references on Gatekeeper `ConstraintTemplate`/constraints, but instead
+add a finalizer of `policy.open-cluster-management.io/cleanup-cluster-scoped-policies` to the `Policy` object that
+defines the Gatekeeper `ConstraintTemplate`/constraints. This finalizer would indicate that Gatekeeper
+`ConstraintTemplate`/constraints need to be removed when the OCM `Policy` object is deleted and if applicable, when the
+`Policy` object is updated. To be able to track which Gatekeeper constraints belong to a `Policy` object, a new label
+should be added to the Gatekeeper `ConstraintTemplate` objects and constraints of
+`policy.open-cluster-management.io/policy` with the policy name. Note that a finalizer will also need to be added to the
+`governance-policy-framework` `Deployment` like the Configuration Policy controller does to prevent the Template Sync
+from being deleted before the finalizer can be handled.
```
JustinKuli, line 187: I think we should make sure this works similarly to our fix for https://issues.redhat.com/browse/ACM-3049 (which I think @willkutler is thinking about).

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,257 @@
...
+added. This is a bit tricky because each Gatekeeper `ConstraintTemplate` produces a new custom resource definition (CRD)
+and thus the watches aren't known at controller instantiation time.
+
+To handle this, the controller would watch all `Policy` objects in the managed cluster namespace. During the reconcile,
+the controller will detect if Gatekeeper constraints are managed by the policy. If so, the
+[dynamic watch](https://github.com/stolostron/kubernetes-dependency-watches) library will be used to watch the
+individual Gatekeeper constraints and trigger reconciles on the `Policy` parent object whenever they have status
+updates.
+
+The new controller would also examine the `Policy` status before sending a status event to avoid sending duplicate
```
JustinKuli, line 218: Do other policy controllers do this? Should they?

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,257 @@
...
+status events. It's possible that duplicate status events could be sent if the Status Sync controller is overwhelmed and
+the Gatekeeper constraint has multiple reconciles in succession. This is considered acceptable because the Status Sync
+controller filters out duplicate status events.
+
+### Risks and Mitigation
+
+### Open Questions [optional]
```
JustinKuli, line 225: ConstraintTemplates and Constraints are `v1beta1` in Gatekeeper, I wonder if we need to carefully consider how to handle if/when they change version?

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,257 @@
...
+    - enforcementAction: dryrun
+      kind: Namespace
+      message: 'you must provide labels: {"gatekeeper"}'
+      name: gatekeeper-system
+```
+
+The status message for the constraint in the OCM `Policy` would be in the format of
+`<message> (<enforcementAction> of <kind> <namespace/name or name>)` for each violation entry each separated by a
+semicolon. For example, the above example would produce a status message of
+`you must provide labels: {"gatekeeper"} (dryrun of Namespace default); you must provide labels: {"gatekeeper"} (dryrun of Namespace gatekeeper-system)`.
```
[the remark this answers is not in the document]
mprahl: Yes, that's better. I'll change it.

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
mprahl: Other controllers don't currently and this is likely to avoid coupling policy controllers to how status is stored in the parent policy, however, I'm not opposed to this if there is a common library (e.g. policy-nucleus).

For this particular case, I think it makes sense because the policy enforcer (Gatekeeper) is not the one producing the policy framework status event, so there needs to be a way to know what status events have been relayed from Gatekeeper.

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
mprahl: Good point. `ConstraintTemplate` seems to be already `v1` in newer Gatekeeper versions. I'll add a note to support both `v1` and `v1beta1`.

Constraints on the other hand, I think we will just need to cross that bridge when it does get to v1 since it's still v1beta1:
https://github.com/open-policy-agent/frameworks/blob/5e694bda2ffa6f35fa6b82af26d3c47b043b7040/constraint/pkg/client/crds/crds.go#L46

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,258 @@
...
+status aggregation.
+
+## Motivation
+
+The Open Cluster Management (OCM) Policy addon does not have native support for an admission controller. Gatekeeper is
+currently favored by the OCM Policy SIG for its flexibility and performance. Though the installation of Gatekeeper will
+not be simplified by this enhancement, there is a desire to provide native OCM Policy support for Gatekeeper
+`ConstraintTemplate` and constraint objects. This will allow users to skip using `ConfigurationPolicy` objects that wrap
+Gatekeeper constraint manifests. This also allows the audit messages from Gatekeeper constraints to be aggregated in the
+policy status on the OCM hub, which is not possible by the wrapping approach.
```
JustinKuli, line 25: If I was managing ConstraintTemplates and constraints with ConfigurationPolicies (without this enhancement), what do the violations look like?

Based on https://github.com/open-cluster-management-io/policy-collection/blob/main/community/CM-Configuration-Management/policy-gatekeeper-container-image-latest.yaml (just the first constraint example I found), it looks like to find violations without this enhancement, it checks the constraint's status.totalViolations count. Which means that the only message would be something generic like `violation - ____ found but not as expected`.

I think the improved messages possible from this enhancement are a big advantage.

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+If the `remediationAction` field is set to `enforce`, the constraint's `spec.enforcementAction` would be overriden with
+`deny`. Existing violations would show up in the constraint's `status.violations` section.
+
+If the `remediationAction` field is not set, the constraint's `spec.enforcementAction` would not be modified.
+
+#### Status Reporting
+
+After applying a Gatekeeper constraint, Kubernetes objects on the cluster that violate the constraint will show up in
+the constraint's `status.violations` field as shown in the example below.
```
JustinKuli, line 117: Will there be a special message if there are no violations?

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
```diff
@@ -0,0 +1,258 @@
...
+`ConstraintTemplate`/constraint will be deleted to match the behavior of other OCM policy templates.
+
+### User Stories
+
+#### Story 1
+
+As a policy user, I would like to have native multicluster support for an admission controller so that I can have a
+consistent distribution and status aggregation mechanism for all my policies.
+
+### Implementation Details/Notes/Constraints [optional]
```
JustinKuli, line 166: If I understand correctly, template-sync will create the ConstraintTemplates and constraints (so, it still creates all the templates in a Policy). And the new controller will just be responsible for watching those and emitting compliance events. The status-sync is still the only thing that puts those events into the status of the policy.

If that's right, then is it necessary for the new controller be included in the framework container/pod? It feels more similar to the other policy controllers, which watch their policy types and emit events. I'm imagining that someone might want to turn this controller off, in which case it being a separate pod might make that easier.

Granted, this new controller (as described here) operates a little differently than other policy controllers: it needs to watch Policies to know what types the constraints will be, and it uses the new `policy.open-cluster-management.io/policy` label instead of ownerreferences to track some things. But I'm not sure if that necessarily means it must be included in the framework container.

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
mprahl: That's right. The error would be something like `violation - k8srequiredlabels not found: [ns-must-have-gk] found but not as specified`. I'll add that to the doc.

enhancements/sig-policy/85-gatekeeper-policy-integration/README.md
mprahl: It doesn't have to be in the Policy Framework. I was doing this for convenience and also because it's essentially just relaying status updates, so it seemed like a companion to the `policy-status-sync` controller.

With the current design, the resource utilization should be very minimal if the Gatekeeper integration is not used since all it will do is loop through the `policy-templates` array of all the policies looking for Gatekeeper constraints, but the cache will be shared with the existing controllers in the Policy Framework, so another watch won't be necessary.

I think the low impact is worth making the installation experience easier, but I'm open to suggestions on this one.

JustinKuli: I'm ok with it either way, I wanted to check if there was a need to keep it in the framework.

I have a slight preference for it to be a separate controller. That follows the pattern of the other policy controllers, and would make it easier to debug and potentially disable. I think that when it is being used, it will have a non-zero impact on the performance of the rest of the framework, and in my opinion the rest of the framework is more essential and should be "protected"

But since it does need to watch all the Policies, a shared cache is better there...

gparvin: Note that we don't want to install any additional standalone controllers on managed clusters in any "default" installation scenarios.  We need to be very mindful of any impacts to the footprint we have on the managed cluster.  That's one reason I'm for consolidating cert and iam into a security controller and maybe this would fit better into that category.  Since iam and cert are lightly used, adding admissions control would be a third "low impact" security controller that could be combined.

gparvin:
/hold
For multiple approvals. 
/lgtm

JustinKuli (approved):
LGTM

PR #87, opened 2023-02-14
WIP: addon template

zhujian7:
Signed-off-by: zhujian <jiazhu@redhat.com>


enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,344 @@
...
+
+- Provide a new addon development model, developers can define an addon YAML template and some parameters to describe
+  how the addon agent will be installed on the managed cluster and how to register it to the hub, developers do not need
+  to deploy an addon-manager on the hub cluster to manage the addon agent.
+- Make the addon agent deployment and registration configurable, modifying the deployment and registration logic does
+  not need to recompile the addon image.
+
+### Non-Goals
+
+- The addon agent image is still required.
```
qiujian16, line 37: For complicated addon which has special rbac requirement etc would still need to use addon-framework

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,406 @@
...
+// Will create a RoleBinding in the same namespace as the managedClusterAddon to bind the user
+// provided ClusterRole/Role to the "system:open-cluster-management:cluster:<cluster-name>:addon:<addon-name>"
+// Group.
+type HubPermissionConfig struct {
+	// Type of the permissions setting. It defines how to bind the roleRef
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=CurrentCluster;SingleNamespace
+	Type HubPermissionsBindingType `json:"type"`
+
+	// RoleRef is an reference to the permission resource. it could be a role or a cluster role,
```
deads2k, line 168: if it's a role, which namespace must it be in and what will create it?

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,406 @@
...
+	SigningCA SigningCARef `json:"signingCA"`
+}
+
+// SigningCARef is the reference to the signing CA secret that must contain the certificate authority data with key
+// "ca.crt"
+type SigningCARef struct {
+	// Namespace of the signing CA secret
+	// +kubebuilder:validation:Required
+	Namespace string `json:"namespace"`
+	// Name of the signing CA secret
```
deads2k, line 209: why a secret and not a configmap?

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,406 @@
...
+	// signerName is the name of signer that addon agent will use to create csr.
+	// +required
+	// +kubebuilder:validation:MaxLength=571
+	// +kubebuilder:validation:MinLength=5
+	SignerName string `json:"signerName"`
+
+	// Subject is the user subject of the addon agent to be registered to the hub.
+	// If it is not set, the addon agent will have the default subject
+	// "subject": {
+	//	"user": "system:open-cluster-management:addon:{addonName}:{clusterName}:{agentName}",
```
deads2k, line 193: What is the substitution policy for getting things like `clusterName` injected?

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: AddOnTemplate
+metadata:
+  name: hello-template
+spec:
+  addonName: hello-template
+  agentManifests:
```
deads2k, line 7: Is the primary benefit of these `agentManifests` to provide a way to have the spokes start running a pod without the add-on  having to provide a hub-side controller to create manifest-works or other distribution method?

deads2k, line 7: Is the primary benefit of these `agentManifests` to provide a way to have the spokes start running a pod without the add-on  having to provide a hub-side controller to create manifest-works or other distribution method?

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
```diff
@@ -0,0 +1,87 @@
...
+          metadata:
+            labels:
+              app: hello-template-agent
+          spec:
+            serviceAccountName: hello-template-agent-sa
+            containers:
+              - name: helloworld-agent
+                image: quay.io/zhujian/addon-examples:latest
+                imagePullPolicy: Always
+                args:
```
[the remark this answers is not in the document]
deads2k: IOW, instead of allowing free-form anything, what if we constrain exactly what we want/need. I can see the need/desire for RBAC on the managedcluster as well.

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
```diff
@@ -0,0 +1,87 @@
...
+        - kind: ServiceAccount
+          name: hello-template-agent-sa
+          namespace: "{{INSTALL_NAMESPACE}}"
+  registration:
+    # kubeClient or custom signer, if kubeClient, user and group is in a certain format.
+    # user is system:open-cluster-management:cluster:...
+    # group is [system:open-cluster-management:cluster:, ...]
+    - type: KubeClient
+      kubeClient:
+        permissions:
```
deads2k, line 61: These are hubPermissions, right?  Perhaps a more precise name?

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
qiujian16: it is still to use manifestwork to distribute, but it is to avoid each add-to providing a hub-side controller.

enhancements/sig-architecture/82-addon-template/README.md
zhujian7: if the type is CurrentCluster, the role should be in the namespace with the same name as the managed cluster.
if the type is SingleNamespace, the role should be in the namespace configured by the field "SingleNamespace"
Our addon manager will not create it, so the addon developer/manager will be responsible for creating it.

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,406 @@
...
+
+	// SigningCA represents the reference of the secret to sign the CSR
+	// +kubebuilder:validation:Required
+	SigningCA SigningCARef `json:"signingCA"`
+}
+
+// SigningCARef is the reference to the signing CA secret that must contain the certificate authority data with key
+// "ca.crt"
+type SigningCARef struct {
+	// Namespace of the signing CA secret
```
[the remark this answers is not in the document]
zhujian7: on the hub

enhancements/sig-architecture/82-addon-template/README.md
zhujian7: in addition to "ca.crt", the "ca.key" is also required, so using secret.
I have added that "ca.key" is also required on the field comment.

qiujian16: this secret is for this manager on the hub to sign the csr with the custom signer for the agent on the spoke

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
qiujian16: the benefit is to avoid add-on providing a hub-side controller to create manifest-works. The manifestworks is creates by a generic controller based on the manifests defined in this API.

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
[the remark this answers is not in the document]
qiujian16: in most cases, the manifests of addon are deployment/sa/RBAC. But for some addons, it also include CRDs and related CRs.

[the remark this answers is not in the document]
zhujian7: the addonTemplate is for developers. Developers may want to define addon agents in a more Kubernetes native way, which is in line with the usage habits of most developers, and we don’t know whether other than deployment/sa/RBAC/CRDs/CRs additional resources are needed

zhujian7:
Hi @deads2k Thank you for reviewing this, the comments have been addressed, could you help to take another look?

enhancements/sig-architecture/82-addon-template/examples/addon-template.yaml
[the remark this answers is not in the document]
deads2k: Those examples help, thanks.

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,407 @@
...
+type AddOnTemplateSpec struct {
+	// AddonName represents the name of the addon which the template belongs to
+	// +kubebuilder:validation:Required
+	// +required
+	AddonName string `json:"addonName"`
+
+	// AgentManifests represents the kubernetes resources of the addon agent to be deployed on a managed cluster.
+	// +kubebuilder:validation:Required
+	// +required
+	AgentManifests []Manifest `json:"agentManifests"`
```
deads2k, line 81: I would nest this in case you end up needing to add more out of ManifestWorkSpec .  Perhaps `template` similar to how we nest a PodTemplateSpec for workload resources.  You don't have to pre-emptively include the entire resource, but I would structure this to allow for expansion to include it.

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,407 @@
...
+type HubPermissionConfig struct {
+	// Type of the permissions setting. It defines how to bind the roleRef
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum:=CurrentCluster;SingleNamespace
+	Type HubPermissionsBindingType `json:"type"`
+
+	// RoleRef is an reference to the permission resource. it could be a role or a cluster role,
+	// the user must make sure it exist on the hub cluster.
+	// +kubebuilder:validation:Required
+	RoleRef rbacv1.RoleRef `json:"roleRef"`
```
deads2k, line 171: A cluster role ref may be convenient here.  We would expect this to be the same in every namespace, rigth?

enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -0,0 +1,407 @@
...
+requirement for the addon agent, it is expected to periodically refresh the lease by implementing
+[lease.LeaseUpdater](https://github.com/open-cluster-management-io/addon-framework/blob/9466c062199e66b9e96e6393f5a4c1ac9cf84cd6/pkg/lease/lease_controller.go#L23-L31)
+interface.
+
+But in order to get rid of the requirement for developers, we are considering another type of healthiness probe:
+injecting a feedback rule to check if the agent deployment resource is available(the output of
+`kubectl get deployment -n <addon-agent-namespace> <addon-agent-name> -ojsonpath="{.status.conditions[?(@.type=='Available')].status}"`
+is `True`).
+
+### Upgrade an addon template
```
deads2k, line 323: very cool use-case.

enhancements/sig-architecture/82-addon-template/README.md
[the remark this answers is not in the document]
deads2k: > on the hub
should clarify in the doc.   Probably also need to explain how the CSR is going to be approved.  I think we can know because the requestor will match the addon identity, but that's not spelled out.

enhancements/sig-architecture/82-addon-template/examples/upgrade/cluster-management-addon.yaml
```diff
@@ -0,0 +1,31 @@
...
+        namespace: cluster1
+  installStrategy:
+    type: Placements
+    placements:
+      - name: placement-all
+        namespace: default
+        configs:
+          - group: addon.open-cluster-management.io
+            resource: addontemplates
+            name: hello-template-v2
```
deads2k, line 27: neat to see this tied together.

deads2k:
> Hi @deads2k Thank you for reviewing this, the comments have been addressed, could you help to take another look?

A couple minor comments.  The examples here are great.

deads2k:
lgtm

PR #88, opened 2023-03-01
Placement Strategy enhancement

serngawy:


enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+ 
+	// PlacementStrategy define canary clusters as subset of the selected clusters and divide the selected clusters to groups.
+	// +optional
+	PlacementStrategy PlacementStrategy `json:"placementStrategy,omitempty"`
+}
+
+type PlacementStrategy struct {
+	// Label selector for canary clusters as subset of the selected clusters from predicates and ClusterSets.
+	// +optional
+	CanaryClusters metav1.LabelSelector `json:"canaryClusters,omitempty"`
```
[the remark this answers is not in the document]
serngawy: Not sure if  the name is confusing, could be something like **FirstGroupSelector** ? but the idea is those selected clusters need to be in the first cluster group.  
"Optionally, we we use another placement and define that placement is a affinity" I have 2 thoughts on that;
1- UX; as a user why do I need to define 2 placements ? usually canary is a small subset of the large cluster set that I want to select. If as user will define 2 placements as example; **1-** first placement should **prod** and **Canary**. **2-** second placement should **prod** and **Not Canary**. I feel its more pushing effort to end user while we can easily handle that. 
2- What if the affinity placement has affinity placement and so on ? it will add complexity to select all clusters

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+Adding NumberOfClusterGroups to the PlacementStatus API 
+```go
+type PlacementStatus struct {
+	// NumberOfClusterGroups determined by placement strategy. If there is no placement strategy defined all selected clusters will be considered in 1 group
+  // +kubebuilder:default:=1
+	// +optional
+	NumberOfClusterGroups int32 `json:"numberOfClusterGroups"`
```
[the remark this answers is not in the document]
qiujian16: How about we create 3 placements, each as a group. And we can set predicate and percentage on each placement. So each placement can select a certain percentage of clusters with the same predicate. We then could set an anti-affinity across placement to ensure each selection of placement is exclusive.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
jc-rh: For the canary use case, the cluster set is not only a subset of the larger set, but also excluded from the rest of the larger set in terms of grouping. Placement affinity probably does the sub selection part, but doesn't provide grouping isolation?

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
jc-rh: Just to clarify, a better way of saying what I meant is "no shuffling between groups due to cluster changes (labels, new cluster, deleted cluster etc)". When the placement spec changes (predicates, strategy), we probably don't want to promise any kind of stability. Users have to expect all groups to be reformed.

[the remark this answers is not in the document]
jc-rh: @qiujian16 By "as a group", do you mean the placements can share the same predicates or we have to duplicate them? My understanding of the intent of this enhancement is to capture the essence of the grouping concepts and knobs in one place rather than spread across many CRs. I suppose variable size grouping could be useful in some cases. If it's part of the goal, we can have a percentage array in the strategy to realize that.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+	CanaryClusters metav1.LabelSelector `json:"canaryClusters,omitempty"`
+
+	// ClustersGroupPercentage is a percentage of the total selected clusters that will divide the clusters to groups.
+	// ex; for a total 100 clusters selected, ClustersGroupPercentage equal to 20% will divide the selected clusters to
+	//     5 groups each group should have 20 clusters.
+	// +kubebuilder:clustersGroupPercentage:Required
+	// +kubebuilder:clustersGroupPercentage:Enum=5%;10%;15%;20%;25%;30%;50%;100%
+	// +kubebuilder:default:=100%
+	// +optional
+	ClustersGroupPercentage string `json:"clustersGroupPercentage,omitempty"`
```
jc-rh, line 66: How about MaxPercentage? I think `Max` describes the use of this field more accurately. The canary group and the last group will likely have less clusters. Same applies to the number of clusters per group if we plan to support that.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
serngawy: How about name it affinityClusterGroup? It gives the meaning of dependent cluster group

serngawy:
adding @jnpacker

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
qiujian16: I think we can have a field in the placement defining that this placement is affinity to another placement, so you do not need to duplicate the predicate

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
qiujian16: I think it looks like a `decisionGroup` which is just to sub-divide the decisions to multiple groups. But I am still not fully convinced that the pros of having a `decisionGroup` than using another placement to generate another set of decision.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+  decisions:
+    - clusterName: cls155
+      reason: ''
+    ...
+```
+
+The Placement Helper [library](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1beta1/helpers.go) should provide functions to facilitate retrieve the cluster groups.
+
+
+### How Workload applier APIs will benefits from using placementStrategy
```
qiujian16, line 187: I think we should have some examples.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Upgrade / Downgrade Strategy
+
+Placement->placementStrategy API will be supported on the target release of OCM (open cluster management) and it should not affect the OCM current APIs functionalities or back compatibility.
+
+## Alternatives
+
+1. End user can create many placements and apply the workloads using each placement one by one to have a progressive roll out. 
+1. Each OCM APIs such as ManifestWorkReplicaSet, ClusterManagementAddon and Policy define its own roll out strategy to determine the progressive roll out.
+1. Current alternative does not satisfy the requirements as explained above in the motivation section
```
qiujian16, line 220: with multiple placements, we can not define the number of clusters or percentage in each group, and dynamically generate each groups

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
zhiweiyin318: the canary update which I understand is similar to [OCP worker nodes rollout canary update](https://docs.openshift.com/container-platform/4.12/updating/update-using-custom-machine-config-pools.html). 
End user selects a set of clusters as canary cluster to update firstly, and evaluates the workloads health after the update is finished, and then to determine whether to update another selected clusters. In most cases the canary clusters are selected rather than a group of random clusters.
During the updating of the canary clusters, end user may not expect they are upgraded at the same time if the count of selected cluster is large, and wants they are upgrade under a maximum concurrently updating number. I prefer to call this update as rollout. In rollout update, the group of clusters can be randomly selected.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management/website/)
+
+## Summary
+
+The proposed work would define a new field in the Placement API (placementStrategy) that will let the end user able to break down the selected managed clusters to groups and define canary clusters.
```
zhiweiyin318, line 13: The placement APIs is to provide a definition on how to select a set of ManagedClusters in multiple ManagedClusterSets. The end user can use it in multiple scenarios including upgrade. canary clusters sounds like only used in upgrade scenarios, right?  not sure if it is proper to define a specified property in a common API spec.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
serngawy: yes agree, the canary has been removed and affinity cluster selection is added. the affinity cluster selection has its own label selector.

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: add example for manifestWorkReplicaSet use of the proposed placement

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
serngawy: yes, can replace it and numberOfClustersPerGroup has been added as well.

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: okay canary definition has been removed and add affinity cluster selection as more generic way to define a special subset.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+	// NumberOfClustersPerGroup is the max number of clusters per group. If ClusterGroupPercentage is used the NumberOfClustersPerGroup will be ignored.
+	// +kubebuilder:default:=100
+	// +optional
+	NumberOfClustersPerGroup int32 `json:"numberOfClustersPerGroup,omitempty"` 
+
+	// Cluster Group Type indicates the group type MandatoryClusterGroup or ProgressiveClusterGroup.
+	// +kubebuilder:default:=ProgressiveClusterGroup
+	// +optional
+	ClusterGroupType ClusterGroupType `json:"clustersGroupType"`
```
deads2k, line 85: How would two consumers of the same placement choose different rollout mechanisms?

deads2k:
Controlling rollout for resources consuming Placements is a good idea, but if Placements contain the description of that rollout, it appears to make it impossible to reuse the same Placement by multiple referencing resources that want to have different rollout patterns.  For instance, if we want ClusterManagementAddon/A to rollout at 13% at a time and ClusterManagementAddon/B to rollout at 17% at a time, they could no longer share Placement/one.

Could we get similar benefit by recommending a common rollout strategy construct like https://github.com/open-cluster-management-io/api/blob/main/addon/v1alpha1/types_clustermanagementaddon.go#L182-L193 (open to other possibilities based on experience and feedback)?  It would likely need to include a status section to handle the "last known good" construct we discovered while designing that API.  Consumers would need to embed the construct, but they would gain the ability to decide on the right rollout strategy per consuming instance versus per-placement instance.

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: If I understand  what u mean it depend on the use-case. Not sure if there a case should be progressive sometimes and mandatory other time.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,335 @@
...
+  - clusterGroupIndex: 0
+    clusterGroupType: MandatoryClusterGroup
+    placementDecisions:
+    - ztp-placement-decision-0
+    clusterCount: 10
+  - clusterGroupIndex: 1
+    clusterGroupType: ProgressiveClusterGroup
+    placementDecisions:
+    - ztp-placement-decision-1
+    - ztp-placement-decision-2
```
zhiweiyin318, line 285: when the clusters in each placementDecision are changed, for example some clusters removed from the placement, some new clusters are added into the placement, the clusters in each group will be re-grouped or not ?

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
haoqing0110: I‘m not sure if below case makes sense: user wants to rolling update the canary/affinity clusters at a rate, eg. 20% first, then upgrade the rest of the clusters at another rate, eg 30%. For the placement strategy API in this proposal, seems all the canary/affinity clusters upgrade at once.
```
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ClusterManagementAddOn
metadata:
  name: helloworld
spec:
  installStrategy:
    type: Placements
    placements:
    - name: aws-placement
      namespace: default
      configs:
      - group: addon.open-cluster-management.io
        resource: addonhubconfigs
        name: hub-config-yyy
      rolloutStrategy:
        type: RollingUpdateWithCanary
        rollingUpdateWithCanary:
          placement:
            name: canary-placement
            namespace: default
          maxConcurrentlyUpdating: 30%
    - name: canary-placement
      namespace: default
      configs:
      - group: addon.open-cluster-management.io
        resource: addonhubconfigs
        name: hub-config-yyy
      rolloutStrategy:
        type: RollingUpdate
        rollingUpdate:
          maxConcurrentlyUpdating: 20%
```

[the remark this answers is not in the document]
serngawy: Well, in this example there are 2 different placements aws-placement & canary placement each has its own percentage  so similar could be done with placement with this proposal. Both aws/canary-placement will be without affinityGroup definition as the placement itself is for canary clusters.
Not sure about the use -case but its a bit strange that the canary clusters also need to be divide to groups and the group percentage is also different than the larger group percentage. If you elaborate more in the use-case may we can discuss further.

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: It should not re-grouped when a new cluster added it will be appended to last group (or create new group if last one reach its max capacity).  Check the special case handling section.

serngawy:
> Controlling rollout for resources consuming Placements is a good idea, but if Placements contain the description of that rollout, it appears to make it impossible to reuse the same Placement by multiple referencing resources that want to have different rollout patterns. For instance, if we want ClusterManagementAddon/A to rollout at 13% at a time and ClusterManagementAddon/B to rollout at 17% at a time, they could no longer share Placement/one.
> 
> Could we get similar benefit by recommending a common rollout strategy construct like https://github.com/open-cluster-management-io/api/blob/main/addon/v1alpha1/types_clustermanagementaddon.go#L182-L193 (open to other possibilities based on experience and feedback)? It would likely need to include a status section to handle the "last known good" construct we discovered while designing that API. Consumers would need to embed the construct, but they would gain the ability to decide on the right rollout strategy per consuming instance versus per-placement instance.

Well, the idea for this proposal is to divide the selected clusters to groups and consider one group as affinity group. 
So feedback from rolling out the workloads it is not really the Placement API responsibility its more up to the consumer API like AddonManagemenet to decide what is the last known good feedback.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
jnpacker: From a user perspective it might seem disjoint, if a placement creates and references other placements. It's understood that a placement creates multiple decisions and extending that to include groupings seems like a logical step.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
jnpacker: I'm trying to think about about how this gets implemented... The controller that is consuming the PlacementDecisions would need to understand index 0 is possibly a canary(affinity Group), but I'm thinking that how the groups are consumed is dependent on the controller that is consuming decisions, and Canary or AffinityGroup is an implementation detail that can be used in different ways depending on what the controller is using placement to do.

For example:
1. The implementation requirement might be that it processes ONE cluster in each group before it starts to process all the others. So the ManifestWorkSet controller, will try the first cluster in each group before it tries the rest.
2. The implementation requirement might be that 4x clusters in the first group (this may change over time as order is not guaranteed) should be upgraded and if successful, continue to the rest.
3. The implementation requirement might be to use a label to find the canary (first) clusters, the controller may use a separate placement (borrowing from @qiujian16 's suggestion) and will remove these clusters when it starts to process the groups.

So  we leave it up to the implementation to decided how they want to process clusters, since it will be different for different scenarios and controllers.  We can provide some examples like the three above of how you might use the grouping to accomplish different canary scenarios.

enhancements/sig-architecture/64-placementStrategy/README.md
jnpacker: Agreed.

enhancements/sig-architecture/64-placementStrategy/README.md
jnpacker: since this is more a hint from placement vs something placement actually uses to dictate its output? It might be something we explain in examples for controllers, but do not include in the api.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
serngawy: Okay then, it make more sense to remove the ClusterGroupType as it is up to the controller to decide how to roll out the workload.  I will just add a label for the affinity groups to distinguish between the cluster groups . AND will add the examples mentioned above to the "Workload applier" section.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,360 @@
...
+	ClusterSets []string `json:"clusterSets,omitempty"`
+
+	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
+	// +optional
+	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+            ...
+ 
+	// PlacementStrategy define affinity clusters as subset of the selected clusters and divide the selected clusters to groups.
+	// +optional
+	PlacementStrategy PlacementStrategy `json:"placementStrategy,omitempty"`
```
qiujian16, line 52:
```suggestion
	DecisionGenerator DecisionGenerator `json:"decisionGenerator,omitempty"`
```

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,360 @@
...
+	// +kubebuilder:validation:Enum="5%";"10%";"15%";"20%";"25%";"30%";"50%";"100%"
+	// +kubebuilder:default:=100%
+	// +optional
+	AffinityGroupPercentage string `json:"affinityGroupPercentage,omitempty"`
+}
+
+type PlacementStrategy struct {
+	// Affinity clusters label selector. Affinity clusters will be placed in cluster group index 0 and it will have MandatoryClusterGroup type.
+	// +optional
+	AffinityClusterGroup AffinityClusterGroup `json:"affinityClusterGroup,omitempty"`
```
qiujian16, line 72: Instead of having an AffinityClusterGroup here. I think we should define two decision generator, one is static with a predefined label, and another is by percentage e.g.

```yaml
decisionGenerators:
- type: static
  static:
    labels:
      # this will add labels in related decision
      purpose: canary
    clusterSelector:
      purpose: canary
- type: byGroup
  byGroup:
    numberOfClustersPerGroup: 150
```

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,360 @@
...
+	ClusterSets []string `json:"clusterSets,omitempty"`
+
+	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
+	// +optional
+	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+            ...
+ 
+	// PlacementStrategy define affinity clusters as subset of the selected clusters and divide the selected clusters to groups.
+	// +optional
+	DecisionGenerator DecisionGenerator `json:"decisionGenerator,omitempty"`
```
serngawy, line 52: looks like by mistake I approved the suggest changes , I would prefer DecisionStrategy

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: So as the suggested list , is it possible to have two items with static type ? how about the byGroup type ? can we also have 2 items with byGroup type ?

qiujian16: I think we can have multiple static groups but we can only have one byGroup type and the byGroup type needs to be the last item in the list.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,360 @@
...
+	// NumberOfClustersPerGroup is the max number of clusters per group. If ClusterGroupPercentage is used the NumberOfClustersPerGroup will be ignored.
+	// +kubebuilder:default:=100
+	// +optional
+	NumberOfClustersPerGroup int32 `json:"numberOfClustersPerGroup,omitempty"` 
+}
+```
+
+Adding ClusterGroupStatus to the PlacementStatus API 
+```go
+type ClusterGroupStatus struct {
```
qiujian16, line 91: Not so sure why we need this field, all related details of decisions can be got from decisions if self, right? or I think we should have a list of decision info in the placement e.g.

```yaml
status:
  decisions:
  - name: decision-1
    labels:
      purpose: canary
      ...
    numberOfClusters: 10
```

qiujian16:
```yaml
decisions:
  - names: 
     - decision-1
     - decision-2
    selector:
      matchLabels:
        purpose: canary
    numberOfClusters: 10
```

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,434 @@
...
+}
+
+type PlacementStrategy struct {
+	// Decision Strategy divide the created placement decision to groups and define number of clusters per decision group.
+	// +optional
+	DecisionStrategy DecisionStrategy `json:"decisionStrategy,omitempty"` 
+
+	// Decision Feedback define the feedback values that will be returned with the cluster decision. 
+	// +optional
+	DecisionFeedBack DecisionFeedBack `json:"decisionFeedBack,omitempty"`
```
qiujian16, line 99: do not need to add this, we just need to ensure this is extensible

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,434 @@
...
+type PlacementStrategy struct {
+	// Rollout type either All or Progressive.
+	// +kubebuilder:validation:Enum=All;Progressive
+	// +kubebuilder:default:=All
+	// +optional
+	RolloutType string `json:"rolloutType"`
+
+	// List of the decision groups labels to apply the workload first.
+	// +optional
+	DecisionGroupsToApplyFirst []string `json:"decisionGroupsToApplyFirst"`
```
qiujian16, line 345: what does this mean? Is this for canary?

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -0,0 +1,407 @@
...
+	// +optional
+	NumberOfClustersPerDecisionGroup int32 `json:"numberOfClustersPerDecisionGroup,omitempty"`
+}
+
+type DecisionFeedBack struct {
+	// List of clusterClaims names to have with the clusterDecision.
+	ClusterClaimsNames []string `json:"clusterClaimsNames,omitempty"`
+}
+
+type PlacementStrategy struct {
```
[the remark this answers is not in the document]
serngawy: how about PlacementDecisions or Decisions ?

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: okay, was just to show the case as we discussed.

enhancements/sig-architecture/64-placementStrategy/README.md
serngawy: Any decisionGroup defined under the placement->decisionStrategy, its label can be set here to apply the workloads for it first. Yes, Canary is one case any other decisionGroup label can be used.

enhancements/sig-architecture/64-placementStrategy/README.md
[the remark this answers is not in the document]
qiujian16: DecisionDecorators or DecisionStrategy I think

enhancements/sig-architecture/64-placementStrategy/README.md
qiujian16: I think it depends on the consumers of the placement. I do not think we can define clearly the semantic of ApplyFirst in the placement API.

serngawy: The `DecisionGroupsToApplyFirst` is not part of placement API.  It is an example of how consumer API such as ManifestWorkReplicaSet can consume the DecisionStrategy.

qiujian16: oh I see, that makes sense

qiujian16:
still have some minor comments on naming. looks good in general

serngawy:
> still have some minor comments on naming. looks good in general

If not **placementStrategy** I prefer **placementDecisions**,  it holds the **decisionStrategy** and in future we can add other fields. Also **placementDecisions** name is relevant to the generated placementDecision CRs

qiujian16 (commented):
/lgtm

lgtm in general. I think we still need some adjustment on the API. But we can decide that on API PR.

PR #89, opened 2023-03-06
Add initial design for OperatorPolicy

JustinKuli:
Refs:
 - https://issues.redhat.com/browse/ACM-3097

For easier reading: https://github.com/JustinKuli/ocm-enhancements/blob/89-operator-policy/enhancements/sig-policy/89-operator-policy-kind/README.md

### DIFF
--- a/enhancements/sig-policy/89-operator-policy-kind/README.md
+++ b/enhancements/sig-policy/89-operator-policy-kind/README.md
@@ -74,6 +74,7 @@ metadata:
 spec:
   remediationAction: enforce # or inform
   severity: medium
+  complianceType: musthave # or mustnothave
   operatorGroup: # optional
     name: my-operator-group
     namespace: own-namespace
@@ -94,10 +95,9 @@ spec:
     source: my-catalog
     sourceNamespace: my-catalog-namespace
     startingCSV: my-operator.v0.1.0 # optional
-  allowedCSVs:
+  versions:
     - my-operator.v0.1.1
     - my-operator.v0.2.0
-  # - "*" # Allow all
   removalBehavior:
     operatorGroups: DeleteIfUnused
     subscriptions: Delete
@@ -106,9 +106,9 @@ spec:
     customResourceDefinitions: Keep
     apiServiceDefinitions: Keep
   statusConfig:
-    catalogSourceUnhealthy: Condition
+    catalogSourceUnhealthy: StatusMessageOnly
     deploymentsUnavailable: NonCompliant
-    upgradesAvailable: Condition
+    upgradesAvailable: StatusMessageOnly
     upgradesProgressing: NonCompliant
 ```
 
@@ -119,11 +119,18 @@ what is missing, and that it will not be created because the policy is not being
 is set to `enforce`, OperatorGroups and Subscriptions may be created by the controller, InstallPlans
 may be approved, and resources may be removed when the policy is deleted.
 
+The `complianceType` is either "musthave" or "mustnothave" and follows the idea of the same field in
+ConfigurationPolicy. In "musthave" mode, the operator must be present on the cluster as specified in
+the policy for the policy to be marked as Compliant. In "mustnothave" mode, if the specified
+operator is on the cluster (found by matching the name, namespace, source and sourceNamespace, and
+optionally matching one of the specified versions), then the policy will be marked as NonCompliant.
+
 The optional `spec.operatorGroup` section defines an
 [OperatorGroup](https://github.com/operator-framework/api/blob/master/crds/operators.coreos.com_operatorgroups.yaml).
 If not specified, and if the operator supports this InstallMode, then it will ensure that an
 [AllNamespaces](https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#targetnamespaces-and-their-relationship-to-installmodes)
-OperatorGroup exists in the namespace of the subscription.
+OperatorGroup exists in the namespace of the subscription. The operatorGroup is not checked in
+"mustnothave" mode.
 
 The `spec.subscription` section includes the fields from the spec of an [OLM
 Subscription](https://github.com/operator-framework/api/blob/master/crds/operators.coreos.com_subscriptions.yaml).
@@ -133,13 +140,17 @@ The `installPlanApproval` field is optional. If the policy is in `enforce` mode 
 are restricted as specified below, then the `installPlanApproval` field on the Subscription will
 always be Manual, regardless of the setting here.
 
-The `spec.allowedCSVs` list specifies what installed versions are considered Compliant when the
-policy is in `inform` mode, and which InstallPlans can be approved when in `enforce` mode. Only
-exact matches are considered, except for a special value, "*" which matches any version. This field
-must be set, and must have at least one item.
+In "musthave" mode, the `spec.versions` list specifies what installed versions are considered
+Compliant when the policy is in `inform` mode, and which InstallPlans can be approved when in
+`enforce` mode. Conversely, in "mustnothave" mode, this list specifies which versions are considered
+NonCompliant. Only exact matches are considered. If the list is unset or empty, then any version on
+the cluster will be considered a match.
 
-The `spec.removalBehavior` field allows configuration of what will be removed when the policy is
-deleted. Each kind that might be deleted should support `Keep` and `Delete`, and can potentially
+The `spec.removalBehavior` field allows configuration of what may be removed by the controller. It
+has no effect when the policy is in "inform" mode. When the policy is in "musthave" mode, and the
+policy is deleted, then controller will remove the types marked for deletion here. In "mustnothave"
+mode, the specified types will be removed if they appear, in order to drive the cluster towards
+compliance. Each kind that might be deleted should support `Keep` and `Delete`, and can potentially
 have additional values specific to the resources in question. For example, the OperatorGroup should
 likely only be removed if it isn't being used by another subscription. 
 
@@ -276,6 +287,7 @@ metadata:
 spec:
   remediationAction: enforce
   severity: medium
+  complianceType: musthave
   subscription:
     channel: stable
     name: strimzi-kafka-operator
@@ -284,7 +296,7 @@ spec:
     source: community-operators
     sourceNamespace: openshift-marketplace
     startingCSV: strimzi-cluster-operator.v0.35.0
-  allowedCSVs:
+  versions:
     - strimzi-cluster-operator.v0.35.0
 ```
 
@@ -304,6 +316,7 @@ metadata:
 spec:
   remediationAction: enforce
   severity: medium
+  complianceType: musthave
   subscription:
     channel: stable
     name: strimzi-kafka-operator
@@ -312,7 +325,7 @@ spec:
     source: community-operators
     sourceNamespace: openshift-marketplace
     startingCSV: strimzi-cluster-operator.v0.35.0
-  allowedCSVs:
+  versions:
     - strimzi-cluster-operator.v0.35.0
     - strimzi-cluster-operator.v0.35.1
 ```
@@ -331,6 +344,7 @@ metadata:
 spec:
   remediationAction: enforce
   severity: medium
+  complianceType: musthave
   subscription:
     channel: stable
     name: strimzi-kafka-operator
@@ -339,7 +353,7 @@ spec:
     source: community-operators
     sourceNamespace: openshift-marketplace
     startingCSV: strimzi-cluster-operator.v0.35.0
-  allowedCSVs:
+  versions:
     - strimzi-cluster-operator.v0.35.0
   statusConfig:
     upgradesAvailable: NonCompliant
@@ -383,6 +397,7 @@ metadata:
 spec:
   remediationAction: inform
   severity: medium
+  complianceType: musthave
   subscription:
     channel: stable
     name: strimzi-kafka-operator
@@ -390,8 +405,6 @@ spec:
     installPlanApproval: Automatic
     source: community-operators
     sourceNamespace: openshift-marketplace
-  allowedCSVs:
-    - "*"
   statusConfig:
     catalogSourceUnhealthy: NonCompliant
     deploymentsUnavailable: NonCompliant
@@ -416,6 +429,7 @@ metadata:
 spec:
   remediationAction: enforce
   severity: medium
+  complianceType: musthave
   operatorGroup:
     name: og-strimzi
     namespace: strimzi-app-one
@@ -429,8 +443,6 @@ spec:
     installPlanApproval: Automatic
     source: community-operators
     sourceNamespace: openshift-marketplace
-  allowedCSVs:
-    - "*"
 ```
 
 #### Story 6
@@ -447,6 +459,7 @@ metadata:
 spec:
   remediationAction: enforce
   severity: medium
+  complianceType: musthave
   subscription:
     channel: stable
     name: strimzi-kafka-operator
@@ -454,8 +467,6 @@ spec:
     installPlanApproval: Automatic
     source: community-operators
     sourceNamespace: openshift-marketplace
-  allowedCSVs:
-    - "*"
   removalBehavior:
     operatorGroups: DeleteIfUnused
     subscriptions: Delete
@@ -467,6 +478,38 @@ spec:
 
 It should be noted that the `removalBehavior` specified here is just the default setting made
 explicit.
+
+#### Story 7
+
+As a policy user, I want to ensure that a certain version of an operator is not present on the
+cluster. If that version of the operator is found, I want its installation to be deleted.
+
+```yaml
+apiVersion: policy.open-cluster-management.io/v1beta1
+kind: OperatorPolicy
+metadata:
+  name: strimzi-policy
+spec:
+  remediationAction: enforce
+  severity: medium
+  complianceType: mustnothave
+  subscription:
+    channel: stable
+    name: strimzi-kafka-operator
+    namespace: strimzi-app-one
+    installPlanApproval: Automatic
+    source: community-operators
+    sourceNamespace: openshift-marketplace
+  versions:
+    - "strimzi-cluster-operator.v0.35.0"
+  removalBehavior:
+    operatorGroups: DeleteIfUnused
+    subscriptions: Delete
+    clusterServiceVersions: Delete
+    installPlans: Keep
+    customResourceDefinitions: Keep
+    apiServiceDefinitions: Keep
+```
 
 ### Implementation Details/Notes/Constraints [optional]
 
--- a/enhancements/sig-policy/89-operator-policy-kind/metadata.yaml
+++ b/enhancements/sig-policy/89-operator-policy-kind/metadata.yaml
@@ -10,5 +10,5 @@ reviewers:
 approvers:
   - TBD
 creation-date: "2023-03-06"
-last-updated: "2023-07-18"
+last-updated: "2023-07-27"
 status: implementable


enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,192 @@
...
+on a cluster. A new controller will help cover specialized cases that come up when managing operator
+installations, rather than teaching the existing configuration-policy-controller to deal with
+specific types.
+
+### Goals
+
+1. It should be easy to install an operator on a managed cluster with just one policy.
+2. It should be possible to limit updates to operators on clusters, so they do not upgrade beyond a
+   known good version defined in the policy.
+3. The policy managing the operator should clearly report back the status of the installation, with
```
gparvin, line 39: Today there are cases where I think something happens in the wrong order and you end up with a failed operator install.  You can't resolve except by doing some cleanup and then trying again.  Does that sound familiar and would that case be covered because the "ordering" is no longer a concern?  Or would there still be a need for a retry on failure?

JustinKuli: By having the new controller manage the "imperative" steps that could be done in the wrong order, I think it can avoid at least some of the issues we've seen before when we just apply all the resources at once with ConfigurationPolicies.

It might be impossible to prevent all installation errors, but depending on what's required to "cleanup and try again", I think that could also be handled by the controller automatically.

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+- [ ] User-facing documentation is created in
+  [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Enhance the policy framework with a new kind, `OperatorPolicy`, to allow for better management of
+OLM Operator installations on managed clusters. The controller for this kind would be included in
+the configuration-policy-controller container, on managed clusters. The new controller would better
+handle imperative installation steps for OLM objects, and provide better support for upgrade and
+uninstallation scenarios.
```
fgiloux, line 18: > The new controller would better handle imperative installation steps for OLM objects

I guess it is referring to the fact that you need to approve the installation when the install plan approval strategy has been set to manual in the subscription.
I would propose to rephrase to: "The new controller would better handle installation requiring approvals of OLM operators."
Technically it is possible to handle everything in a declarative way. At the end of the day the API is a resource, whose fields can be modified. I understand the challenge with InstallPlans that are a second level resources (not directly created) but need to be modified for the approval.

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Motivation
+
+It is currently possible to install OLM Operators on managed clusters with `ConfigurationPolicies`,
+but it can be complex and error-prone. For example, multiple policies are often needed in order to
+manage the different OLM resources such as OperatorGroup, Subscription, InstallPlan, and
+ClusterServiceVersion. Upgrading or uninstalling an operator can also be difficult with
+ConfigurationPolicies, since it is not generally as easy as just updating one value or deleting a
+resource.
```
fgiloux, line 27: > Upgrading or uninstalling an operator can also be difficult with ConfigurationPolicies, since it is not generally as easy as just updating one value or deleting a resource.

Can you be specific? It should actually be the case for upgrades. Is it because of the InstallPlan approval?

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+OperatorGroup will be created.
+
+The `spec.subscription` section includes fields from the spec of an [OLM
+Subscription](https://github.com/operator-framework/api/blob/master/crds/operators.coreos.com_subscriptions.yaml),
+excluding `config` and `installPlanApproval`. The `endingCSV` field allows configuration of which
+InstallPlans will be approved by the controller when the policy is set to `enforce`.
+
+The `spec.requireLatestVersion` field configures whether the "latest" version of the operator
+defined in the catalog on this cluster "should" be installed. Note that if the "latest" version is
+greater than the allowed `endingCSV`, then that version will not be automatically approved even if
```
fgiloux, line 106: > The spec.requireLatestVersion field configures whether the "latest" version of the operator defined in the catalog on this cluster "should" be installed.

Wouldn't it be the default behavior?

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+For simplicity, the initial implementation will not use a `config` field in the Subscription, which
+could be used to configure things like `nodeAffinity`. Since this is a feature in OLM, it may be
+something that is added to OperatorPolicy later.
+
+### Risks and Mitigation
+
+### Open Questions [optional]
+
+1. Is making `spec.operatorGroup` optional, and defaulting to an AllNamespaces OperatorGroup, the
+   best option?
```
fgiloux, line 279: I think so.

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+    #       mydomain.io/prod: "true"
+    serviceAccountName: my-og-account # optional
+  subscription:
+    channel: stable
+    name: my-operator
+    namespace: own-namespace
+    source: my-catalog
+    sourceNamespace: my-catalog-namespace
+    startingCSV: my-operator.v0.1.0 # optional
+    endingCSV: Any # or None or e.g. 'my-operator.v0.2.0`
```
jc-rh, line 83: What does none mean?

enhancements/sig-policy/89-operator-policy-kind/README.md
jc-rh: In addition to the need of install plan approval, issues we ran into: (may be specific to disconnected env)
- catalog src policy becomes compliant before it's truly ready
- subscription policy with 'AtLatestKnown` takes a while to become non-compliant after catalog src is ready
- after install plan/subscription status is updated, there is still a chance for the csv to be stuck in failed state

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+### OperatorPolicy Syntax
+
+```yaml
+apiVersion: policy.open-cluster-management.io/v1beta1
+kind: OperatorPolicy
+metadata:
+  name: my-operator
+spec:
+  remediationAction: enforce # or inform
```
mprahl, line 63: What is checked in `inform` mode? What kind of non-compliant message would you get?

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+excluding `config` and `installPlanApproval`. The `endingCSV` field allows configuration of which
+InstallPlans will be approved by the controller when the policy is set to `enforce`.
+
+The `spec.requireLatestVersion` field configures whether the "latest" version of the operator
+defined in the catalog on this cluster "should" be installed. Note that if the "latest" version is
+greater than the allowed `endingCSV`, then that version will not be automatically approved even if
+the `remediationAction` is set to enforce: instead, the policy will be NonCompliant.
+
+The `pruneOnDeletion` field allows configuration of what will be removed when the policy is deleted.
+A "recommended" setting for most workloads would be `[subscriptions, clusterserviceversions]`,
```
mprahl, line 110: Is this the default?

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+    endingCSV: strimzi-cluster-operator.v0.33.1
+  requireLatestVersion: true
+status:
+  compliant: NonCompliant
+  conditions:
+    - lastTransitionTime: '2023-03-14T15:09:26Z'
+      message: 'operator can be updated to v0.33.2'
+      reason: OperatorUpdateAvailable
+      status: 'True'
+      type: NonCompliant
```
mprahl, line 242: Did you mean this?
```suggestion
      status: 'False'
      type: Compliant
```

enhancements/sig-policy/89-operator-policy-kind/README.md
mprahl: I like this idea.

enhancements/sig-policy/89-operator-policy-kind/README.md
JustinKuli: It probably should be... I'm still thinking about the best way to do that. Since I want it to be possible to configure nothing to be deleted, an empty list does not imply the "default" behavior. And filling it with an empty list, vs not specifying the field at all, feels too subtle of a difference - I think it could be confusing.

enhancements/sig-policy/89-operator-policy-kind/README.md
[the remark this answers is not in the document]
JustinKuli: Great comments! I think most are covered in my recent update, but I will call out these specific points:

5. I updated the goal to specify that *creation* of CatalogSources will not be covered - as noted by other details, the CatalogSource's health will be monitoried.

7. Renamed `requireLatestVersion` to `latestVersionNecessary`.

9. Added. Initially I wanted to try and keep things smaller here, but that doesn't seem necessary.

10. I didn't want to over-specify in this design what status might be reported. I do agree that what you've mentioned here is important and likely should be reported. I've added an open question for how much specification is required here as part of the design.

enhancements/sig-policy/89-operator-policy-kind/README.md
JustinKuli: I don't know what you mean.

One case I was trying to account for was where the policy was created with the (current) "newest" version available and tested, but not wanting to immediately update whenever something new comes out. Instead, the user could use this so that they are notified (via a NonCompliant policy) when an update is available, so they can test it out.

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,310 @@
...
+--for=...`. At the very least, this will include a condition of type "Compliant", but additional
+fields for internal state may be exposed, for example when waiting for a ClusterServiceVersion after
+approving an InstallPlan.
+
+### User Stories
+
+#### Story 1
+
+As a policy user, I want to install an operator of a specific version, from a specific source, on my
+managed clusters. I should be able to view the health of these installations through policies.
```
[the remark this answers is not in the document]
JustinKuli: In my opinion of what a user story should be, it doesn't make sense to pull the health reporting out into a separate story. I do agree it could help feature development, but I didn't want to artificially split it up here.

brian-jarvis:
One thing to note, InstallPlans contain details of all operators in the same namespace.  You can't approve/upgrade one individually.   This causes issues for customers today (perhaps olm 2.0 will correct this) trying to manage manually approving operators through cluster environments.

Given the InstallPlan below:
- which OperatorPolicy should own updating it?
- what happens if one OperatorPolicy has limited openshift-gitops to 1.7.x and servicemesh to 2.3.x?  

`apiVersion: operators.coreos.com/v1alpha1
kind: InstallPlan
metadata:
  creationTimestamp: "2023-03-24T15:23:01Z"
  generateName: install-
  generation: 1
  labels:
    operators.coreos.com/kiali-ossm.openshift-operators: ""
    operators.coreos.com/servicemeshoperator.openshift-operators: ""
  name: install-j9k5k
  namespace: openshift-operators
  ownerReferences:
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: devspaces
    uid: 1266cac3-5703-405d-8088-4c5f5cfbee28
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: devworkspace-operator-fast-redhat-operators-openshift-marketplace
    uid: 09cfe1be-bd9c-4ca5-b6ae-26bc101a134f
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: jaeger-product
    uid: b58625e0-2e4a-4229-8dcf-b73d28e411ef
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: kiali-ossm
    uid: 7023b9d3-6c48-4611-951f-e36d72289350
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: openshift-gitops-operator
    uid: 33f2927b-6033-42eb-b743-a1606fb35d4f
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: openshift-pipeline-operator
    uid: 68200532-924d-46e0-8882-5bf3b0478e6c
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: rh-service-binding-operator
    uid: 0e093920-e581-4fde-8438-18a33ff1bb88
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: Subscription
    name: servicemeshoperator
    uid: ec09113a-7aac-46f9-be44-37ffa55fc79a
  resourceVersion: "1891393943"
  uid: 5b4c51ca-a97c-40a2-8cae-f66e4c6e8d11
spec:
  approval: Automatic
  approved: true
  clusterServiceVersionNames:
  - openshift-gitops-operator.v1.8.1
  - kiali-operator.v1.57.6
  - servicemeshoperator.v2.3.2
  - openshift-pipelines-operator-rh.v1.9.2
`

JustinKuli:
Newest revision notes. Overall the idea was to improve ease-of-use, and incorporate comments from the previous reviews.

The configuration of what upgrades are allowed has been changed from the `endingCSV` field that was in `spec.subscription`, to a new field `spec.allowedCSVs`. This makes the subscription field almost exactly match the OLM subscription spec (only adding a `namespace` field), and makes the behavior of the policy in `inform` mode more obvious.

The `spec.latestVersionNecessary` field has been brought into a new section `spec.violationConditions` which will hold more configurations of that type: things that *could* be cause for a violation, but not necessarily for all users/policies.

The specification of the `status` is more complete now.

The compliance event now has an example YAML.

There are more User Stories, and they are generally smaller / more specific, ideally each should be implementable in a reasonable amount of time (some requiring other stories to already be completed).

Some details on approving InstallPlans has been added to the notes, with observations from existing InstallPlans on clusters.

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,557 @@
...
+For more information on the `config` field, see:
+https://github.com/operator-framework/api/blob/v0.17.3/pkg/operators/v1alpha1/subscription_types.go#L42.
+The `installPlanApproval` field is optional. If the policy is in `enforce` mode and the allowed CSVs
+are restricted as specified below, then the `installPlanApproval` field on the Subscription will
+always be Manual, regardless of the setting here.
+
+The `spec.allowedCSVs` list specifies what installed versions are considered Compliant when the
+policy is in `inform` mode, and which InstallPlans can be approved when in `enforce` mode. Only
+exact matches are considered, except for a special value, "*" which matches any version. This field
+must be set, and must have at least one item.
```
jc-rh, line 139: Is there a valid case for this to be set with empty value? If not, is it better to keep it optional and allow all if not set?

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,557 @@
...
+    clusterServiceVersions: Delete
+    installPlans: Keep
+    customResourceDefinitions: Keep
+    apiServiceDefinitions: Keep
+  statusConfig:
+    catalogSourceUnhealthy: Condition
+    deploymentsUnavailable: NonCompliant
+    upgradesAvailable: Condition
+    upgradesProgressing: NonCompliant
+```
```
irinamihai, line 113: How would we use this to ensure the Subscription's status has **status.state: AtLatestKnown** without specifying the exact CSV?

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,557 @@
...
+### Non-Goals
+
+1. The installation of OLM on managed clusters will not be automated by this enhancement.
+2. Management of "Operands" will not be included in this new type - they should be defined in
+   separate `ConfigurationPolicies`, possibly with a
+   [dependency](../74-order-policy-execution/README.md) on the `OperatorPolicy`. ("Operand" here
+   refers to resources that would be created by a user, and managed by the operator, eg instances of
+   CRDs which are defined by the operator.)
+3. Creation of `CatalogSources` will not be included in the new type - they should be defined in
+   separate `ConfigurationPolicies`.
```
jc-rh, line 63: I agree that the catalog source should be handled in a different policy. Just to highlight that currently the catalog source API does not provide `observerdGeneration` in its status. Any thing we derive from its status can be stale if its spec just got updated. This can make the policy ordering ineffective. The operator policy can go/stay compliant when it shouldn't be. Same concern goes with the `catalogSourceUnhealthy` field in `statusConfig`. It wouldn't do its job if we set it to `NonCompliant` in cases like this. I opened an RFE for this but it is set to lowest priority possible ...
https://issues.redhat.com/browse/RFE-3846

I've linked it to this enhancement JIRA. If we can push for it together and get it implemented by OLM, this enhancement will be more successful :)

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,557 @@
...
+deleted. Each kind that might be deleted should support `Keep` and `Delete`, and can potentially
+have additional values specific to the resources in question. For example, the OperatorGroup should
+likely only be removed if it isn't being used by another subscription. 
+
+The `spec.statusConfig` field allows the author to specify what effect specific resource statuses
+will have on the OperatorPolicy status and compliance events. The currently planned values are:
+`Condition`, which will record the status in `status.conditions` and in the compliance event; and
+`NonCompliant`, which additionally updates the `status.compliant` field to NonCompliant which can
+cause the root policy on the hub to become NonCompliant and may throw an alert. Additional values
+could be added in the future.
```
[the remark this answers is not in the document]
jc-rh: I guess if we put `NonCompliant` for both `upgradesAvailable` and `upgradesProgressing`, it might cover 99% of the whole process. It might be hard to guarantee it will always work like that as things evolve though. Is it better to introduce an explicit field for `atLatestKnown`? This is the main use case for telco and it's part of our current du reference spec.

enhancements/sig-policy/89-operator-policy-kind/README.md
JustinKuli: Good question! I hadn't realized it, but yes, an empty list here could be useful: in particular when the policy is in `inform` mode, that could be used to verify that an operator is *not* installed (similar to `mustnothave` in ConfigurationPolicy).

The design right now doesn't specify whether already-existing subscriptions should be deleted by an OperatorPolicy in `enforce` mode, that might need to be considered...

enhancements/sig-policy/89-operator-policy-kind/README.md
JustinKuli: Setting `upgradesAvailable: NonCompliant` would cause the policy to become non-compliant when the Subscription has `status.state: UpgradePending`.

enhancements/sig-policy/89-operator-policy-kind/README.md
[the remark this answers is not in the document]
JustinKuli: I realized I didn't fully specify this, but I was expecting `upgradesAvailable` to basically correspond to the `atLatestKnown` case... is there a situation I'm missing there? I guess I don't know what the Subscription's state will look like if it's not at the newest version, but it can't upgrade to that because of a version incompatibility or something.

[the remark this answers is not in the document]
jc-rh: Yes, there are other states in addition to these two:
`const (
	SubscriptionStateNone             = ""
	SubscriptionStateFailed           = "UpgradeFailed"
	SubscriptionStateUpgradeAvailable = "UpgradeAvailable"
	SubscriptionStateUpgradePending   = "UpgradePending"
	SubscriptionStateAtLatest         = "AtLatestKnown"
)`

enhancements/sig-policy/89-operator-policy-kind/README.md
```diff
@@ -0,0 +1,557 @@
...
+  subscription:
+    config: {} # optional
+    channel: stable
+    name: my-operator
+    namespace: own-namespace
+    installPlanApproval: Automatic # may be overridden to Manual by other settings
+    source: my-catalog
+    sourceNamespace: my-catalog-namespace
+    startingCSV: my-operator.v0.1.0 # optional
+  allowedCSVs:
```
serngawy, line 97: better name could be **approvedCSV** as allowedCSV meaning more like it is allowed to show up for subscriptions but not approved.

JustinKuli: I'm thinking about this in the new possible mode of a "mustnothave" OperatorPolicy. In that case, it would be nice to be able to supply a list of specific versions I *don't* want to allow on the managed cluster.

Would a simpler name like `versions` work? Then in a "musthave" OperatorPolicy, this list indicates the versions that can be present for the policy to be compliant. And in "mustnothave" mode, the list would indicate versions that must not be present.

serngawy: okay, make sense for mustnothave case.

jc-rh: Are you saying two policies are needed if we want to allow some versions and disallow some others?

JustinKuli: @jc-rh it depends on if you know the specific versions you want to allow. If you make a "musthave" policy with specific versions, than any versions other than what is specified will make the policy NonCompliant (and in enforce mode, the controller might try to remove that operator).

If you can't restrict the allowed versions to a specific list - maybe you want something like "any version, except x, y, or z" - then in this proposal, two policies would be necessary: a "musthave" with no versions specified, and a "mustnothave" with x, y, and z. (This case wasn't really covered by any previous design iterations.)

I was anticipating "mustnothave" being more useful if the operators aren't necessarily being managed by an OperatorPolicy.

jc-rh: Hmm, OK. Our main use case would be `AtLatestKnown`. It would be a pain to update allowed version in the policy for every single update. Occasionally, when they want to disallow a problematic version, it would be awkward to list all currently running/upgradable versions as allowed.

mprahl (approved):
/hold for other reviews

PR #91, opened 2023-04-03
Add KEP for ClusterPermission

mikeshng:


enhancements/sig-architecture/98-managed-rbac/README.md
```diff
@@ -0,0 +1,185 @@
...
+### Component & API
+
+We purpose to adding a new custom resource named
+`ManagedRBAC` introduced into OCM by this proposal:
+
+A sample of the `ManagedRBAC` will be:
+
+```yaml
+apiVersion: rbac.open-cluster-management.io/v1alpha1
+kind: ManagedRBAC
```
qiujian16, line 80: maybe `ClusterPermission`?

enhancements/sig-architecture/98-managed-rbac/README.md
```diff
@@ -0,0 +1,185 @@
...
+  name: my-managed-rbac
+  namespace: cluster1
+spec:
+  clusterRole:
+    rules:
+    - apiGroups: ["apps"]
+      resources: ["deployments"]
+      verbs: ["update"]
+  roles:
+  - namespace: default
```
qiujian16, line 91: how about a namespace selector? What if you want to multiple namespaces to have the same rule?

enhancements/sig-architecture/98-managed-rbac/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+```yaml
+apiVersion: rbac.open-cluster-management.io/v1alpha1
+kind: ManagedRBAC
+metadata:
+  name: my-managed-rbac-with-refs
+  namespace: cluster1
+spec:
+  roleRefs:
+  - kind: Role
```
qiujian16, line 146: this create a rolebinding on which ns?

qiujian16 (commented):
I think some thing missing here is:
1. who can create this permission API on hub, how to control it?
2. in the manifestwork, we have an executor field so not all user can create role/clusterrole if that feature is enabled. How does that impact this API
3. Also it will be good to compare this proposal with the alternative that we have a permission API within a clusterset rather than per cluster, what is the pros and cons?

mikeshng:
@qiujian16 could you please take a look again when you have some cycles? Thank you!

enhancements/sig-architecture/98-cluster-permission/README.md
```diff
@@ -0,0 +1,219 @@
...
+of the `ManifestWork` is the work agent which has privilege to create
+RBAC resources on the managed cluster. Therefore, the spec of the `ManifestWork`
+will hard the klusterlet-work-sa service account hardcoded:
+
+```
+  executor:
+    subject:
+      type: ServiceAccount
+      serviceAccount:
+        namespace: open-cluster-management-agent
```
zhujian7, line 147: nit, for hosted mode klusterlet, the sa is not in the `open-cluster-management-agent` ns.

mikeshng: Thank you @zhujian7 for the great comment. Do you have any recommendation what we should do in the hosted mode case?

zhujian7: I'd suggest keeping the executor as nil, but granting the "clusterPermission-controller-serviceaccount" permission to [use the `system:serviceaccount::klusterlet-work-sa` as the executor](https://github.com/open-cluster-management-io/api/blob/54a83a1650d6aa62f4bfe14b719a88a9f191b05a/feature/feature.go#L49-L50)

because if the executor is nil, the work-webhook will [assume it is using the klusterlet-work-sa as the executor](https://github.com/open-cluster-management-io/work/blob/main/pkg/webhook/v1/manifestwork_validating.go#L83-L94).

mikeshng: Updated enhancement based on Zhu Jian's recommendations.

enhancements/sig-architecture/98-cluster-permission/README.md
```diff
@@ -0,0 +1,226 @@
...
+It will be useful if there is an approval process that a cluster admin can approve,
+which will allow for other users to create `ClusterPermission` resources.
+- Support creating a RoleBinding/ClusterRoleBinding without defining any Role/ClusterRole
+rules.
+
+## Design
+
+### Component & API
+
+We purpose to adding a new custom resource named
```
qiujian16, line 79:
```suggestion
We propose to adding a new custom resource named
```

enhancements/sig-architecture/98-cluster-permission/README.md
```diff
@@ -0,0 +1,226 @@
...
+      status: True    
+```
+
+The `ClusterPermission` resource is expected to be created under the
+"cluster namespace" which is a namespace with the same name as
+the managed cluster, the Role/RoleBinding/ClusterRole/ClusterRoleBinding
+delivered to the managed cluster will have the same name as the
+`ClusterPermission` resource.
+
+We purpose adding a new hub cluster controller that watches
```
qiujian16, line 130:
```suggestion
We propose adding a new hub cluster controller that watches
```

enhancements/sig-architecture/98-cluster-permission/README.md
```diff
@@ -0,0 +1,226 @@
...
+      - key: foo.com/managed-state
+        operator: In
+        values:
+        - managed
+    rules:
+    - apiGroups: [""]
+      resources: ["configmaps"]
+      verbs: ["update"]
+  subject:
+    apiGroup: authentication.open-cluster-management.io
```
qiujian16, line 113: will this create a clusterrolebinding for clusterrole? What if I want to create a rolebinding for clusterrole?

qiujian16 (commented):
just some minor comments. I think we also need to clarify the use cases on who will and how to consume this API. Will that be mainly use together with cluster-proxy and user can use it to access? or will some controller will use the managedServiceAccount token to access spoke apiserver?

enhancements/sig-architecture/98-cluster-permission/README.md
mikeshng: Fixed as suggested. Thanks.

enhancements/sig-architecture/98-cluster-permission/README.md
mikeshng: This is a great point. Added example of how to handle rolebinding and clusterrolebinding separately.

PR #92, opened 2023-04-06
Grant additional access to hub templates

ckandag:
Refs:
https://issues.redhat.com/browse/ACM-4179


ckandag:
FYI @gparvin

mprahl:
@ckandag thanks for the idea of using a service account! Since you're no longer working on Open Cluster Management, let's close this and I'll use your idea for inspiration.

PR #93, opened 2023-04-13
Add KEP for AddOnPlacementScoreGenerator

mvazquezc:
As discussed at the community meeting on April 6th, this is the KEP for adding the `AddOnPlacementScoreGenerator` API.

### DIFF
--- a/enhancements/sig-architecture/32-extensiblescheduling/score-generator/README.md
+++ b/enhancements/sig-architecture/32-extensiblescheduling/score-generator/README.md
@@ -1,0 +1,196 @@
+# New API: AddOnPlacementScoreGenerator
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implemented`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal will be adding a new OCM API named `AddOnPlacementScoreGenerator`
+which helps the administrators to provide custom values to Addon controllers
+that generate `AddOnPlacementScores`.
+
+A valid `AddOnPlacementScoreGenerator` resource should be in a "cluster namespace" and
+the associated config resources will be delivered to the associated managed cluster
+with that "cluster namespace".
+
+## Motivation
+
+### Influence Addon controllers behavior
+
+Currently, when writing an Addon in order to extend the scheduling capabilities of OCM
+there is no way to influence behavior of that addon via an API. The controller will run
+and update the status for the `AddOnPlacementScores` object.
+
+One of the use-cases we want to cover is testing latency from managed clusters to a set
+of user-define locations. With the current `AddOnPlacementScores` implementation we would
+need to hard code these locations or maybe use something like a `ConfigMap` that gets consumed
+by the controller. We don't find these solutions flexible enough, so our proposal would be having
+a new API to influence the behavior of such controllers.
+
+Let's say I want to test latencies to redhat.com and google.com and place my application based on
+the managed cluster with the lowest latency to redhat.com.
+
+Providing I created the required controller with the hardcoded locations (redhat.com and google.com)
+an `AddOnPlacementScores` similar to this one would be created on each managed cluster running this addon:
+
+~~~yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: AddOnPlacementScore
+metadata:
+  name: cluster1-generator
+  namespace: cluster1
+status:
+  conditions:
+  - lastTransitionTime: "2021-10-28T08:31:39Z"
+    message: AddOnPlacementScore updated successfully
+    reason: AddOnPlacementScoreUpdated
+    status: "True"
+    type: AddOnPlacementScoreUpdated
+  validUntil: "2021-10-29T18:31:39Z"
+  scores:
+  - name: "redhat-com-avgLatency"
+    value: 30
+  - name: "google-com-avgLatency"
+    value: 50
+~~~
+
+Now, a `Placement` like this could be used:
+
+~~~yaml
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Placement
+metadata:
+  name: latency-placement
+  namespace: ns1
+spec:
+  numberOfClusters: 3
+  prioritizerPolicy:
+    mode: Exact
+    configurations:
+      - scoreCoordinate:
+          type: AddOn
+          addOn:
+            resourceName: cluster1-generator
+            scoreName: redhat-com-avgLatency
+        weight: -1
+~~~
+
+Other applications may use latency to google.com.
+
+Now, we want to add linux.com to the list of locations to test. With the current implementation we 
+will need to edit the code of the addon controller and include the test to linux.com + the result to be added
+to the `AddOnPlacementScore`.
+
+To fix above issue, the proposal is to create a new API `AddOnPlacementScoreGenerator`, which could 
+look like this for the example above:
+
+~~~yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: AddOnPlacementScoreGenerator
+metadata:
+  name: cluster1-generator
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/32-extensiblescheduling/score-generator/README.md
```diff
@@ -0,0 +1,192 @@
...
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: AddOnPlacementScoreGenerator
+metadata:
+  name: cluster1-generator
+  namespace: cluster1
+spec:
+  addOnPlacementSelector:
+    name: latency
+    namespace: cluster1
+  latencies:
```
haoqing0110, line 149: My question is, should this field `latencies` be a more general field? and how could a user define a URL here, what's the requirement for the URL and the output?

mvazquezc: Hey @haoqing0110 yes, other than more generic to me this would be "any string" and even we could have more than one, something like this for example: 

~~~yaml
kind: AddOnPlacementScoreGenerator
metadata:
  name: cluster1-generator
  namespace: cluster1
spec:
  addOnPlacementSelector:
    name: latency
    namespace: cluster1
  latencies:
    - name: redhat-com-avgLatency
      url: https://redhat.com
      runs: 2
      waitBetweenRuns: 10s
    - name: google-com-avgLatency
      url: https://google.com
    - name: linux-com-avgLatency
      url: https://linux.com
  energy:
    - name: some-name
      someValue: some-value
~~~

haoqing0110: Thanks for the explanation. This `AddOnPlacementScoreGenerator` sounds like a configuration for the Addon controller, and to generate different scores, developers still need to develop their own addon controller, correct?  Sounds similar to the idea of https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/58-addon-configuration, not sure if this can satisfy the motivation of this proposal.

haoqing0110: cc @qiujian16 , want to know your suggestions.

mvazquezc: @haoqing0110 the use cases defined by 58-addon-configuration would work for us, I have checked the examples and the example with `ManagedProxyConfiguration`  is very similar to what we want.

And answering your second comment, yes, the same controller could use `latencies` and `energy` data.

qiujian16: I think this is more like a cluster scoped resource that applies to all clusters, it is less likely that we want different configs for different clusters. Also how could we parse the result of the http responses? It might be straightforward for latency. But if we try to fetch data from a metrics server or 3rd party services to generate the score, the returned output varies and we might need a way to define how to parse the result.

mvazquezc: Hey @qiujian16,

I think that for latencies global object or per-cluster object would both work. In terms of data,  I was thinking of an unstructured structure where people developing custom controllers with the addon framework will consume the data the way they need it. I don't think this should be a controller taking care of multiple data sources, instead a single controller taking care of a data source, so if anything were wrong with the data itself it wouldn't affect the other add-ons.

fenar: +1 on Mario's approach on a single controller taking care of a single data source type.

qiujian16 (commented):
put some comment, but I haven't thought through the whole idea yet. Will take some time this week and add more comments.

enhancements/sig-architecture/32-extensiblescheduling/score-generator/README.md
```diff
@@ -0,0 +1,192 @@
...
+look like this for the example above:
+
+~~~yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: AddOnPlacementScoreGenerator
+metadata:
+  name: cluster1-generator
+  namespace: cluster1
+spec:
+  addOnPlacementSelector:
```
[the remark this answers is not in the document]
mvazquezc: not really, it's the name of the `AddOnPlacementScore`. This selector would be used by the controller generating the `AddOnPlacementScore` to get the `AddOnPlacementScoreGenerator` that applies to this specific `AddOnPlacementScore`.

enhancements/sig-architecture/32-extensiblescheduling/score-generator/README.md
```diff
@@ -0,0 +1,192 @@
...
+Now, we want to add linux.com to the list of locations to test. With the current implementation we 
+will need to edit the code of the addon controller and include the test to linux.com + the result to be added
+to the `AddOnPlacementScore`.
+
+To fix above issue, the proposal is to create a new API `AddOnPlacementScoreGenerator`, which could 
+look like this for the example above:
+
+~~~yaml
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: AddOnPlacementScoreGenerator
```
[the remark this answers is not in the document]
mvazquezc: wfm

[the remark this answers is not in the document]
mvazquezc: To the comment below:

1. I don't think we want to create types... I was thinking of an unstructured data type where the controller consuming the API should take care of reading the fields it expects to be present in order to generate the required scores.

qiujian16 (commented):
I am still not convinced that this should be put in the API repo. The major uncertainty is how this API can be evolved and consumed by multiple score providers. It seems to me a new score provider has to update this API or build their own CRDs to config how score is generated. It does not seems change the fact that AddOnPlacementScore is the only integration point. But I am 100% ok if we create a separated repo and maintain this API and controller there.

mvazquezc:
> I am still not convinced that this should be put in the API repo. The major uncertainty is how this API can be evolved and consumed by multiple score providers. It seems to me a new score provider has to update this API or build their own CRDs to config how score is generated. It does not seems change the fact that AddOnPlacementScore is the only integration point. But I am 100% ok if we create a separated repo and maintain this API and controller there.

I'm not sure I follow you. The idea behind this CRD is to allow the customization of score generations by introducing a new API that can be consumed by the addon controllers, I don't think we want to have a dedicated controller for this API. 

Multiple score providers will be required to read this API in case they want to get score-generation customizations. In case these score providers don't need customizations they're fine not reading this CRD.

To me, this API would be like having your config in a ConfigMap and reading it upon controller start.

qiujian16:
I think I understand the intention, but I do not see how it can be achieved when a new provider comes without adding additional field in this API. For example, if I have a new provider comes and allow user to configure it to score the cluster based on resource usage, I have to update the API spec or introduce a new API to config. I do not see how you can define a common api schema that can adapt to different providers.

mvazquezc:
> I think I understand the intention, but I do not see how it can be achieved when a new provider comes without adding additional field in this API. For example, if I have a new provider comes and allow user to configure it to score the cluster based on resource usage, I have to update the API spec or introduce a new API to config. I do not see how you can define a common api schema that can adapt to different providers.

That's why I think it should be unstructured, so anyone can put their customization data without having to change the API. The controllers are responsible for reading their data and making sure that in case of misconfiguration, they can gracefully handle it.

qiujian16:
> That's why I think it should be unstructured, so anyone can put their customization data without having to change the API. The controllers are responsible for reading their data and making sure that in case of misconfiguration, they can gracefully handle it.
if it is unstructured, why we need to define an API schema instead of directly using a configmap :)? I think we need to at least  define some general concept in this API rather than arbitrary type...This is why I am struggling with how this API can provide commonality in this scenario.

mvazquezc:
> if it is unstructured, why we need to define an API schema instead of directly using a configmap :)? I think we need to at least define some general concept in this API rather than arbitrary type...This is why I am struggling with how this API can provide commonality in this scenario.

It could be done via ConfigMaps, we won't be able to leverage things defined in the KEP like `addOnPlacementSelector`  but I guess that's not that bad.

I cannot think of a common structured API that can be leveraged without having to update API types for every new provider. Requirements for different data providers may be very different and as such, hard to put in common.

PR #95, opened 2023-06-14
Update changes to placement strategy proposal

serngawy:


enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -301,35 +313,189 @@ status:
...
+    clusterCount: 20
+```
+
+The Placement Helper [library](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1beta1/helpers.go) should provide functions to facilitate retrieve the decision groups.
 
 #### Special case handling
 
 * Clusters have been removed from decision group index 1 and decision group index 2 has been created. In order to not shift current selected clusters in the decision groups, placement controller will keep append clusters to decision group index 2 till it reach max number of clusters per decision group then append new cluster to decision group index 1 till it reach the max number of clusters per decision group.
 
+* The PrioritizerPolicy has been defined by end user and a new cluster has been added to the placement selection after the decision Groups are created. The clusters wil need to be adjusted within the decision groups based on the new cluster decision group index.
```
haoqing0110, line 430: Could here provide an example of how the group and index for the clusters will adjust? 

My understanding whether a placement defines PrioritizerPolicy or not, the clusters in the placement decision are ordered by name rather than score (this is the current behavior to ensure when the cluster score changes but selected clusters keep the same, the placement decision won't change), so the special case here is mainly about how to handle the newly added clusters.

enhancements/sig-architecture/64-placementStrategy/README.md
```diff
@@ -301,35 +313,189 @@ status:
...
+	// GroupName of the decision group should match the placementDecisions label value with label key cluster.open-cluster-management.io/decision-group-name
+	// +optional
+	GroupName string `json:"groupName,omitempty"`
+
+	// GroupIndex of the decision group should match the placementDecisions label value with label key cluster.open-cluster-management.io/decision-group-index
+	// +optional
+	GroupIndex int32 `json:"groupIndex,omitempty"`
+}
+
+type RolloutStrategy struct {
```
haoqing0110, line 460: How about 
```
type RolloutStrategy struct {
	// +kubebuilder:default:=All
	// +optional
	Type RolloutStrategyType `json:"type,omitempty"`
	
	// +optional
	All *RolloutAll `json:”all,omitempty"`

	// +optional
	Progressive *RolloutProgressive `json:”progressive,omitempty"`

	// +optional
	ProgressivePerGroup *RolloutProgressivePerGroup `json:”progressivePerGroup,omitempty"`
}

type RolloutAll struct {
// list the supported field for each type
// ...
}

type RolloutProgressive struct {
// ...
}

type RolloutProgressivePerGroup struct {
// ...
}

File: enhancements/sig-architecture/64-placementStrategy/README.md (hunk 1 above)
Comment (serngawy) [reply enhancements/sig-architecture/64-placementStrategy/README.md:430]: yes, it is about the newly added clusters. I'm not quit familiar with PrioritizerPolicy but mainly this section is to be sure at implementation we take care of this case and we don't break the behavior.

File: enhancements/sig-architecture/64-placementStrategy/README.md (hunk 2 above)
Comment (serngawy) [reply enhancements/sig-architecture/64-placementStrategy/README.md:460]: Well, the RolloutStrategyType defined as string type in order to give more flexibility for consumer APIs. For example a consumer APIs (controller) need a different rolloutType than All, Progressive and ProgressivePerGroup.  So it can define a new rolloutType in the consumer API and use it without affecting the commonly used rolloutStrategy.

**Review by qiujian16 — COMMENTED** [review]
/approve

File: enhancements/sig-architecture/64-placementStrategy/README.md [hunk 3]
```diff
@@ -52,53 +92,24 @@ type PlacementSpec struct {
...
-}
-
 ```
 
 Adding DecisionGroupStatus to the PlacementStatus API 
+
 ```go
 // Present decision groups status based on the DecisionStrategy definition.
 type DecisionGroupStatus struct {
 	// Present the decision group index. If there is no decision strategy defined all placement decisions will be in group index 0
```
haoqing0110, line 102: Today, the placement decision starts from index1, for example, placement `demo` will generate placementdecision with name `demo-decision-1`. My question is should the DecisionGroupIndex also starts from 1 to be consistent.

serngawy: Not a big deal, but better to start with index 0 cause the groups count and index start with 1 might be miss leading.

haoqing0110: Yes, after consideration, the group starting from 0 makes sense. Placement decision name is just for readability, don't need to be considered together.

enhancements/sig-architecture/64-placementStrategy/README.md
haoqing0110: For the added or deleted cluster, I think a better way is always to recalculate the groups after a change, to ensure 2 placements (1 placement created before cluster change, another placement created after cluster change) have the same result.  
As for the consumer (addon, policy, etc), they could record the groups and determine how to handle the added/deleted clusters.

PR #96, opened 2023-07-05
update addon template description for health probe and custom sign ca

zhujian7:


enhancements/sig-architecture/82-addon-template/README.md
```diff
@@ -36,8 +36,8 @@ straightforward addon development way.
 
 - The addon agent image is still required.
 - The hosted mode addon is not supported.
-- The pre-delete-hook feature is not supported.
 - For complicated addon which has special RBAC requirement etc would still need to use addon-framework.
+- For the agent runtime, only support Deployment, not support Daemonset, Statefulset, etc.
```
qiujian16, line 40: I think you support that, but cannot return status? we should clarify this. A user can set sts/deploy in the same template, but addon status only reflect status of deploy right now.

zhujian7: but we only inject volumes that can access the hub cluster into the deployment.

zhujian7: patl

qiujian16 (commented):
/approve
/lgtm

PR #97, opened 2023-07-18
add KEP for event based manifestwork.

morvencao:
ref: https://github.com/open-cluster-management-io/ocm/issues/224

### DIFF
--- a/enhancements/sig-architecture/224-event-based-manifestwork/README.md
+++ b/enhancements/sig-architecture/224-event-based-manifestwork/README.md
@@ -117,6 +117,8 @@ sequenceDiagram
     // deletionTimestamp is an optional timestamp property representing the manifests of this message are deleting from the source, the work-agent needs to clean up the manifests from its cluster. It is represented in RFC3339 form and is in UTC. If not set, the agent will deploy or update the manifests on its cluster.
     // If present, MUST adhere to the format specified in RFC 3339.
     "deletionTimestamp": "<timestamp-of-the-resource-deletion>",
+    // clusterName indecates the resource will be deployed to which clsuter
+    "clusterName": "<cluster-name>"
 }
 ```
 
@@ -497,14 +499,16 @@ sequenceDiagram
 2. Extensions
 ```JSON5
 {
-    // specResourceID is from identifier of the corresponding spec message.
+    // ResourceID is from identifier of the corresponding spec message.
     // It is represented in string.
     // It is a required property.
-    "specResourceID": "<uuid-of-the-resource>",
-    // specResouceVersion is resource version of the corresponding spec message.
+    "ResourceID": "<uuid-of-the-resource>",
+    // ResouceVersion is resource version of the corresponding spec message.
     // It is represented in int64.
     // It is a required property.
-    "specResouceVersion": "<resource-version-in-int64>",
+    "ResouceVersion": "<resource-version-in-int64>",
+    // clusterName indecates the resource status from which clsuter
+    "clusterName": "<cluster-name>"
 }
 ```
 
@@ -886,8 +890,7 @@ cloud events.
 type Interface[T Object] interface {
 	Resync(ctx context.Context, eventType CloudEventType) error
 	Publish(ctx context.Context, eventType CloudEventType, obj T) error
-	Subscribe(ctx context.Context) error
-	SubscriptionResultChan() <-chan Event[T]
+	Subscribe(ctx context.Context, handlers ...ResourceHandler[T]) error
 }
 ```
 
@@ -903,17 +906,12 @@ type Object interface {
 }
 ```
 
-and the object should have the encoder/decoder to encode/decode the object/cloudevent to cloudevent/object
+and the object should have its Codec to encode the object to cloud event and decode the cloud event to the object.
 
 ```golang
-type Encoder[T Object] interface {
-	EncodeSpec(eventType, source string, obj T) (*cloudevents.Event, error)
-	EncodeStatus(eventType, source string, obj T) (*cloudevents.Event, error)
-}
-
-type Decoder[T Object] interface {
-	DecodeSpec(*cloudevents.Event) (T, error)
-	DecodeStatus(*cloudevents.Event) (T, error)
+type Codec[T Object] interface {
+	Encode(eventType, source string, obj T) (*cloudevents.Event, error)
+  Decode(*cloudevents.Event) (T, error)
 }
 ```
 
@@ -929,7 +927,7 @@ at last, we need lister to list the objects to resync the objects between source
 
 ```golang
 type Lister[T ManifestObject] interface {
-	List() []T
+	List(source, clusterName string) ([]T, error)
 }
 ```
 
@@ -945,16 +943,8 @@ mwCloudEventClient := NewCloudEventClient[*workv1.ManifestWork](
 		)
 
 // add multiple encoders/decoders to support different cloud event payload
-mwCloudEventClient.AddPayloadCoders(
-  {Group: "io.open-cluster-management.works", Version: "v1alpha1", Resource: "manifestbundle"},
-  manifestbundleEncoder,
-  manifestbundleDecoder,
-)
-mwCloudEventClient.AddPayloadCoders(
-  {Group: "io.open-cluster-management.works", Version: "v1alpha1", Resource: "manifest"},
-  manifestEncoder,
-  manifestDecoder,
-)
+mwCloudEventClient.AddCodec("io.open-cluster-management.works.v1alpha1.manifestbundle", manifestbundleCodec)
+mwCloudEventClient.AddCodec("io.open-cluster-management.works.v1alpha1.manifest", manifestCodec)
 ```
 
 By default, we provide a work client builder for devlopers, devlopers can enable the cloud event based on the MQTT
--- a/enhancements/sig-architecture/224-event-based-manifestwork/README.md
+++ b/enhancements/sig-architecture/224-event-based-manifestwork/README.md
@@ -25,7 +25,7 @@ To facilitate easy adoption of the event-based mechanism, a client interface sho
 ### Terminology
 
 - Source: The API consumer or component that utilizes this specification for ManifestWorks. It can be a controller on the hub cluster or a RESTful service handling resource requests. For RESTful service source, it often has a database to store the requested resources. Source should have a unique identifier to ensure its uniqueness and consistency, for example, a hub controller can generate a source ID by hashing the hub cluster URL and appending the controller name. Similarly, a RESTful service can select a unique name or generate a unique ID in the associated database for its source identification.
-- Work Agent: The controller that handles the deployment of requested resources on the managed cluster and status report to the source. Work agent should have a unique identifier to ensure its uniqueness and consistency. Usually, work agent is uniquely identified by a string that combines mangaed cluster name and agent name for consistency.
+- Work Agent: The controller that handles the deployment of requested resources on the managed cluster and status report to the source. Work agent should have a unique identifier to ensure its uniqueness and consistency. Usually, work agent is uniquely identified by a string that combines managed cluster name and agent name for consistency.
 - Broker: The message broker that handles the event transfer between the source and the work agent. It is represented by a message broker, such as MQTT or Kafka.
 
 ### Goals
@@ -86,7 +86,7 @@ sequenceDiagram
     // The version of the CloudEvents specification which the event uses.
     // Compliant event producers MUST use a value of "1.0" when referring to this version of the specification.
     "specversion": "1.0",
-    // Unique indentifier for cloud events message.
+    // Unique identifier for cloud events message.
     "id": "<cloud-events-message-id>",
     // <source-id> should be a unique identifier for the source to identify the context in which an event occurred.
     // For example, a hub controller can generate a source ID by hashing the hub cluster URL and appending a controller name,
@@ -110,15 +110,15 @@ sequenceDiagram
 2. Extensions
 ```JSON5
 {
-    // resourceID is a required string property representing the unique identifier for this manifestwork. It is the responsibility of the source to ensure its uniqueness and consistency. MUST adhere to the format specified in RFC 4122.
-    "resourceID": "<uuid-of-the-resource>", 
-    // resourceVersion is a required int64 sequence number property that must be incremented by the source whenever the data of this message changes. It is the responsibility of the source to guarantee its incremental nature.
-    "resourceVersion": "<resource-version-in-int64>",
-    // deletionTimestamp is an optional timestamp property representing the manifests of this message are deleting from the source, the work-agent needs to clean up the manifests from its cluster. It is represented in RFC3339 form and is in UTC. If not set, the agent will deploy or update the manifests on its cluster.
+    // resourceid is a required string property representing the unique identifier for this manifestwork. It is the responsibility of the source to ensure its uniqueness and consistency. MUST adhere to the format specified in RFC 4122.
+    "resourceid": "<uuid-of-the-resource>", 
+    // resourceversion is a required int64 sequence number property that must be incremented by the source whenever the data of this message changes. It is the responsibility of the source to guarantee its incremental nature.
+    "resourceversion": "<resource-version-in-int64>",
+    // deletiontimestamp is an optional timestamp property representing the manifests of this message are deleting from the source, the work-agent needs to clean up the manifests from its cluster. It is represented in RFC3339 form and is in UTC. If not set, the agent will deploy or update the manifests on its cluster.
     // If present, MUST adhere to the format specified in RFC 3339.
-    "deletionTimestamp": "<timestamp-of-the-resource-deletion>",
-    // clusterName indecates the resource will be deployed to which clsuter
-    "clusterName": "<cluster-name>"
+    "deletiontimestamp": "<timestamp-of-the-resource-deletion>",
+    // clustername indicates the resource will be deployed to which cluster
+    "clustername": "<cluster-name>"
 }
 ```
 
@@ -136,7 +136,7 @@ The payload of the spec event varies based on the resource name. Two supported r
 }
 ```
 
-- **manbifestbundle (for multiple resources)**
+- **manifestbundle (for multiple resources)**
 
 ```JSON5
 {
@@ -460,7 +460,7 @@ The work agent communicates the state of the resources it manages to the respect
 ```mermaid
 sequenceDiagram
     Source -->> Broker: Subscribe topic - /sources/<source-id>/clusters/+/status
-    WorkAgent -->> WorkAgent: Reconcile Manifestwork and encode manifestwork status into status event
+    WorkAgent -->> WorkAgent: Reconcile manifestwork and encode manifestwork status into status event
     WorkAgent -->> Broker: Publish status event to topic - /sources/<source-id>/clusters/<cluster-name>>/status
     Broker -->> Source: Send status event to subscribers
     Source -->> Source: Decode status event and retrieve manifestwork status
@@ -475,7 +475,7 @@ sequenceDiagram
     // The version of the CloudEvents specification which the event uses.
     // Compliant event producers MUST use a value of "1.0" when referring to this version of the specification.
     "specversion": "1.0",
-    // Unique indentifier for cloud events message.
+    // Unique identifier for cloud events message.
     "id": "<cloud-events-message-id>",
     // <source-id> should be a unique identifier for the source to identify the context in which an event occurred.
     // For example, a hub controller can generate a source ID by hashing the hub cluster URL and appending a controller name,
@@ -499,16 +499,16 @@ sequenceDiagram
 2. Extensions
 ```JSON5
 {
-    // ResourceID is from identifier of the corresponding spec message.
+    // resourceid is from identifier of the corresponding spec message.
     // It is represented in string.
     // It is a required property.
-    "ResourceID": "<uuid-of-the-resource>",
-    // ResouceVersion is resource version of the corresponding spec message.
+    "resourceid": "<uuid-of-the-resource>",
+    // resouceversion is resource version of the corresponding spec message.
     // It is represented in int64.
     // It is a required property.
-    "ResouceVersion": "<resource-version-in-int64>",
-    // clusterName indecates the resource status from which clsuter
-    "clusterName": "<cluster-name>"
+    "resouceversion": "<resource-version-in-int64>",
+    // clustername indicates the resource status from which cluster
+    "clustername": "<cluster-name>"
 }
 ```
 
@@ -587,8 +587,8 @@ In this example, we have a hub controller called "mwrs-hub-controller" serving a
 }
 // Extensions
 {
-  "specResourceid": "a52adbe8-b6f2-52c8-9378-c4f544502fb7",
-  "specResourceversion": 1
+  "resourceid": "a52adbe8-b6f2-52c8-9378-c4f544502fb7",
+  "resourceversion": 1
 }
 // Data
 {
@@ -652,8 +652,8 @@ In this example, we have a hub controller called "mwrs-hub-controller" serving a
 }
 // Extensions
 {
-  "specResourceid": "a52adbe8-b6f2-52c8-9378-c4f544502fb7",
-  "specResourceversion": 1
+  "resourceid": "a52adbe8-b6f2-52c8-9378-c4f544502fb7",
+  "resourceversion": 1
 }
 // Data
 {
@@ -748,7 +748,7 @@ sequenceDiagram
     WorkAgent -->> WorkAgent: Generate resource version map and encode as spec resync event
     WorkAgent -->> Broker: Publish spec resync event to topic - /sources/clusters/<cluster-name>/specresync
     Broker -->> Source: Send spec resync event to subscribers
-    Source -->> Source: Compare the resource versions in event message and local soource and generate spec event
+    Source -->> Source: Compare the resource versions in event message and local source and generate spec event
     Source -->> Broker: Publish spec event to topic - /sources/<source-id>/clusters/<cluster-name>/spec
     Broker -->> WorkAgent: Send spec event to subscribers
     WorkAgent -->> WorkAgent: Decode spec event to manifestwork and reconcile it
@@ -762,7 +762,7 @@ Event schema:
     // The version of the CloudEvents specification which the event uses.
     // Compliant event producers MUST use a value of "1.0" when referring to this version of the specification.
     "specversion": "1.0",
-    // Unique indentifier for cloud events message.
+    // Unique identifier for cloud events message.
     "id": "<cloud-events-message-id>",
     // <source-id> should be a unique identifier for the source to identify the context in which an event occurred.
     // For example, a hub controller can generate a source ID by hashing the hub cluster URL and appending a controller name,
@@ -831,7 +831,7 @@ sequenceDiagram
     // The version of the CloudEvents specification which the event uses.
     // Compliant event producers MUST use a value of "1.0" when referring to this version of the specification.
     "specversion": "1.0",
-    // Unique indentifier for cloud events message.
+    // Unique identifier for cloud events message.
     "id": "<cloud-events-message-id>",
     // <source-id> should be a unique identifier for the source to identify the context in which an event occurred.
     // For example, a hub controller can generate a source ID by hashing the hub cluster URL and appending a controller name,
@@ -894,7 +894,7 @@ type Interface[T Object] interface {
 }
 ```
 
-The resouce object should implement the following interface, so that we can get the `resourceID` with `GetUID` function
+The resource object should implement the following interface, so that we can get the `resourceID` with `GetUID` function
 and get the `resourceVersion` with `GetResourceVersion` function.
 
 ```golang
@@ -931,7 +931,7 @@ type Lister[T ManifestObject] interface {
 }
 ```
 
-Devlopers can use `NewCloudEventClient[T Object]` function to create the client for their resources, for example, we can
+Developers can use `NewCloudEventClient[T Object]` function to create the client for their resources, for example, we can
 easily to build a client for `ManifestWork`
 
 ```golang
@@ -947,7 +947,7 @@ mwCloudEventClient.AddCodec("io.open-cluster-management.works.v1alpha1.manifestb
 mwCloudEventClient.AddCodec("io.open-cluster-management.works.v1alpha1.manifest", manifestCodec)
 ```
 
-By default, we provide a work client builder for devlopers, devlopers can enable the cloud event based on the MQTT
+By default, we provide a work client builder for developers, developers can enable the cloud event based on the MQTT
 for `ManifestWork` with MQTT options, the builder returns a work client holder, we can get the work client set from this
 holder, e.g.
 
--- a/enhancements/sig-architecture/224-event-based-manifestwork/README.md
+++ b/enhancements/sig-architecture/224-event-based-manifestwork/README.md
@@ -10,7 +10,7 @@
 
 ## Summary
 
-This proposal optimizes scaling and performance by treating manifestworks as event messages, reducing excessive storage on the kube-apiserver of hub cluster. This allows managed cluster agents to subscribe to relevant manifestworks, eliminating the need for extensive hub storage. The proposed changes also enable diverse architectures to consume ManifestWorks, while providing a client interface for seamless adoption of the event-based mechanism.
+This proposal optimizes scaling and performance by treating manifestworks as event messages, reducing excessive storage on the kube-apiserver of hub cluster. This allows managed cluster agents to subscribe to relevant manifestworks, reducing the data being stored in ETCD of hub cluster. The proposed changes also enable diverse architectures to consume ManifestWorks, while providing a client interface for seamless adoption of the event-based mechanism.
 
 ## Motivation
 
--- a/enhancements/sig-architecture/224-event-based-manifestwork/README.md
+++ b/enhancements/sig-architecture/224-event-based-manifestwork/README.md
@@ -83,13 +83,17 @@ Event Schema:
     "specversion": "1.0",
     // Unique indentifier for cloud events message.
     "id": "<cloud-events-message-id>",
-    // Use "/sources/<source-id>/clusters/<cluster-id>/<resource-type>" as the cloud events source to identify the context in which an event occurred.
-    // <source-id> should be a unique identifier for the source. For instance, a hub controller can generate a source ID by hashing the hub cluster URL and appending a controller name.
-    // <cluster-id> represents the identifier of the target cluster to which this cloud event is sent.
-    // <resource-type> refers to the resource included in the cloud event payload.
-    "source": "/sources/<source-id>/clusters/<cluster-id>/<resource-type>",
-    // Use "<reverse-domain-name-of-resource-organization>.<resource-version>.<resource-name>.<subsource>.<action>" as the format for this attribute, which contains a value describing the type of event related to the originating occurrence.
-    // For example, "io.open-cluster-management.v1.manifestwork.spec.created" can be used as a sample format.
+    // Use "<source-id>" as the cloud events source to identify the context in which an event occurred.
+    // <source-id> should be a unique identifier for the source.
+    // For example, a hub controller can generate a source ID by hashing the hub cluster URL and appending a controller name,
+    // a work agent can generate a source ID with <hub-id>/<agent-id>.
+    "source": "<source-id>",
+    // Use "<reverse-group-of-resource>.<resource-version>.<resource-name>.<subresource>.<action>" as the format for this attribute, which contains a value describing the type of event data.
+    // The "<reverse-group-of-resource>.<resource-version>.<resource-name>" identifies the type of the resource included in the event payload.
+    // The valid value of "<subresource>" includes: spec or status
+    // The "<action>" represents the action for this resource, e.g. create_request, update_request and delete_request and resync_request.
+    // For example, "io.open-cluster-management.works.v1alpha1.manifestbundle.spec.create_request", "io.open-cluster-management.works.v1alpha1.manifestbundle.status.update_request",
+    // "io.open-cluster-management.works.v1alpha1.manifest.spec.create_request", etc.
     "type": "<reverse-domain-name-of-resource-organization>.<resource-version>.<resource-name>.<subsource>.<action>",
     // Timestamp of when the occurrence happened. Optional, if present, MUST adhere to the format specified in RFC 3339.
     "time": "<timestamp-of-the-event-occurrence>",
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+    "type": "io.open-cluster-management.v1.manifestwork.spec.created", //This attribute contains a value describing the type of event related to the originating occurrence. Often this attribute is used for routing, observability, policy enforcement, etc.
+    "time": "2023-04-05T17:31:00Z", // Timestamp of when the occurrence happened.
+    "datacontenttype": "application/json",
+    // resourceID is the unique id for this manifestwork.
+    // It is represented in string.
+    // It is a required property.
+    "resourceID": "748d04ad-887a-44dd-b788-da1373227072", 
+    // resourceVersion is a sequence number that grows as the data of this message changes.
+    // It is represented in int64.
+    // It is a required property.
```
qiujian16, line 90: let's clarify that it is source's responsibility to ensure it is incremental

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+{
+    "specversion": "1.0", //The version of the CloudEvents specification which the event uses.
+    "id": "<cloud-events-message-id>",
+    "source": "<source-id>", //Identifies the context in which an event happened. Often this will include information such as the type of the event source, the organization publishing the event or the process that produced the event.
+    "type": "io.open-cluster-management.v1.manifestwork.spec.created", //This attribute contains a value describing the type of event related to the originating occurrence. Often this attribute is used for routing, observability, policy enforcement, etc.
+    "time": "2023-04-05T17:31:00Z", // Timestamp of when the occurrence happened.
+    "datacontenttype": "application/json",
+    // resourceID is the unique id for this manifestwork.
+    // It is represented in string.
+    // It is a required property.
```
qiujian16, line 86: let's clarify that it is source's responsibility to ensure it is unique

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+```json
+{
+    "specversion": "1.0",
+    "id": "<cloud-events-message-id>",
+    "source": "<source-id>",
+    "type": "io.open-cluster-management.v1.manifestwork.specresync.requested”",
+    "datacontenttype": "application/json",
+    "time": "2018-04-05T17:31:00Z",
+    "data": {
+       resourceVersions: [
```
qiujian16, line 263: format error?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+    Source -->> Broker: Subscribe topic - /sources/<source-id>/clusters/+/manifestsstatus
+    Source -->> Source: Generate status hash map and encode as status resync event
+    Source -->> Broker: Publish status resync event
+    Broker -->> WorkAgent: Send status resync event
+    WorkAgent -->> WorkAgent: Compare the status hashes in event message and agent and generate status event
+    WorkAgent -->> Broker: Publish status event for manifestwork
+    Broker -->> Source: Send status event for manifestwork
+    Source -->> Source: Decode status event and retrieve manifestwork status
+```
+
```
qiujian16, line 306: should add a section on failure handling.

Also what is the impact of message resync?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+}
+```
+
+### Status Event
+
+The work agent communicates the state of the resources it manages to the respective source using this message.
+
+Event transfer flow:
+
+1. The source subscribes to the status event topic `/sources/<source-id>/clusters/+/manifestsstatus` from the broker, for example: `/sources/sd3ded4-mwrs-controller/clusters/+/manifestsstatus`.
```
qiujian16, line 159: we need to define what source means

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+    Source -->> Broker: Publish spec event for manifestwork
+    Broker -->> WorkAgent: Send spec event for manifestwork
+    WorkAgent -->> WorkAgent: Decode spec event and apply manifestwork
+```
+
+Event Schema:
+
+```json
+{
+    "specversion": "1.0", //The version of the CloudEvents specification which the event uses.
```
skeeey, line 78: the json cannot support use `//` as comment, we may consider use a schema here, or use golang object to define the event

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+                // conditions represents the conditions of this resource on a managed cluster.
+                // The condition format is same with the Kuberenetes
+                // condition (k8s.io/apimachinery/pkg/apis/meta/v1) 
+                "conditions": []
+            }]
+        }
+    }
+}
+```
+
```
skeeey, line 219: do we need an example here?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+    WorkAgent -->> WorkAgent: Decode spec event and apply manifestwork
+```
+
+Event Schema:
+
+```json
+{
+    "specversion": "1.0", //The version of the CloudEvents specification which the event uses.
+    "id": "<cloud-events-message-id>",
+    "source": "<source-id>", //Identifies the context in which an event happened. Often this will include information such as the type of the event source, the organization publishing the event or the process that produced the event.
```
skeeey, line 80: let's have an exact definition for source filed, I think it should be message producer topic, e.g.
`/sources/sd3ded4-mwrs-hub-controller/clusters/cluster1/manifests`
`/sources/sd3ded4-mwrs-agent-controller/clusters/cluster1/manifestsstatus`

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+    "type": "io.open-cluster-management.v1.manifestwork.spec.created", //This attribute contains a value describing the type of event related to the originating occurrence. Often this attribute is used for routing, observability, policy enforcement, etc.
```
skeeey, line 81: I think we need clarify it is source's responsibility to define this, I think different resource can have its own event type, e.g.
for manifestwork, it may have
```
io.open-cluster-management.v1.manifestwork.create_request
io.open-cluster-management.v1.manifestwork.delete_request
io.open-cluster-management.v1.manifestwork.update_request
io.open-cluster-management.v1.manifestwork.status.update_request
```
and for other resource, it may have
```
io.example.v1.myresouce.create_request
io.example.v1.myresouce.delete_request
io.example.v1.myresouce.update_request
io.example.v1.myresouce.status.update_request
```

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+To ensure uniqueness and immutability of source components, each source, whether it is a consumer on the open cluster management hub cluster or a restful service handling resource requests, should possess a unique identifier. For instance, the open cluster management hub consumer can generate a source ID by hashing the hub cluster URL and appending a consumer name. Similarly, a restful service can choose a unique name or generate a unique ID in the associated database for its source identification.
+
+### Spec Event
+
+Sources send their resources along with the message to the work agent, enabling the creation, updating, or deletion of the respective resources on the managed cluster.
+
+Event transfer flow:
+
+1. The work agent subscribes to the spec event topic `/sources/+/clusters/<cluster-id>/manifests` from the broker, for example: `/sources/+/clusters/cluster1/manifests`.
+2. The source receives high level API request and publishes its resources using the spec event to the topic `/sources/<source-id>/clusters/<cluster-id>/manifests`, for example: `/sources/sd3ded4-mwrs-controller/clusters/cluster1/manifests`.
```
skeeey, line 62: nit: use /sources/sd3ded4-mwrs-hub-controller/clusters/cluster1/manifests` instead of `/sources/sd3ded4-mwrs-controller/clusters/cluster1/manifests`

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,360 @@
...
+```
+
+### Status Event
+
+The work agent communicates the state of the resources it manages to the respective source using this message.
+
+Event transfer flow:
+
+1. The source subscribes to the status event topic `/sources/<source-id>/clusters/+/manifestsstatus` from the broker, for example: `/sources/sd3ded4-mwrs-controller/clusters/+/manifestsstatus`.
+2. The work agent publishes the status of the resources using the status event to topic `/sources/<source-id>/clusters/<cluster-id>/manifestsstatus`, for example: `/sources/sd3ded4-mwrs-controller/clusters/cluster1/manifestsstatus`.
```
skeeey, line 160: nit: use `/sources/sd3ded4-mwrs-agent-controller/clusters/cluster1/manifestsstatus` instead of `/sources/sd3ded4-mwrs-controller/clusters/cluster1/manifestsstatus`

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: done.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: done.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: missed double quotes for the field name, fixed.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: We have description of **source** under **Design Details** section, do we still need it here?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: It looks like the github mardown issue about json, fixed it by the solution from https://gist.github.com/DamianEdwards/31d2457737304ca73556

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: done

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: done.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: done

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: added examples.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: done

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: Added failure section(mainly about resync) in https://github.com/open-cluster-management-io/enhancements/pull/97/commits/dccf37032d8d9df6b9434ac1e7761a0dd37405c4

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,674 @@
...
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning),
+or by redefining what graduation means.
+
+In general, we try to use the same stages (alpha, beta, stable), regardless how the functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
```
clyang82, line 655: Can we add a section to describe what's changed for other components in order to use the event based framework? for example: if the policy will use this component, Policy-agent does not need to report status to hub cluster namespace. Work-agent to report the policy status. Policy-propagator controller aggregate the status based on status cloudevent. there is no need to have replicatedPolicy. we won't have cluster namespace in hub side. etc.

morvencao: sure, will add that.

clyang82: maybe add another story?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,674 @@
...
+Event transfer flow:
+
+1. The source subscribes to the spec resync event topic `/sources/resync/+/manifests` from the broker to receive spec resync request.
+2. The work agent generates a spec resync event based on its current maintained resources.
+3. The work agent sends the spec resync event to the broker with the topic `/sources/resync/<cluster-id>/manifests`, for example: `/sources/resync/cluster1/manifests`.
+4. Upon receiving the spec resync event, the source processes the event message and interacts with the broker as follows:
+  - If the request event message is empty, the source returns all resources associated with the work agent.
+  - If the request event message contains resource IDs and versions, the source retrieves the resource with the specified ID and compares the versions.
+  - If the requested resource version matches the source's current maintained resource version, the source does not resend the resource.
```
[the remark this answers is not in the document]
skeeey: yes, it should be

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,674 @@
...
+
+## Motivation
+
+Manifestworks are used to deliver manifests from a hub cluster to managed clusters. As manifestwork adoption and managed cluster numbers increase, we encounter scaling issues due to excessive manifestwork creation on the hub. This adversely affects the performance of the hub apiserver. This proposal is to explore treating manifestworks as event messages instead of storing them in etcd.
+
+By considering each cluster namespace as a separate topic or channel, managed cluster agents can subscribe to relevant manifestworks. This approach eliminates the need to store manifestworks on the hub cluster, as they are primarily consumed by controllers like the addon manager or manifestworkreplicaset controller. Additionally, this change allows other architectures, such as non-Kubernetes restful services or databases, to consume ManifestWorks.
+
+To facilitate easy adoption of the event-based mechanism, a client interface should be provided, enabling existing work agents or consumers to transition without extensive code changes.
+
```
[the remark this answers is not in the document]
morvencao: done.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: not sure about this, but policy may not a good example for manifestwork, they have their own transport/propagation logic,  not using manifestwork.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal optimizes scaling and performance by treating manifestworks as event messages, reducing excessive storage on the hub cluster. This allows managed cluster agents to subscribe to relevant manifestworks, eliminating the need for extensive hub storage. The proposed changes also enable diverse architectures to consume ManifestWorks, while providing a client interface for seamless adoption of the event-based mechanism.
+
```
bjoydeep, line 14: You know this - we really do not save `hub storage`. We divert the storage from apiserver/etcd to other horizontally scalable systems. They still need to be stored. Should we clarify that.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+### Goals
+
+- Establish a standardized event protocol for delivering manifestworks and returning status information.
+- Define a resync mechanism to handle failures and ensure synchronization.
+- Develop a client interface that facilitates easy integration for consumers.
+
+### Non-Goals
+
+- Define the controlplane to act as a consumer of the protocol.
+
```
bjoydeep, line 34: There are several other constraints we are imposing. These need to be clarified because they are not the only way of doing this
Are we messaging framework agnostic or are we talking about working with MQTT Broker.
We need a topic per managed cluster specifically for manifestwork (there could be other topics for the same cluster)

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+### Spec Event
+
+Sources send their resources along with the message to the work agent, enabling the creation, updating, or deletion of the respective resources on the managed cluster. "Source" is used to refer to the API consumers or components that utilize the CloudEvents specification for ManifestWorks.
+
+Event transfer flow:
+
+1. The work agent subscribes to the spec event topic `/sources/+/clusters/<cluster-id>/manifests` from the broker, for example: `/sources/+/clusters/cluster1/manifests`.
+2. The source receives high level API request and publishes its resources using the spec event to the topic `/sources/<source-id>/clusters/<cluster-id>/manifests`, for example: `/sources/sdbdeavo-mwrs-hub-controller/clusters/cluster1/manifests`.
```
bjoydeep, line 64: These examples are supportable in MQTT Broker and not kafka. And IMHO I think we align with MQTT for optimization. It will be much harder to optimize for both at the same time.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: Where and how is source going to store the result of `Decode status event and retrieve manifestwork status` . We need to make that clear.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: It is self defeating to have policy propagate on its own. We end up increasing the agent footprint on the cluster. Policy propagation has the same constraint as of manifestwork

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+
+- Resync Mechanism:
+
+To ensure the integrity of the event-based manifestworks system during source or work agent restart or reconnection, the following optimized resync mechanism is employed:
+
+1. On start, the source subscribes to the spec resync event topic /sources/resync/+/manifests from the broker.
+2. On start, the work agent subscribes to the status resync event topic /sources/+/resync/clusters/manifestsstatus from the broker.
+3. Upon work agent restart or reconnect, the work agent generates a spec resync event with resource versions and sends it to the broker.
+4. Upon source restart or reconnect, the source calculates a status hash of its owned resources and sends a status resync event with the resource hash to the broker.
+5. The source compares the resource versions in the spec resync event message with the hub's state and generates spec events to the broker for the work agent to receive missing spec events.
```
bjoydeep, line 628: As said above, we need to be clear on where/how is hub storing this state. This should not be Kube API

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+~~The retry mechanism should employ exponential backoff, gradually increasing the delay between retries to prevent overwhelming the system with failed events. Each retry attempt should be logged, including relevant information such as the event payload, timestamp, and error details.~~
+
+- Resync Mechanism:
+
+To ensure the integrity of the event-based manifestworks system during source or work agent restart or reconnection, the following optimized resync mechanism is employed:
+
+1. On start, the source subscribes to the spec resync event topic /sources/resync/+/manifests from the broker.
+2. On start, the work agent subscribes to the status resync event topic /sources/+/resync/clusters/manifestsstatus from the broker.
+3. Upon work agent restart or reconnect, the work agent generates a spec resync event with resource versions and sends it to the broker.
+4. Upon source restart or reconnect, the source calculates a status hash of its owned resources and sends a status resync event with the resource hash to the broker.
```
bjoydeep, line 627: AFAIK, MQTT broker has the framework to handle 1,2,3 & 4. We should not need any code to do it... beyond changing configurations in MQTT connection.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
clyang82: we do not care about where the data is stored. the proposal enables to get the message from event streaming for the work agent. As you said, the data can be on kube api server. it can be in another rest api.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
clyang82: the message needs to follow the cloudevents. we do not limit to MQTT only. it can be other transport layers. But we take MQTT as an example or initial implementation now.

maybe add storage and persistence requirement as Non-Goals. WDYT @morvencao

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+  "datacontenttype": "application/json",
+}
+// Extensions
+{
+  "resourceid": "a52adbe8-b6f2-52c8-9378-c4f544502fb7",
+  "resourceversion": 3,
+  "deletiontimestamp": "2023-07-19T03:05:46.841308779Z"
+}
+```
+
```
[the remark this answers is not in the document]
clyang82: are you talking about the message order? I suggest to use at least QoS 1 and as mentioned to wait for the PUBACK. refer to https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/

enhancements/sig-architecture/224-event-based-manifestwork/README.md
clyang82: `resync` need to handle by our code. MQTT can retain message, but it has limitation. that is why we need to handle `resync` by ourselves.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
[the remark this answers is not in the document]
morvencao: This depends on implementation? 
Agent should also have a cache for the received manifestwork, if the deleted manifestwork is not found in the cache, do nothing.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,788 @@
...
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal optimizes scaling and performance by treating manifestworks as event messages, reducing excessive storage on the hub cluster. This allows managed cluster agents to subscribe to relevant manifestworks, eliminating the need for extensive hub storage. The proposed changes also enable diverse architectures to consume ManifestWorks, while providing a client interface for seamless adoption of the event-based mechanism.
+
+## Motivation
+
+Manifestworks are used to deliver manifests from a hub cluster to managed clusters. As manifestwork adoption and managed cluster numbers increase, we encounter scaling issues due to excessive manifestwork creation on the hub. This adversely affects the performance of the hub apiserver. This proposal is to explore treating manifestworks as event messages instead of storing them in etcd.
+
```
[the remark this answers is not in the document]
skeeey: yes, we focus how to deliver manifests between the sources and agents in this enhancement, for persistent part, it should be the responsibility of the sources/agents, e.g. they can store the data in the db/etch/memory, we can add this to out of scope

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: This proposal talks about source. And source generates events to which the work agent reacts. And the first proposed User Story is related to source.
Now source, as you will agree cannot just send event w/o storing it. In pure kube world, it was getting stored in etcd. In this case where will it be stored. It should not fire an event w/o ever knowing what it fired. Or when the status feedback comes, it will have to store somewhere.
So, we should care about where it is stored. Am I missing something?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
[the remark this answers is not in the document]
skeeey: this depends on the work agent implement, for current implement, if we delete a nonexistent resource, the work agent will ignore it

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: We did not use specific features of the MQTT protocol to optimize, we use the cloud events sdk to do this, the sdk hides the protocol details for us,  in fact, we just need a transporter/broker, we can pub/sub message from this transporter,  regardless of the transporter is MQTT/Kafka/AMQ, etc, we may need to describe this part more generic.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: yes, we may leverage the MQTT session mechanism to relieve this problem, but it still cannot avoid this totally, e.g. when the agent connect the broker after its session is expiration, the agent will lost the message, and even if we set the session will never expire (this will occupy the more broker resources), if the broker is crash, the source/agent  is still possible to lose messages, so we still need the resync

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: We mainly cover two kinds of sources. one is hub controller, another is restful service. For restful service, it is usually backend by a database for storage. For hub controller, the resources are  often high-level resources above ManifestWork, for example, ManifestWorkReplicateSet or Policy(we will make some change to policy controller to use manifesowork API for policy propagation and status report). These high-level resources are abstract of ManifestWork API, will be stored in ETCD.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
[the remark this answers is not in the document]
morvencao: Added a new section about `Source` definition, and also updated the non-goals for the Source architecture, please take another look and give your comments.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: Added a new section about Source and Broker definition, also updated the non-goals for the Broker storage and persistence, please take another look and give your comments.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: For the manifestwork itself, we store it in cache. For high level API that abstract manifestwork API, they will be saved in DB or ETCD, depending on different kinds of Source arahitecture, but that's out of the scope of the proposal.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: `resync` is part of the communication protocol, while the transport configuration, is it part of the protocol? Should we let the user to enable message retain?

bjoydeep: @skeeey you hit the nail on the head. We should leverage MQTT session mechanism, set the session will never expire and assume that broker does not lose messages. And then see what happens before adding more code IMO. We should assume that broker should be resilient - as they are today. WDYT?
In Global Hub we added a powerful concept of sending delta message events and periodic complete events. That is not relevant here - looking at the current events being exchanged.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: Agreed. High level API in etcd. The manifestwork and state in cache can be a  little problematic IMO. But let it go for now as its outside scope.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: @morvencao how does source `calculates a status hash of its owned resources and sends a status resync event with the resource hash to the broker`  if its memory is empty on restart (assuming this is all stored in memory)

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: In fact there should be only one way in ocm to create objects in managed cluster. It should be through work agent.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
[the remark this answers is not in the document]
bjoydeep: Agreed on delete. I was not talking about message ordering - but thanks for pointing out the PUBACK @clyang82 . 
I guess the bigger question is - how can we scale this work agent by running multiple copies of it ? Refer: https://www.hivemq.com/article/mqtt-client-load-balancing-with-shared-subscriptions/

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: @skeeey can we look at the code a bit.  In the example that I see here: https://github.com/cloudevents/sdk-go/blob/main/samples/mqtt/receiver/main.go and from Kafka etc - the sdk wrapper is really minimal. We have access to all broker functions - as it should be. Each broker has different config settings and if we do not use them, it's a wastage. 
Having said that, if we want to make it generic, we should think through how would handle it in Kafka at least. For starter Kafka cannot handle `/` in its topics. So that has to change. The session retention in Kafka is different- this may or may not affect the design.  So @skeeey if we need to make it generic, we probably need to do more experimentation with prototypes.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: looks good. Thank you Longlong

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: @morvencao totally agreed. So do you think changing this -  `eliminating the need for extensive hub storage.` to `reducing the data being stored in ETCD of hub.` will work?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: If the source is restful service with DB, then the status can be from DB. If the source is hub controller, then emty hash slice will be sent to agent.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: yes, updated this. thanks @bjoydeep

bjoydeep: looks good to me

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: Are you saying resync operation does not do anything useful for hub controllers?

morvencao: no, for empty status hash, agent will send back all the status back to source.

qiujian16: I think resync can be optional if the broker implementation has other way to handle the drift. But we should still need resync as the last resort.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: @bjoydeep yeah, we will provide a client (it is described in the clients section) that help developers to finish the pub/sub/resync resources, https://github.com/skeeey/work/blob/mqtt/pkg/clients/mqclient/client.go#L33

With this client, developers provide their  own cloud event based clients (https://github.com/skeeey/work/blob/mqtt/pkg/clients/mqclient/client.go#L35), the clients may be from the MQTT, Kafka, etc. , like you mentioned

And considering different protocols may have their own different feature, we also provide the interface https://github.com/skeeey/work/blob/mqtt/pkg/clients/workclient/protocal/context.go to allow developers to build their cloud event context, e.g.
for MQTT, it may be
```
mqttSubClient, mqttPubClient, err := mqttOptions.GetCloudEventClients(ctx, b.clusterName)
mqttCTXGetter := mqttOptions.NewMQTTContextGetter()

ceClient := NewCloudEventClient[*workv1.ManifestWork](b.source,
			mqttSubClient,
			mqttPubClient,
// ....

// pub resources

ceClient.Publish(mqttCTXGetter.Context(...), work) error
``` 

for Kafka, it may be
```
kafkaSubClient, kafkaClient, err := kafkaOptions.GetCloudEventClients(...)
kafkaCTXGetter := kafkaOptions.NewMQTTContextGetter()

ceClient := NewCloudEventClient[*workv1.ManifestWork](b.source,
			kafkaSubClient,
			kafkaPubClient,
// ....

// pub resources

ceClient.Publish(kafkaCTXGetter.Context(...), work) error
```

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: @bjoydeep, yeah, I think we just provide a resync mechanism here (we do not make any assumptions about the broker), and we have a client to provide the pub/sub/resync functions, for deveployers/users, if they have a reliable
 broker, they can choose not to do resync between sources and agents

qiujian16: also we can not make an assumption on who stores the source of truth, it could be transport layer or it could be the consumer, resync ensures the agent can get the source of truth ultimately.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -497,14 +499,16 @@ sequenceDiagram
...
     // It is represented in string.
     // It is a required property.
-    "specResourceID": "<uuid-of-the-resource>",
-    // specResouceVersion is resource version of the corresponding spec message.
+    "ResourceID": "<uuid-of-the-resource>",
+    // ResouceVersion is resource version of the corresponding spec message.
     // It is represented in int64.
     // It is a required property.
-    "specResouceVersion": "<resource-version-in-int64>",
+    "ResouceVersion": "<resource-version-in-int64>",
```
morvencao: should be `resourceID` and `resourceVersion`? @skeeey

skeeey: yes, it should be, and should we use lower case for all attributes in cloud event?

morvencao:
Following our meeting discussion, I've inserted a note in the "Non-Goals" section to address message optimization. For instance, when utilizing MQTT, consider enabling the 'clean_session' flag to reduce message redundancy. Similarly, in the case of Kafka, optimizing the retention policy can effectively reduce unnecessary message proliferation. Moreover, the 'last will testament' feature in MQTT is indispensable for managing agent failures.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,1042 @@
...
+- Work Agent: The controller that handles the deployment of requested resources on the managed cluster and status report to the source. Work agent should have a unique identifier to ensure its uniqueness and consistency. Usually, work agent is uniquely identified by a string that combines managed cluster name and agent name for consistency.
+- Broker: The message broker that handles the event transfer between the source and the work agent. It is represented by a message broker, such as MQTT or Kafka.
+
+### Goals
+
+- Establish a standardized event protocol for delivering manifestworks and returning status information.
+- Define a resync mechanism to handle failures and ensure synchronization.
```
mprahl, line 34: In the case where the agent is disconnected from the broker and the agent restarts, what would be the expected behavior of the agent?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,1042 @@
...
+      }
+    ]}]
+}
+```
+
+### Spec Resync Event
+
+#### Event Scenarios
+
+1. After a work agent restarts, it sends a spec resync event with empty payload to the broker to synchronize all resources to avoid missing events during the restart period.
```
mprahl, line 727: Does the agent reapply manifests periodically or does it just apply it once after receiving the command from the broker?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,1042 @@
...
+5. The source compares the resource versions in the spec resync event message with the hub's state and generates spec events to the broker for the work agent to receive missing spec events.
+6. The work agent compares the status hashes in the status resync event message with its own state and generates status events to the broker for the source to receive missing status events.
+
+### Impact to Relevant Components
+
+Using cloud event based manifestwork can streamline resource delivery and status retrieval for various components managing resources. For instance, the policy controller can utilize this feature, eliminating the necessity for Policy agent to report status to the hub cluster namespace. Instead, Work-agent will handle reporting the policy status. The Policy propagator controller will aggregate the status through status report cloud event, obviating the need for replicatedPolicy. Additionally, there will be no cluster namespace in the hub side. This approach enhances efficiency and simplifies resource management across managed clusters.
```
mprahl, line 983: Is the idea that ManifestWork would deliver the replicated policy on the managed cluster side?

We need a local cache on the managed cluster side to account for when the cluster is disconnected from the Hub.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,1042 @@
...
+manifestWorkLister := workInformerFactory.Work().V1().ManifestWorks().Lister().ManifestWorks(o.SpokeClusterName)
+```
+
+### Failure Handling
+
+To ensure the integrity of the event-based manifestworks system during source or work agent restart or reconnection, the following optimized resync mechanism is employed:
+
+1. On start, the source subscribes to the spec resync event topic `/sources/clusters/+/specresync` from the broker.
+2. On start, the work agent subscribes to the status resync event topic `/sources/+/clusters/statusresync` from the broker.
+3. Upon work agent restarts or reconnects, the work agent generates a spec resync event with resource versions and sends it to the broker.
```
mprahl, line 976: How about status events that happened while the broker is offline or not reachable? Those would be lost?

I'm thinking in the policy case, we want to record every single compliance event which is why #98 proposed to have a local queue cache of compliance events until they can be recorded in the centralized place.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: the sources/agents need to send the resync request to sync the resource status/spec

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: yes, there is a cache in managed cluster side for the ManifestWorks from the broker.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: this depends on the version of the resource, the agent will compare the resource version, If the version of the resource received by the agent is greater than the version of the resource maintained by the agent, the agent will reapply the resource to update it

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: yes, I think we can deliver the replicated policy with the manifestwork

enhancements/sig-architecture/224-event-based-manifestwork/README.md
morvencao: This depends on message QoS?

skeeey: By default, we will set QoS to 1 to make sure the event is delivered to the broker, but if the broker is offline/unreachable, I think we need to have a retry for the agent to send the event when the broker becomes reachable again, so a local cache may be required for the policy case

morvencao:
@qiujian16 @bjoydeep Is there any concern regarding the PR merge?

qiujian16:
/approve

I do not have other concern now. leave this for 48h and we can merge it.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
mprahl: If the broker is still disconnected, what would the behavior be?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
mprahl: In general, if there is no change by the user, it will not constantly reapply it even if the cluster state has changed?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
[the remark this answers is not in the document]
bjoydeep: I think we are good here. Agent expects that it will have **`guaranteed & ordered message delivery`** to be honored by the system. Parallelization is a thig we can worry at a later time.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: Agent has a MQTT interface to get the CR. Thereafter, it should be business as usual. And then it has MQTT interface to relay back the status (or CR update)

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: It is source `side` responsibility to maintain shadow status of the agent. So if source specifically is disconnected from the broker, there should be other components on the source `side` which should maintain that state. I think we should try to avoid pushing the burden of this onto the agent. This is the role that https://github.com/kube-orchestra/maestro/ plays. WDYT

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: @mprahl MQTT Agent (or our agent) can be in a ship where it cannot connect to the MQTT Broker for an extended period. If for that agent, it is important to replay history, this has to stored by the agent. Not all agents will need this though. Does th make sense @skeeey @mprahl

enhancements/sig-architecture/224-event-based-manifestwork/README.md
```diff
@@ -0,0 +1,1042 @@
...
+- Work Agent: The controller that handles the deployment of requested resources on the managed cluster and status report to the source. Work agent should have a unique identifier to ensure its uniqueness and consistency. Usually, work agent is uniquely identified by a string that combines managed cluster name and agent name for consistency.
+- Broker: The message broker that handles the event transfer between the source and the work agent. It is represented by a message broker, such as MQTT or Kafka.
+
+### Goals
+
+- Establish a standardized event protocol for delivering manifestworks and returning status information.
+- Define a resync mechanism to handle failures and ensure synchronization.
+- Develop a client interface that facilitates easy integration for consumers.
+
```
bjoydeep, line 36: Should we add a section here that states -
if we had a klusterlet agent running and connected to OCM hub
and now that agent is changed to MQTT based 
how does the network requirements change [port, protocol, initiation direction] or does it not change. This backward compatibility statement should be added here. WDYT guys?

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: Agreed.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
qiujian16: this is a good point, but I think this is not in the scope of this design, we should add it in the milestone. I'd want this design focus on protocol only, followed by another one on how deployment/configuration is done. Which would involves component like operator and registration agent @morvencao

enhancements/sig-architecture/224-event-based-manifestwork/README.md
qiujian16: I think maestro may need to send a status resync message to agent if restart or under certain circumstance. status resync might not be required, but I think it should be a supported message and agent should respond to the message properly. It is broker or server's decision to send or not send the status resync message.

bjoydeep: i am with you. Thanks

enhancements/sig-architecture/224-event-based-manifestwork/README.md
bjoydeep: Should we then mention this is in non-goal perhaps, stating it will be tackled separately?

morvencao: I agreed with incorporating this under "non-goal," ensuring clarity as a separate subject.

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: the sources/agents may need to retry connecting the broker periodically and we may also need a retry times for source/agents, the sources/agents should fail when the number of retries is reached

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: for work agent, it maintains the spec of the manifests on the managed and reports the status of the manifests, if only the status is changed, the work agent will not reapply the manifests, but it will report the manifests new status in the manifestwork

enhancements/sig-architecture/224-event-based-manifestwork/README.md
skeeey: For work agent, it uses the memory as the cache, after the pod restarts, the work agent will send a resync request to the source to sync the resources it maintained  and rebuild its cache

morvencao: The cache, based on [client-go's store](https://pkg.go.dev/k8s.io/client-go/tools/cache#Store) implementation, automatically clears upon pod restarts.

PR #98, opened 2023-07-26
Add the design on long-term policy compliance history

mprahl:


enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+policy (i.e. in the `policy-templates` array). A policy's history is limited to 10 events.
+
+##### Recording a Compliance Event
+
+As previously mentioned, the compliance events will be stored in a relational database. This database will not be
+managed by Open Cluster Management and is limited to just Postgresql at the momement. The connection information for
+this database will be in a secret called `governance-policy-database` in the `open-cluster-management` namespace and
+require a key of `connectionURL` in the format of `postgres://username:password@localhost:5432/database_name`. For
+security reasons, each managed cluster cannot directly write to the database. Instead, an HTTP endpoint is to be added
+to the Policy Propagator to perform authorization checks and translating compliance events to the appropriate SQL
```
gparvin, line 124: So now a service will be defined for the policy propagator.  In OCM it needs to be exposed over an ingress and in OCP it would have a route.  TLS termination might be different between the 2 scenarios.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+  "cluster": { "name": "cluster1" },
+  "parent_policy": { "name": "policies.etcd-encryption" },
+  "policy": {
+    "apiGroup": "policy.open-cluster-management.io",
+    "kind": "ConfigurationPolicy",
+    // Optional if specHash is provided and an entry already exists for that spec hash in the database.
+    // This approach saves on network bandwidth since the spec can be quite large.
+    "spec": {},
+    // Optional if spec is provided
+    "specHash": {}
```
gparvin, line 165: Would it just always provide the `specHash`?  I expect the "spec" is only initialized on the hub by the propagator so managed clusters don't get to define a new spec.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+      "cluster": {
+        "name": "cluster2"
+      },
+      "parent_policy": {
+        "name": "policies.etcd-encryption"
+      },
+      "policy": {
+        "apiGroup": "policy.open-cluster-management.io",
+        "kind": "ConfigurationPolicy",
+        "spec": {},
```
gparvin, line 240: So the entire spec is in all of these output entries?

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+`policy.open-cluster-management.io/policy-evaluated-generation` set to the `generation` of the policy that was
+evaluated. With this additional information, the Status Sync controller can look up the policy by the `resourceVersion`,
+and if by the time the lookup occurs the policy with that resource version is no longer in etcd, the controller will
+lookup the current version of the policy and see if the generation from the annotation matches. Then the spec from that
+lookup will be recorded in the compliance event. If the spec couldn't be determined, a value of `unknown` will be
+recorded.
+
+Note, to reduce bandwidth consumption, the Status Sync controller should cache a list of spec hashes and their
+associated `generation` in memory as a compliance event is recorded. Then, this cache can be used to determine if the
+full spec needs to be retrieved by the Status Sync controller and provided to the `POST` request on the
```
gparvin, line 285: I guess I'm wondering if there's a difference between the hub managed cluster replicated spec and the spec that the managed cluster is sending here in the POST.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+    "per_page": 20,
+    "total": 123
+  }
+}
+```
+
+##### Generating a Report
+
+Another HTTP endpoint would be added at `/api/v1/reports/compliance-events` that would accept a `GET` request with the
+same query parameters as the `/api/v1/compliance-events` HTTP endpoint (excluding pagination), but would generate a CSV
```
gparvin, line 262: In the CSV I am guessing the spec would not be included and the message would be reformatted to not contain commas

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: That's a good idea. The propagator *could* create entries for `spec` in the `policies` table so only the `specHash` needs to be provided from the managed cluster. That does simplify things. I'll update the doc with that.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+	log.Fatal(<-e.Err())
+}
+```
+
+### Risks and Mitigation
+
+### Open Questions [optional]
+
+1. How to enable it on the managed cluster? Should it be per managed cluster?
+1. Should existing policy history be scraped and recorded in the database even though it would not be complete? It may
```
gparvin, line 423: My opinion is no.  That is not necessary to try to capture.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+   policies that mostly stay compliant.
+1. The compliance history retains data of deleted managed clusters and deleted policies.
+1. The outcome is that policy compliance events can be queried by a date range and at least the cluster name, parent
+   policy name, policy name, compliance state, and compliance message can be shown. The solution should allow extending
+   with more data.
+1. This feature should be opt-in.
+
+### Non-Goals
+
+1. Open Cluster Management will not manage the Postgresql database. The user must bring their own.
```
gparvin, line 44: It's not clear that only one hub can write data to this database.  Could a customer with multiple hubs use the same database?  If so maybe the schema needs a hub reference too.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: Yes, do you think it needs to be noted in the design?

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: Yes, I think that's the easiest way to consume the data. Do you have other suggestions?

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: This will be covered in the conversation in https://github.com/open-cluster-management-io/enhancements/pull/98#pullrequestreview-1548522146.

enhancements/sig-policy/98-long-term-compliance-history/README.md
gparvin: It feels a lot like the `managedFields` data for k8s resources.  It's nice to have it there and necessary to be able to get it, but most of the time you don't need to see it... Maybe an `includeSpec` option in the request would be nice to think about.

enhancements/sig-policy/98-long-term-compliance-history/README.md
gparvin: Yes, I think it would be valuable for clarity.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+##### Viewing Compliance Events
+
+Another HTTP endpoint would be added at `/api/v1/compliance-events` that would accept a `GET` request. This would be
+responsible for exposing the compliance events and performing the correct authorization. The authorization would require
+`get` permissions on the cluster's associated `ManagedCluster` object to be able to list the compliance events for the
+cluster. For performance reasons, this is preferred to performing a `SubjectAccessReview` on every cluster namespace for
+a user per request. We can utilize the
+[github.com/stolostron/rbac-api-utils](https://github.com/stolostron/rbac-api-utils/blob/159deac7d398c3470ae730c3fbb8ffbbcca1ac5a/pkg/rbac/rbac.go#L187)
+library to do this.
```
JustinKuli, line 187: I'm just thinking about this permission. Should everyone that can `get` that ManagedCluster be able to see all the compliance events on all the policies on that cluster?

I don't see a great alternative (it seems like custom rbac verbs aren't a thing, otherwise `get-compliance-events` on the ManagedCluster would be ideal), but maybe a non-resource-url `/compliance-events` would help? Users would need a clusterrole for that (since non-resource-urls are cluster-scoped, I think), and additionally they would need `get` on the `ManagedClusters` they're asking for.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+and produce a compliance event. This is currently recorded as a Kubernetes event. The Status Sync controller will
+process the Kubernetes event and add the status entry in the local and hub replicated policy that owns the evaluated
+policy (i.e. in the `policy-templates` array). A policy's history is limited to 10 events.
+
+##### Recording a Compliance Event
+
+As previously mentioned, the compliance events will be stored in a relational database. This database will not be
+managed by Open Cluster Management and is limited to just Postgresql at the momement. The connection information for
+this database will be in a secret called `governance-policy-database` in the `open-cluster-management` namespace and
+require a key of `connectionURL` in the format of `postgres://username:password@localhost:5432/database_name`. For
```
gparvin, line 122: Customers may want to use ESO so having a secret with username and password directly in the secret instead of that URL format I think would be needed.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+would have the etcd GRPC endpoint limited to localhost to avoid needing authentication enabled. The Status Sync
+controller would first try to record the compliance event on the hub, however, if there is no connection, the compliance
+event would be recorded in etcd with the key format of
+`compliance-events/<timestamp from Kubernetes event>-<event-uid>`. For example,
+`compliance-events/2023-07-25T13:15:08Z-4b400c5f-3828-4138-ad4f-03de4d69655b`. The value would be in the format of the
+`POST` request for the `/api/v1/compliance-events` HTTP endpoint on the hub.
+
+The Governance Policy Framework addon pod would run a separate goroutine that would periodically poll etcd for new
+compliance events and try to record them on the hub. It would do this in order by sorting by the timestamps in the keys.
+When successful, it would delete the entry in etcd.
```
JustinKuli, line 321: Why not always send the events to the embedded etcd first, and then always just POST them to the hub from that? I'm thinking that sending the event to the hub would be the slow part of the operation, so waiting for that to complete (or fail) would limit the number of policies the status-sync can handle at once.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+of 1kB. This is very limiting for when the policy components would need to migrate to this new `Event` API. That being
+said, the old `Event` API is not expected to be removed in the near future though.
+
+It is proposed to expose the embedded etcd GRPC endpoint and have policy controllers directly record compliance events
+using it instead of recording Kubernetes events when the embedded etcd is backed by a persistent volume. The GRPC
+endpoint would be protected with authentication and the credentials would be stored in a secret called
+`governance-policy-framework-cache` in the managed cluster namespace. Each policy controller would read this secret on
+start up. Recording the compliance event directly from the controller would allow recording rich metadata about the
+compliance event. One example is the diff between what is in the policy and what is on the cluster, similar to what
+ArgoCD does.
```
JustinKuli, line 347: Is a shared secret the best way to do this? I was expecting the same kind of authorization that is protecting the hub API endpoint (requiring "write access to the status subresource of the policies.policy.open-cluster-management.io resource").

The shared secret does make the embedded etcd simpler, since it doesn't need a proxy in front of it or anything...

enhancements/sig-policy/98-long-term-compliance-history/README.md
JustinKuli: It depends on how hard it is to capture? I think a one-off looking through all the existing policies on the hub and posting compliance events for their histories could be ok. Just set the spec to unknown or something.

mprahl:
I posted up some updates but haven't gotten to all the feedback yet. I'll work on this again tomorrow.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: This schema allows the same database to be used for multiple Open Cluster Management hubs, though if managed cluster names overlap, it'd be less confusing if a separate database on the same Postgresql database be used.

I'll add this in the document.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+| id         | an internal primary key                                                                    |
+| name       | a string column for the cluster name in the ManagedCluster object; this will have an index |
+| cluster_id | the ACM generated unique cluster ID; this will have a unique constraint                    |
+
+##### parent_policies Table
+
+| Column | Description                                                  |
+| ------ | ------------------------------------------------------------ |
+| id     | an internal primary key                                      |
+| name   | a string column for the name of the parent replicated policy |
```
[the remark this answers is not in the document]
mprahl: Yes, that was kept in by accident, though to address @gparvin's comment of storing the categories, controls, and standards, it is now required and the `policies` table was adjusted accordingly.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+##### policies Table
+
+| Column             | Description                                                                  |
+| ------------------ | ---------------------------------------------------------------------------- |
+| id                 | an internal primary key                                                      |
+| kind               | a string column of the policy Kind (e.g. ConfigurationPolicy)                |
+| api_group          | a string column of the policy group (e.g. policy.open-cluster-management.io) |
+| name               | a string column for the name of the policy                                   |
+| parent_policy_name | a string column for the name of the parent replicated policy                 |
```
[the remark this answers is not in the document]
mprahl: I'll clarify this and the format can be seen in the example API calls b elow.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: I was thinking to limit the chance that the message is lost since we aren't guaranteed a persistent volume but we can just prevent the poller/syncer goroutine from letting the pod shutdown until all compliance events are synced on the hub when embedded etcd is not backed by a persistent volume.

I change the design to have it like this.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: Since the embedded etcd is an internal implementation detail, I'd like to avoid using the Kubernetes API for authorization since this adds an unnecessary additional load on the Kubernetes API and would slow recording the compliance events.

mprahl:
@gparvin and @JustinKuli thanks for the feedback! This is ready for review again.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,496 @@
...
+  provided with `?parent_policy_standard=NIST%20SP%20800-53&parent_policy_standard=NIST%20SP%20231-41` syntax.
+- policy - the name of the policy (e.g. a `ConfigurationPolicy`) that generated the compliance event to filter on.
+- policy_kind - the kind of the policy (e.g. `ConfigurationPolicy`) that generated the compliance event to filter on.
+- policy_api_group - the `apiGroup` of the policy (e.g. `policy.open-cluster-management.io`) that generated the
+  compliance event to filter on.
+- include_spec - a boolean to include the policy's spec in the return value. This is `false` by default.
+- before - an ISO-8601 timestamp to indicate only compliance events before this time should be shown.
+- after - an ISO-8601 timestamp to indicate only compliance events after this time should be shown.
+- page - the page number in the query. This defaults to 1.
+- per_page - the number of compliance events returned per page. This defaults to 20 and maxes out at 100.
```
yiraeChristineKim, line 235: How about add `search` query? search with **words**

mprahl: Good idea. I'll add the `message_includes` query parameter.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,496 @@
...
+    "page": 1,
+    "per_page": 20,
+    "total": 123
+  }
+}
+```
+
+##### Generating a Report
+
+Another HTTP endpoint would be added at `/api/v1/reports/compliance-events` that would accept a `GET` request with the
```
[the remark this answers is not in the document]
mprahl: By default, it shows all the compliance events for all managed clusters the user has access to. You can filter with `?cluster=local-cluster&cluster=managed1`.

mprahl:
@gparvin @JustinKuli @yiraeChristineKim, I just added diagrams for each "phase" to help clarify the proposed processes.

enhancements/sig-policy/98-long-term-compliance-history/README.md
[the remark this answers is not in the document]
mprahl: Thanks for pointing that out! It should be `clusters'` and not `cluster's`.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,506 @@
...
+	log.Fatal(<-e.Err())
+}
+```
+
+### Risks and Mitigation
+
+### Open Questions [optional]
+
+1. How to enable it on the managed cluster? Should it be per managed cluster?
+1. Is the RBAC sufficient? Instead of "get" access on a managed cluster, should we have a custom verb of
```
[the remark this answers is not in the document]
mprahl: That's a good point on deleted clusters. I think you could still have a custom verb since `resourceNames` would either be not set or they would include the deleted clusters.

mprahl:
@gparvin, @JustinKuli, and @yiraeChristineKim could you please review this again?

enhancements/sig-policy/98-long-term-compliance-history/README.md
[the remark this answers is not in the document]
gparvin: This can be further clarified in the threat model

gparvin:
To make sure others can get comments considered adding a hold
/hold

gparvin (approved):
This looks good.  Regarding some of the comments around security (rbac/etcd usage/customer provided postgress and secret), I think one of the early work items needs to be creation of a threat model PR.  Can you make sure that is started early in dev?  Thanks!

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,455 @@
...
+between scrapes will be lost, therefore, the data cannot be relied upon.
+
+When a user wants to pass an audit or determine when an incident occurred, it is required to have accurrate, historical,
+and queryable compliance data.
+
+### Goals
+
+1. Provide an accurate compliance history on the hub or other centralized location.
+1. The compliance history is stored in a queryable format.
+1. The complaince history can contain data for at least 2 years for a hub that manages 3000 managed clusters with 30
```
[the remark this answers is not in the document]
JustinKuli: This seems good. I was more just wondering if there was a firm "we must support x" kind of requirement somewhere.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,508 @@
...
+| id         | an internal primary key                                             |
+| name       | a string column for the name of the parent replicated policy        |
+| categories | an array column of the sorted categories associated with the policy |
+| controls   | an array column of the sorted controls associated with the policy   |
+| standards  | an array column of the sorted standards associated with the policy  |
+
+There would be a combined unique constraint on the `name`, `categories`, `controls`, and `standards` columns. Note that
+the array column is a Postgresql feature and not supported in databases such as SQLite. If we do need to support other
+database types in the future, these can be string columns and perhaps have filtering on the API disabled for these
+columns.
```
JustinKuli, line 78: Do other SQL flavors allow filters like "if string x is found inside this column?" Like a substring match...

The way the policy already stores these is just as a comma-delimited list, since each annotation is a string, not a list. So if we could just put each annotation directly into these columns, that could be slightly neater on that side... and then regardless of the database flavor, it might still be possible to get the filtering.

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,508 @@
...
+
+| Column     | Description                                                         |
+| ---------- | ------------------------------------------------------------------- |
+| id         | an internal primary key                                             |
+| name       | a string column for the name of the parent replicated policy        |
+| categories | an array column of the sorted categories associated with the policy |
+| controls   | an array column of the sorted controls associated with the policy   |
+| standards  | an array column of the sorted standards associated with the policy  |
+
+There would be a combined unique constraint on the `name`, `categories`, `controls`, and `standards` columns. Note that
```
JustinKuli, line 75: What would happen if I update a policy, and add a new standard to it?

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -0,0 +1,508 @@
...
+  ],
+  "metadata": {
+    "page": 1,
+    "per_page": 20,
+    "total": 123
+  }
+}
+```
+
+##### Generating a Report
```
JustinKuli, line 304: This report might be very large, right? The proposed GET `/api/v1/compliance-events` has pagination, but a report might have more than 100 entries (the maximum page size). For one "report" style API I worked with (as a consumer), I think I had to do a POST with the parameters for the report, the API responded with some kind of Accepted response, and gave a report ID that could be used later in a GET when the report is ready.

JustinKuli (commented):
Overall looks good to me! Nothing we can't clarify while working on it.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: @JustinKuli, I think a JSON column type may be better for portability to another database type. Postgres and MySQL both have it. SQLite has JSON functions that can parse text columns.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: There would be a new `parent_policies` row. That's what has been the tricky part in this database schema design is we need to store what things were in that point in time and not what they currently are, so it does lead to some data duplication.

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: After doing some more research, I don't think a JSON type is good for unique constraints. I like your idea of allowing substring checking for non-Postgresql databases.

JustinKuli (approved):

PR #99, opened 2023-08-14
[99][sig-policy] Policy placement strategy

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-6523

### DIFF
--- a/enhancements/sig-policy/99-policy-placement-strategy/README.md
+++ b/enhancements/sig-policy/99-policy-placement-strategy/README.md
@@ -1,0 +1,545 @@
+# Policy placement strategy
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Following from the [`DecisionStrategy` enhancement](../../sig-architecture/64-placementStrategy/README.md) in the
+`Placement` API, policies can leverage this new logic to have a configurable and systematic way to roll out policy
+updates to clusters.
+
+## Motivation
+
+Currently policies and subsequent updates are pushed out to clusters en masse based on the placement to which it has
+been bound. The new `DecisionStrategy` field in the `Placement` API creates cluster groupings that controllers can
+leverage to have segmented, configurable rollouts of resources. This will aid use cases where high availability may be a
+priority or a set of test clusters should receive and verify the updates before the remaining clusters get updated.
+
+### Goals
+
+- Make `Placement` the primary API for placement (currently the governance propagator is somewhat
+  `PlacementRule`-centric)
+- Leverage the `Placement` helper library to retrieve placement decisions
+- Reflect rollout status per cluster in the root policy for discoverability (whether a cluster is up-to-date or not)
+- Implement the `RolloutStrategy` struct for policies, including:
+  - `RolloutStrategy`
+    - "All": all clusters at once
+    - "Progressive": one cluster at a time
+    - "ProgressivePerGroup": one group of clusters at a time
+  - `ProgressDeadline`: Maximum amount of time to wait for status before timing out
+  - `MinSuccessTime`: Minimum amount of time to wait before proceeding (i.e. "soak" time)
+  - `MaxFailures`: Tolerance of percentage or number of clusters that can fail and still continue the rollout
+  - `MandatoryDecisionGroups`: groups that should be handled before other groups
+  - `MaxConcurrency`: Concurrency during a progressive rollout
+- Add an aggregated rollout status for the root policy status.
+
+### Non-Goals
+
+Any specialized behaviors outside of those provided by the `Placement` library and, by extension, the `DecisionStrategy`
+enhancement, that might require additional code other than that already provided/required by the `Placement` library.
+This includes the ability to roll back a rollout. A rollback should be implemented by the user by applying the previous
+version of the policy, and GitOps is the recommended method for this.
+
+## Proposal
+
+### Background
+
+**Note:** This enhancement references the previous
+[`DecisionStrategy` enhancement](../../sig-architecture/64-placementStrategy/README.md) in the `Placement` API, so it is
+recommended to review it before proceeding.
+
+In current policy flows, users create policies on the hub cluster (which are referred to as the root policy). These
+policies are replicated by the `governance-policy-propagator` controller to managed cluster namespaces based on the
+placement to which they have been bound by the associated `PlacementBinding` object. The `governance-policy-framework`
+sync controllers on managed clusters monitor the associated managed cluster namespace on the hub for policy updates.
+
+Additionally, policies have a `remediationAction` specified, which can either be `inform`, meaning the policy controller
+monitors objects on the managed cluster reporting status without making changes, or `enforce`, meaning the policy
+controller attempts to make changes on the managed cluster based on the policy definition and reports status accordingly
+based on the success or failure of the update.
+
+### Design
+
+Upon policy creation or update, all clusters receive the same version, preventing version skew among different clusters.
+However, the `remediationAction` for enforced policies is initially set to `inform` and the rollout occurs as each
+cluster in turn has its policy updated to `enforce`. For each rollout strategy, this looks like:
+
+| Rollout Strategy      | Remediation Inform                      | Remediation Enforce                                                                                                                                                            |
+| --------------------- | --------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `All`                 | Policy rolled out to all clusters as-is | Policy rolled out to all clusters as-is                                                                                                                                        |
+| `Progressive`         | (Same as `All`)                         | Policy applied initially to all clusters as `inform` and <br/>then serially enforces each cluster's policy and moves on to the next cluster once compliant (or timed out)      |
+| `ProgressivePerGroup` | (Same as `All`)                         | Policy applied initially to all clusters as `inform` and <br/>rollout enforces a cluster group's policy and waits for compliant (or timed out) before moving to the next group |
+
+Given this, policy propagation by the `governance-policy-propagator` controller on the hub takes place in a
+multiple-pass approach:
+
+1. First pass: Replicate policies to all clusters set with `remediationAction: inform`
+
+   On the first pass following a root policy creation or update, the `governance-policy-propagator` controller on the
+   hub cluster will replicate all policies to bound managed cluster namespaces and are created regardless of the rollout
+   strategy specified, with the remediation action set to `inform` (the exception being if the rollout strategy is
+   `All`). This way, all policies are up-to-date and reporting a compliance status based on the current version of the
+   root policy. The aggregated rollout status on the root policy is set to `Progressing`. If the remediation action on
+   the root policy is already `inform` or the strategy is `All`, the rollout status on each cluster's policy is set to
+   `Progressing` and no second pass occurs. If the remediation action is `enforce`, the rollout status on each cluster's
+   policy is set to `ToApply`.
+
+2. Subsequent passes: Use the given rollout strategy to set `remediationAction: enforce`, stopping the rollout on
+   non-compliance
+
+   On the subsequent passes, if the `remediationAction` is `enforce` and has a progressive rollout strategy, the
+   `governance-policy-propagator` fetches the managed clusters returned from the `Placement` library (based on the
+   user-configured rollout strategy) and sets the rollout status of each returned cluster to `Progressing`, updating any
+   `remediationAction` to `enforce` as defined in the root policy.
+
+From here, work is picked up by the `governance-policy-framework` controllers on the managed clusters:
+
+1. The template sync controller of the `governance-policy-framework` updates the applicable `ConfigurationPolicy`
+   objects, which triggers reevaluation from the `config-policy-controller`.
+2. Once the `ConfigurationPolicies` have a generation that matches the `lastEvaluatedGeneration` set on the replicated
+   policy status, the status sync controller of the framework will set the rollout status on the replicated policy on
+   the managed cluster and the replicated policy on the hub to `Succeeded` after all of the `ConfigurationPolicies` have
+   a compliant status or `Failed` if there is any non-compliant status. (The `governance-policy-propagator` will also
+   set the rollout status accordingly on the root policy for that managed cluster.)
+
+Bringing this together, the rollout status of a managed cluster policy is defined as follows:
+
+| Status        | Description                                                                                                                                                                                                                                          |
+| ------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ToApply`     | A policy to be enforced in a progressive rollout is created/updated in the first pass of the propagator on the hub                                                                                                                                   |
+| `Progressing` | Policy has no status or the `lastEvaluatedGeneration` does not match the generation on the managed cluster, and one of: Rollout is `All`, policy was defined as `inform`, or policy was selected by the `Placement` library and updated to `enforce` |
+| `Succeeded`   | Policy is compliant and the `lastEvaluatedGeneration` matches the generation on the managed cluster                                                                                                                                                  |
+| `Failed`      | Policy is non-compliant and the `lastEvaluatedGeneration` matches the generation on the managed cluster                                                                                                                                              |
+| `TimeOut`     | Time has passed beyond the timeout specified in the `RolloutStrategy` without a returned status                                                                                                                                                      |
+| `Skip`        | (unused)                                                                                                                                                                                                                                             |
+
+And the aggregated root policy status would be: `Progressing` (this would include `ToApply`), `Succeeded`, or `Failed`.
+(The `TimeOut` status would not be reflected.)
+
+#### Handling managed cluster additions to the placement
+
+When a new managed cluster is added to the placement, if the rollout is `Progressing`, it will be handled with the
+current rollout where the `Placement` library will abstract this and return the cluster at the appropriate time
+according to the rollout strategy. If the rollout is `Succeeded`, the `governance-policy-propagator` will update the
+rollout status on the root policy back to `Progressing` and handle only the newly added managed clusters in a "miniature
+rollout" as described in the [Design section](#design).
+
+#### Handling complexities around the `PlacementBinding`
+
+There is some additional complexity brought into the rollout because policies are bound to `Placement` using
+`PlacementBindings`, so there could be multiple `Placements` bound to each policy. To remedy this, `All` and
+`Progressive` (which is per-cluster) would be handled by pulling in all `PlacementDecisions` from all `Placements`. In
+the case of multiple `Placements` where `ProgressivePerGroup` is specified, this would be handled by taking groups from
+one placement and then moving on to the next placement.
+
+Additionally, the `PlacementBinding` has specifications for selective policy enforcement to override the
+`remediationAction` and selectively apply policies to managed clusters. (See the
+[Selective Policy Enforcement enhancement](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-policy/28-selective-policy-enforcment).)
+Since these may conflict with the rollout and has similar intentions, these specifications will be ignored if they are
+provided with a policy specifying a rollout strategy other than `All`.
+
+### User Stories
+
+#### Story 1
+
+As a system administrator, I want to know the status of a rollout for a policy and what clusters have been updated.
+
+- **Summary**
+
+  - Add a `RolloutStatus` to the `Policy` status in the CRD.
+
+- **Details**
+
+  - The `RolloutStatus` is added to the status on the replicated policy.
+  - The `RolloutStatus` is added on the root policy per-cluster and as an aggregated status.
+
+  (For technical notes see [Development Phase 1](#development-phase-1))
+
+- **Snippet**
+
+  **Root policy status**
+
+  ```yaml
+  status:
+    compliant: Compliant
+    rolloutStatus: Progressing # <-- New field
+    placement:
+      - placement: sample-placement
+        placementBinding: sample-placementbinding
+    status:
+      - clustername: cluster1
+        clusternamespace: cluster1
+        compliant: Compliant
+        rolloutStatus: Succeeded # <-- New field
+      - clustername: cluster2
+        clusternamespace: cluster2
+        compliant: ""
+        rolloutStatus: Progressing
+      - clustername: cluster3
+        clusternamespace: cluster3
+        compliant: ""
+        rolloutStatus: ToApply
+  ```
+
+  **Replicated policy status**
+
+  ```yaml
+  status:
+    compliant: Compliant
+    rolloutStatus: Succeeded # <-- New field
+    details:
+      - compliant: Compliant
+        history:
+          - ...
+  ```
+
+#### Story 2
+
+As a system administrator, I want to specify a placement `DecisionStrategy` bound with policies, configure the order of
+the updates, and continue enforcement rollout only when the policy on previous clusters show as "Compliant".
+
+- **Summary**
+
+  - Policies are initialized with `remediationAction: inform` and enforced policies are updated to
+    `remediationAction: enforce` according to the `rolloutStrategy` defined on the policy.
+  - Policy enforcement is continued only once the previously deployed clusters show as "Compliant".
+
+  (For technical notes see [Development Phase 2](#development-phase-2) and [Development Phase 3](#development-phase-3))
+
+- **Snippet**
+
+  **Policy with `rolloutStrategy`**
+
+  (See the
+  [DecisionStrategy enhancement](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/64-placementStrategy#how-workload-applier-apis-will-benefits-from-using-placementstrategy)
+  for details on the `RolloutStrategy` struct.)
+
+  ```yaml
+  apiVersion: policy.open-cluster-management.io/v1
+  kind: Policy
+  metadata:
+    name: sample-policy
+    namespace: policy-namespace
+  spec:
+    disabled: false
+    remediationAction: enforce
+    rolloutStrategy:               # <-- New field
+      type: ProgressivePerGroup
+    policy-templates:
+      - objectDefinition:
+          apiVersion: policy.open-cluster-management.io/v1
+          kind: ConfigurationPolicy
+          ...
+  ---
+  apiVersion: policy.open-cluster-management.io/v1
+  kind: PlacementBinding
+  metadata:
+    name: sample-placementbinding
+    namespace: policy-namespace
+  placementRef:
+    name: sample-placement
+    apiGroup: cluster.open-cluster-management.io
+    kind: Placement
+  subjects:
+    - name: sample-policy
+      apiGroup: policy.open-cluster-management.io
+      kind: Policy
+  ```
+
+  **Replicated policy in managed cluster namespace**
+
+  ```yaml
+  apiVersion: policy.open-cluster-management.io/v1
+  kind: Policy
+  metadata:
+    name: policy-namespace.sample-policy
+    namespace: cluster1
+  spec:
+    disabled: false
+    remediationAction: inform # <-- Remediation action updated to prepare for rollout
+    rolloutStrategy:
+      type: ProgressivePerGroup
+      progressivePerGroup:
+        minSuccessTime: 5m
+        progressDeadline: 10m
+        maxFailures: 2%
+        mandatoryDecisionGroups: []
+    policy-templates:
+      - objectDefinition:
+          spec:
+            apiVersion: policy.open-cluster-management.io/v1
+            kind: ConfigurationPolicy
+            ...
+  status:
+    compliant: ""
+    rolloutStatus: ToApply # <-- Policy is waiting to be enforced
+  ```
+
+  **Placement**
+
+  (provided for reference--this enhancement doesn't intend to update the Placement API. See the
+  [DecisionStrategy enhancement](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/64-placementStrategy#placement-api)
+  for details on the `DecisionStrategy` struct.)
+
+  ```yaml
+  apiVersion: cluster.open-cluster-management.io/v1beta1
+  kind: Placement
+  metadata:
+    name: sample-placement
+    namespace: policy-namespace
+  spec:
+    predicates:
+      - requiredClusterSelector:
+          labelSelector:
+            matchExpressions:
+              - key: environment
+                operator: In
+                values:
+                  - dev
+    decisionStrategy:
+      groupStrategy:
+        clustersPerDecisionGroup: 10
+        DecisionGroups:
+          - groupName: dev-emea
+            clusterSelector:
+              matchExpressions:
+                - key: dev-emea
+                  operator: Exist
+          - groupName: dev-apac
+            clusterSelector:
+              matchExpressions:
+                - key: dev-apac
+                  operator: Exist
+  ```
+
+### Implementation Details
+
+#### Development Phase 1
+
+- **Summary**
+
+  - Add a `RolloutStatus` to the `Policy` status in the CRD.
+
+- **Details**
+
+  - The `RolloutStatus` would be added to reflect: rollout status on the replicated policy, per-cluster on the root
+    policy, and an aggregated status on the root policy. The aggregated root policy status would be: `Progressing` (this
+    would include `ToApply`), `Succeeded`, or `Failed`. (The `TimeOut` status would not be reflected.)
+
+- **Snippet**
+
+  ```golang
+  // CompliancePerClusterStatus defines compliance per cluster status
+  type CompliancePerClusterStatus struct {
+    ComplianceState  ComplianceState               `json:"compliant,omitempty"`
+    RolloutStatus    clusterv1alpha1.RolloutStatus `json:"rolloutStatus,omitempty"` // <-- New field
+    ClusterName      string                        `json:"clustername,omitempty"`
+    ClusterNamespace string                        `json:"clusternamespace,omitempty"`
+  }
+
+  // PolicyStatus defines the observed state of Policy
+  type PolicyStatus struct {
+    Placement []*Placement                  `json:"placement,omitempty"`
+    Status    []*CompliancePerClusterStatus `json:"status,omitempty"`
+
+    // +kubebuilder:validation:Enum=Compliant;Pending;NonCompliant
+    ComplianceState ComplianceState               `json:"compliant,omitempty"`
+    RolloutStatus   clusterv1alpha1.RolloutStatus `json:"rolloutStatus,omitempty"` // <-- New field
+    Details         []*DetailsPerTemplate         `json:"details,omitempty"`
+  }
+
+  // Policy is the Schema for the policies API
+  // +kubebuilder:subresource:status
+  // +kubebuilder:resource:path=policies,scope=Namespaced
+  // +kubebuilder:resource:path=policies,shortName=plc
+  // +kubebuilder:printcolumn:name="Remediation action",type="string",JSONPath=".spec.remediationAction"
+  // +kubebuilder:printcolumn:name="Compliance state",type="string",JSONPath=".status.compliant"
+  // +kubebuilder:printcolumn:name="Rollout status",type="string",JSONPath=".status.rolloutStatus" // <-- New printer column
+  // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+  type Policy struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata"`
+
+    Spec   PolicySpec   `json:"spec"`
+    Status PolicyStatus `json:"status,omitempty"`
+  }
+  ```
+
+#### Development Phase 2
+
+- **Summary**
+
+  - Use the `Placement` library to gather the list of clusters and iterate over them.
+
+- **Details**
+
+  - Make `Placement` the primary placement resource and parse `PlacementRule` decisions into the `Placement` struct
+    instead of the `PlacementRule` struct.
+  - Using the `Placement` library and implementing a `ClusterRolloutStatusFunc`, iterate over all clusters as usual
+    using the `All` rollout strategy.
+
+- **Snippet (untested)**
+
+  ```golang
+  // common.go
+  type placementDecisionGetter struct {
+    c client.Client
+  }
+
+  func (pd placementDecisionGetter) List(selector labels.Selector, namespace string) ([]*clusterv1beta1.PlacementDecision, error) {
+    pdList := &clusterv1beta1.PlacementDecisionList{}
+    lopts := &client.ListOptions{
+      LabelSelector: selector,
+      Namespace:     namespace,
+    }
+
+    err := pd.c.List(context.TODO(), pdList, lopts)
+    if err != nil {
+      return nil, err
+    }
+
+    pdPtrList := []*clusterv1beta1.PlacementDecision{}
+    for _, pd := range pdList.Items {
+      pdPtrList = append(pdPtrList, &pd)
+    }
+
+    return pdPtrList, err
+  }
+
+  func GetRolloutHandler(c client.Client, placement *clusterv1beta1.Placement) (*clusterv1alpha1.RolloutHandler, error) {
+    pdTracker := clusterv1beta1.NewPlacementDecisionClustersTracker(placement, placementDecisionGetter{c}, nil, nil)
+
+    _, _, err := pdTracker.Get()
+    if err != nil {
+      log.Error(err, "Error retrieving PlacementDecisions from tracker")
+    }
+
+    return clusterv1alpha1.NewRolloutHandler(pdTracker)
+  }
+
+  var GetClusterRolloutStatus clusterv1alpha1.ClusterRolloutStatusFunc = func(clusterName string) clusterv1alpha1.ClusterRolloutStatus {
+    // Fetch and return the rollout status and the last transition time from the status in the replicated policy
+
+    return clusterv1alpha1.ClusterRolloutStatus{
+      Status: "<rollout-status>",
+      LastTransitionTime: "<transition-time>",
+    }
+  }
+  ```
+
+  ```golang
+  rolloutHandler := common.GetRolloutHandler(client, placement)
+  strategy, rolloutResult, err := rolloutHandler.GetRolloutCluster(clusterv1alpha1.All, common.GetClusterRolloutStatus)
+
+  // ... Use rolloutResults for propagation
+  ```
+
+#### Development Phase 3
+
+- **Summary**
+
+  - Only continue policy enforcement once the previously deployed clusters show as "Compliant".
+
+- **Details**
+
+  - Update the `Policy` CRD to contain the `RolloutStrategy` struct. (See the
+    [`v1alpha1/RolloutStrategy`](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/types_rolloutstrategy.go))
+    Defaults to `All` if a strategy is not provided or the remediation action is `inform`.
+  - On create/update, replicate policy to all managed clusters with `remediationAction: inform` (if the rollout strategy
+    isn't `All`).
+  - When the `remediationAction` is set to `enforce`, policies not currently being rolled out will be set to `inform` to
+    continue to report status without making changes on the managed cluster while waiting for the new version of the
+    policy to be enforced.
+  - Update the `ClusterRolloutStatusFunc`, (the `GetClusterRolloutStatus` function from
+    [Development Phase 2](#development-phase-2)) to fetch the rollout status and last transition time from the
+    replicated policy status. (See the [Design section](#design).)
+
+    | Status        | Description                                                                                                                                           |
+    | ------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+    | `ToApply`     | A policy to be enforced in a progressive rollout is created/updated in the first pass of the propagator on the hub                                    |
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,267 @@
...
+  - `Timeout`: Maximum amount of time to wait for success before continuing the rollout
+  - `MandatoryDecisionGroups`: groups that should be handled before other groups
+  - `MaxConcurrency`: Concurrency during a progressive rollout
+- (stretch) Add an aggregated rollout status for the root policy status.
+
+### Non-Goals
+
+Any specialized behaviors outside of those provided by the `Placement` library and, by extension, the `DecisionStrategy`
+enhancement, that might require additional code other than that already provided/required by the `Placement` library.
+
```
gparvin, line 45: Maybe list Rollback here specifically.  Rollback would require the user to re-apply the old policy after seeing the canary failure.

dhaiducek: Added:
> This includes the ability to roll back a rollout. A rollback should be implemented by the user by applying the previous version of the policy, and GitOps is the recommended method for this.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+- Leverage the `Placement` helper library to retrieve placement decisions
+- Reflect rollout status per cluster in the root policy for discoverability (whether a cluster is up-to-date or not)
+- (stretch) Implement the `RolloutStrategy` struct for policies, including:
+  - `RolloutStrategy`
+    - "All": all clusters at once
+    - "Progressive": one cluster at a time
+    - "ProgressivePerGroup": one group of clusters at a time
+  - `Timeout`: Maximum amount of time to wait for success before continuing the rollout
+  - `MandatoryDecisionGroups`: groups that should be handled before other groups
+  - `MaxConcurrency`: Concurrency during a progressive rollout
```
yiraeChristineKim, line 38: When `RolloutStrategy: Progressive`, MaxConcurrency isn't needed?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **Summary**
+
+  - Add an annotation to the replicated Policy indicating whether a policy is up-to-date.
+
+- **Details**
+
+  - Update the `governance-policy-propagator` controller to add a `open-cluster-management.io/root-policy-generation`
+    annotation to the replicated policy for each cluster.
+  - Update the annotation with the policy's generation at the time of the update when a policy has been replicated.
```
yiraeChristineKim, line 64: I am confused here. So this annotation is updated when a root policy has been replicated when the root policy changed e.g. pod limit change, Is this annotation changed?

dhaiducek: Yes, when the policy is replicated to the managed cluster namespace on the hub after the root policy is updated, the generation annotation will also be updated.

yiraeChristineKim: You mean when the root policy spec updated, policy is replicated, not update?

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: Per the decision strategy struct, `MaxConcurrency` is only defined for `Progressive` rollouts: https://github.com/open-cluster-management-io/api/blob/2b2399b5f6e88e28a962aa82307ccc00bc492009/cluster/v1alpha1/types_rolloutstrategy.go#L95-L99

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: Correct--there might be too many words in there--I'll update it.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Goals
+
+- Make `Placement` the primary API for placement (currently the governance propagator is somewhat
+  `PlacementRule`-centric)
+- Leverage the `Placement` helper library to retrieve placement decisions
+- Reflect rollout status per cluster in the root policy for discoverability (whether a cluster is up-to-date or not)
+- (stretch) Implement the `RolloutStrategy` struct for policies, including:
+  - `RolloutStrategy`
+    - "All": all clusters at once
```
mprahl, line 33: Would this be considered stretch because it's the current behavior? So at a minimum, "All" would be supported.

dhaiducek: Correct. At a minimum, we'd implement the `Placement` library using `All` without making it configurable by the user.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+As a system administrator, I want to know the status of a rollout for a policy and what clusters have been updated.
+
+- **Summary**
+
+  - Add an annotation to the replicated Policy indicating whether a policy is up-to-date.
+
+- **Details**
+
+  - Update the `governance-policy-propagator` controller to add a `open-cluster-management.io/root-policy-generation`
```
mprahl, line 62: This alone isn't sufficient because it doesn't take into account if the managed cluster acted upon it.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+    annotation to the replicated policy for each cluster.
+  - Update the annotation with the policy's generation at the time when a policy has been replicated.
+  - (Optional--requirement for phase 3) Update the `governance-policy-propagator` controller to add a
+    `open-cluster-management.io/root-policy-timestamp` annotation to the replicated policy for each cluster indicating
+    when the replicated policy was last applied.
```
mprahl, line 67: As in the last time the managed cluster synced the propagated policy?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+    timeout.
+
+- **Details**
+
+  - Update the `PlacementBinding` CRD to contain the `RolloutStrategy` struct. (See the
+    [`v1alpha1/RolloutStrategy`](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/types_rolloutstrategy.go))
+    Default to `All` if a strategy is not provided.
+  - Update the `governance-policy-propagator` controller to add a `open-cluster-management.io/root-policy-timestamp`
+    annotation to the replicated policy for each cluster indicating when the replicated policy was last applied.
+  - Store a `lastEvaluated` in the status of the replicated policies so that the `RolloutStatus` can be determined
```
mprahl, line 153: Could you please elaborate on this more?

I'm thinking about the case where the root policy is updated and the ConfigurationPolicy gets evaluated but the status doesn't change, how would this work? Also, if the root policy is updated but the generation of the ConfigurationPolicy doesn't change and its evaluation interval is like `2h` or `never` when compliant, this would not trigger an evaluation in time.

Other controllers also don't post when they last evaluated.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+    // +kubebuilder:validation:MinItems=1
+    Subjects []Subject              `json:"subjects"`
+    Status   PlacementBindingStatus `json:"status,omitempty"`
+  }
+  ```
+
+  ```golang
+  // ComplianceHistory defines compliance details history
+  type ComplianceHistory struct {
+    LastEvaluated metav1.Time `json:"lastEvaluated,omitempty" protobuf:"bytes,7,opt,name=lastEvaluated"`  // <-- New field
```
mprahl, line 199: Similar to my last comment, how is this field set?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Open Questions
+
+1. Should the rollout handler be instantiated with the dynamic watcher instead of the controller client?
+2. Should `RolloutStatus` be reflected per-cluster in the root policy? (Removing `clusterNamespace` could essentially
+   make the CRD the same size as it currently is, remediating concerns around bloat, but the rollout status is also
+   reflected in each replicated policy.)
+3. Is there opposition to removing `Placement.Decisions` from the `status`? (This could alternatively be handled outside
+   of this enhancement.)
+4. Should the `RolloutStrategy` be implemented in the `Policy` CRD instead of `PlacementBinding`?
```
mprahl, line 253: I prefer this level.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: Right--that's the purpose of the `lastEvaluated` status update in Phase 3. I was thinking for Phase 1 and 2, match current behaviors without figuring in other rollout strategies other than `All` so that the annotations could be implemented separately from the `Placement` implementation.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: The time that the governance-policy-propagator create/updated the replicated policy. That'd be the baseline for the timeout.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: @dhaiducek on a rollout, we could consider resetting the policy status from the Policy Propagator or setting it to a new status indicating a rollout is in progress. The status sync could recognize this and reset the compliance status on the policy templates. There is already logic in the controllers to always send a compliance event when there is no status and it would cause configuration policies to be evaluated immediately regardless of evaluation interval. Then when the policies get a new status, we know the deployment has occurred.

You'd have to think about the specific workflow but I think that may be a good option to communicate between the hub and managed clusters during a rollout.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: We do provide the `LastEvaluated` in the ConfigurationPolicy, so I was thinking we can add logic to have the status-sync controller update the evaluation time when the `root-policy-timestamp` is greater than the last evaluated time, but I'd have to dig into when the `LastEvaluated` gets sent if it's not on every reconcile...
```golang
        // An ISO-8601 timestamp of the last time the policy was evaluated
        LastEvaluated string `json:"lastEvaluated,omitempty"`
```
https://github.com/open-cluster-management-io/config-policy-controller/blob/7e648171a8bd1ac051fe130f5539fbc36347907b/api/v1/configurationpolicy_types.go#L197 

Good point about other controllers, though. I'm not sure--I'll have to consider it more and look at our code. Maybe wipe the status on an update to let it repopulate or something?

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: Moving this thread to https://github.com/open-cluster-management-io/enhancements/pull/99#discussion_r1294997249 to prevent duplication.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: Moving thread to https://github.com/open-cluster-management-io/enhancements/pull/99#discussion_r1294997249

enhancements/sig-policy/99-policy-placement-strategy/README.md
gparvin: This means it needs to be in PolicySet as well.

gparvin (commented):
I added my comments.  Thanks for the work on this!

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+  ```
+
+#### Story 3 / Phase 3 (Stretch goal)
+
+As a system administrator, I want policies to be rolled out only when the policy on previous clusters show as
+"Compliant" and control their concurrency and priority.
+
+- **Summary**
+
+  - Only continue policy deployment once the previously deployed clusters show as "Compliant" after a configurable
```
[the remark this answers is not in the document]
dhaiducek: In the decision strategy, there's a timeout to wait for "Success" (which in the case of Policy is "Compliant"). The rollout will continue when clusters in the current rollout have succeeded OR have timed out.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+    | ------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
+    | `ToApply`     | No generation set                                                                                                                                    |
+    | `Progressing` | Current generation and `lastEvaluated` has not been updated                                                                                          |
+    | `Succeed`     | Compliant and current generation                                                                                                                     |
+    | `Failed`      | NonCompliant and current generation                                                                                                                  |
+    | `TimeOut`     | Current generation, `lastEvaluated` is less than `root-policy-timestamp`, </br>and time has passed beyond timeout specified in the `RolloutStrategy` |
```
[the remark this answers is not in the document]
dhaiducek: The `lastEvaluated` might be reworked to use the transition time instead (See https://github.com/open-cluster-management-io/enhancements/pull/99#discussion_r1294678804). The root-policy-timestamp would be when the replicated policy spec was created/updated with the current generation, so any time before that would have been a previous version of the policy.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+  - Add an annotation to the replicated Policy indicating whether a policy is up-to-date.
+
+- **Details**
+
+  - Update the `governance-policy-propagator` controller to add a `open-cluster-management.io/root-policy-generation`
+    annotation to the replicated policy for each cluster.
+  - Update the annotation with the policy's generation at the time of the update when a policy has been replicated.
+  - (Optional--requirement for phase 3) Update the `governance-policy-propagator` controller to add a
+    `open-cluster-management.io/root-policy-timestamp` annotation to the replicated policy for each cluster indicating
```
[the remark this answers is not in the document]
dhaiducek: Correct--and that might be a better name for it 🙂

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+  - Use the `Placement` library to gather the list of clusters and iterate over them as it is today.
+
+- **Details**
+
+  - Make `Placement` the primary placement resource and parse `PlacementRule` decisions into the `Placement` struct
+    instead of the `PlacementRule` struct.
+  - Using the `Placement` library and implementing a basic `ClusterRolloutStatusFunc`, iterate over all clusters as
+    usual using the `All` rollout strategy. The `RolloutStatus` will initially be either `ToApply` or `Succeed`.
+  - (Optional cleanup) Remove the current `Placement.Decisions` from the `status` of the `Policy` CRD. It's uncertain
```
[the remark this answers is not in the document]
dhaiducek: Right--I might make a separate issue to look at the CRD cleanup outside of this enhancement. But I'm also not sure it's used anywhere...

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+2. Should `RolloutStatus` be reflected per-cluster in the root policy? (Removing `clusterNamespace` could essentially
+   make the CRD the same size as it currently is, remediating concerns around bloat, but the rollout status is also
+   reflected in each replicated policy.)
+3. Is there opposition to removing `Placement.Decisions` from the `status`? (This could alternatively be handled outside
+   of this enhancement.)
+4. Should the `RolloutStrategy` be implemented in the `Policy` CRD instead of `PlacementBinding`?
+5. How should a policy handle when placement decision groups shift? (This implementation leans on the `Placement`
+   library to handle such shifts, only handling the clusters that it returns.)
+6. (An aside) What's with the pointer in `[]*clusterv1beta1.PlacementDecision`?
+
```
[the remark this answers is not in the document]
dhaiducek: Right, I'm not sure placement impacts PolicyAutomation, but definitely something to verify.

Odd indeed, but the old templates ought to be reflected in the replicated policy, and it'd hopefully be relatively straightforward to find the rollout status in the root policy to figure out that it hadn't been updated yet. That said, maybe storing a per-cluster rollout status in the root policy would be good for visibility.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+- **Snippet**
+
+  ```golang
+  type PlacementBinding struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    // +kubebuilder:validation:Optional
+    BindingOverrides BindingOverrides `json:"bindingOverrides,omitempty"`
+    // +kubebuilder:validation:Optional
+    RolloutStrategy clusterv1alpha1.RolloutStrategy `json:"rolloutStrategy,omitempty"` // <-- New field
```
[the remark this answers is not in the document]
dhaiducek: Good questions. I need to figure out how selective policy enforcement works, but I think they could work together, with the propagator filtering out clusters from the clusters returned from the Placement library.

The `additionalPrinterColumns` would be a good addition to the CRD!

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+### User Stories
+
+#### Story 1 / Phase 1
+
+As a system administrator, I want to know the status of a rollout for a policy and what clusters have been updated.
+
+- **Summary**
+
+  - Add an annotation to the replicated Policy indicating whether a policy is up-to-date.
```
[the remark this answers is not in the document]
dhaiducek: No--it's mainly implementing the metadata. I meant to put Story 1 and Story 2 both in Phase 1 since they could be completed concurrently and apparently neglected to do so. 🙂

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: @mprahl I just realized that "this" was unclear. Are you saying you prefer it being in the `Policy`?

enhancements/sig-policy/99-policy-placement-strategy/README.md
JustinKuli, line 62: Just thinking out loud, but I don't know if the generation number is enough here. Maybe I'm being paranoid, but if I delete the root policy and recreate it quickly enough that the propagator hasn't removed the old ones yet, I may have changed the policy while keeping the generation the same (or maybe the generation here is actually a larger than the one on the root policy). With the long-term compliance history stuff, we're already considering making a hash of the policy to identify it, I think? Maybe that could also be used here...

enhancements/sig-policy/99-policy-placement-strategy/README.md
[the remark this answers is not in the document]
JustinKuli: +1 on something like `open-cluster-management.io/replication-timestamp`

enhancements/sig-policy/99-policy-placement-strategy/README.md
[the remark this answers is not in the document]
JustinKuli: Re: API updates, I think we could stop populating the field, without removing it from the CRD (and therefore without requiring a version bump). Especially since it's just in the status, it doesn't feel like it's as much of a breaking change.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,269 @@
...
+### Implementation Details/Notes/Constraints
+
+For the `Placement` library, this requires importing at least this package version (the `Placement` library is in the
+`v1alpha1` version):
+
+```
+  open-cluster-management.io/api v0.11.1-0.20230809113658-2b2399b5f6e8
+```
+
+For testing, the `governance-policy-propagator` doesn't currently account for multiple managed clusters. As part of this
```
JustinKuli, line 226: I think it does currently consider multiple clusters, just in a fake way (which is probably what we want). I'm not saying that what it does is necessarily great, it probably needs to be looked at in detail for testing this feature, but it is already there to an extent: https://github.com/stolostron/governance-policy-propagator/blob/main/test/e2e/case1_propagation_test.go#L78

enhancements/sig-policy/99-policy-placement-strategy/README.md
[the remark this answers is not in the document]
dhaiducek: How about `last-replicated`?

enhancements/sig-policy/99-policy-placement-strategy/README.md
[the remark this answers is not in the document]
dhaiducek: @JustinKuli We'll have to investigate it (and I'll probably create a new issue to not muddle this enhancement), but I don't think we populate it currently 😅

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: I thought about the hash and especially if the compliance history would make use of it, that'd definitely be the way to go. It becomes complex, however, if hub templates are involved since the root policy and replicated policy would look different, which is why I went with the generation. If the root policy generation is less than the annotation on the replicated one, that could also be a signal to delete the replicated policy and recreate it.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: @dhaiducek after brainstorming with @gparvin since he has a good sense of the requirements from a core user, I'd like to propose the following:

* New policies are progressively propagated to clusters regardless of whether they are inform or enforce. The status of the root policy should indicate that cluster in later groups are pending rollout or some kind of other state.
* Existing policies that are updated should only have their `remediationAction: enforce` progressively applied. Take for instance a scenario where we have clusters A, B, and C and A is a canary cluster. An update to my policy that sets `remediationAction: enforce` would at first have `remedation: enforce` on cluster A and `remediationAction: inform` on clusters B and C. The root policy status needs to indicate that clusters B and C are pending enforcement. For example, their compliance state could be `NonCompliant - Pending Rollout`.
* PolicyAutomation (separate story) would have a new mode of `everyRollout` so can be triggered when a rollout fails as opposed to each cluster becoming noncompliant.


@dhaiducek and @JustinKuli , what do you think?

dhaiducek: I like the idea to switch to inform to prevent unintended updates. We could add the `RolloutStatus` to each cluster's status, but I'm not sure we should change the compliance state to NonCompliant during a rollout, especially if users have alerts set up or are active users of the dashboards--that could get unintentionally noisy? Maybe we just use the `Pending` state? But overall I think this is the right direction to having a controlled rollout.

mprahl: It's a fair point but at the same time, as a user, I would like to see the state of the policy on the cluster if there are inform policies applied even if they are pending rollout.

We could use the `Pending` state but include the noncompliant message so at least the information is there.

What do you think @dhaiducek and @gparvin?

gparvin: I prefer the compliance being NonCompliant when that reflects reality and "maybe" have a separate Remediation Pending status (new field).  Though there may be too many things we could do in this extra remediation status field:
- pending rollout
- remediation applied by policy controller
- automated remediation applied
- remediated by automation or user intervention
- no remediation needed

dhaiducek: @gparvin I'm unclear on when you say "when that reflects reality". Are you saying policies should continue to report based on the content it has at the time, or are you saying that if the policy is out-of-date, it should be considered NonCompliant? For the latter, I suppose I'm thinking about a case where it might have been a minor update to the policy where the ultimate compliance wouldn't have changed due to the update? But maybe there's a "Policy" best practice I'm missing there if that's the meaning.
And, if we set it to "NonCompliant" because a rollout is happening, how would setting it to "NonCompliant" affect Ansible automation? As @gparvin points out, we might need some way to distinguish "NonCompliant--this is out of date" vs "NonCompliant--this is broken"? In that case, should Ansible automation only act when the `RolloutStatus` is `Succeed`? (I'm crossing my fingers that gets updated to `Succeeded` 😆)
Interestingly, to the "too many things we could do" point, having an `enum` for status is apparently against the Kubernetes status conventions (though it seems `RolloutStatus` is already set as such), where "Status" should simply be "True", "False", or "Unknown": https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

mprahl: @dhaiducek, I think the point is that if we follow the idea that a progressive rollout to a policy update makes the enforce part progressive and nothing else, then the pending rollout clusters should still accurately report back their inform status.

Having an additional field in the root policy status, propagated policy status, or both indicating that a rollout is pending would be sufficient in my opinion.

For the PolicyAutomation front, I don't want us to get caught up on this since it's a very underutilized feature. For now, we could have an additional optional spec field on the PolicyAutomation such as `ignorePendingRollout`.

dhaiducek: Okay, that all sounds good to me, too. I've updated the enhancement with a point about setting the `remediationAction` to `inform`. I still need to check whether removing the status might work as a better alternative to the `lastEvaluated` status.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,282 @@
...
+- Reflect rollout status per cluster in the root policy for discoverability (whether a cluster is up-to-date or not)
+- (stretch) Implement the `RolloutStrategy` struct for policies, including:
+  - `RolloutStrategy`
+    - "All": all clusters at once
+    - "Progressive": one cluster at a time
+    - "ProgressivePerGroup": one group of clusters at a time
+  - `Timeout`: Maximum amount of time to wait for success before continuing the rollout
+  - `MandatoryDecisionGroups`: groups that should be handled before other groups
+  - `MaxConcurrency`: Concurrency during a progressive rollout
+- (stretch) Add an aggregated rollout status for the root policy status.
```
mprahl, line 39:
```suggestion
- Add an aggregated rollout status for the root policy status.
```

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+For this rollout enhancement, policy propagation by the `governance-policy-propagator` controller on the hub would take
+place in a multiple-pass approach:
+
+1. On the first pass following a root policy creation or update, the `governance-policy-propagator` controller on the
+   hub cluster will replicate all policies to managed cluster namespaces and are created regardless of the rollout
+   strategy specified, with the remediation action set to `inform`. This way, all policies are up-to-date and reporting
+   a compliance status based on the current version of the root policy. The aggregated rollout status on the root policy
+   is set to `Progressing`. If the remediation action on the root policy is already `inform`, the rollout status on each
+   cluster is set to `Progressing` and no second pass occurs. If the remediation action is `enforce`, the rollout status
```
gparvin, line 71: Does an inform policy that rolls out with some clusters being NonCompliant ever transition out of Progressing to Succeeded?  I think a clarification is needed on how an inform policy rollout can be successful and still identify NonCompliant clusters.

dhaiducek: It's an interesting question. I was thinking `inform` would follow the `enforce` behaviors, transitioning to `Succeeded` or `Failed` based on the compliance.

But for `inform`, I suppose we could transition the rollout to `Succeeded` once a status is returned regardless of compliance? It'd differ in behavior from `enforce`, but maybe that's okay?

gparvin: Yes, for inform I think as customers learn where problems "may" be they will improve inform policies to detect those issues.  Then the rollout will help identify those problem clusters and I don't think the rollout should fail or be limited because inform policies are working as expected.  Example: Initially we want our addon to just be available.  In v2 we learn we want to look out for terminating config policies....

mprahl: A user shouldn't define a progressive rollout for an inform policy since it will cause the policy to be enforced unless the ConfigurationPolicy has `remediationAction: informOnly` which in that event, the rollout would have no effect.

I think the rollout status should be set to `Succeeded` if any status is provided as Dale mentioned. Syntax errors in the ConfigurationPolicy would lead to a failed rollout status.

mprahl: The nice part of this is that even users that don't use progressive policy rollouts will get a single field to know if all policy templates have reported an up to date status for the cluster.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+   managed clusters returned from the `Placement` library (based on the user-configured rollout strategy) and sets the
+   rollout status of each returned cluster to `Progressing`, updating any `remediationAction` to `enforce` as defined in
+   the root policy.
+
+From here, work is picked up by the `governance-policy-framework` controllers on the managed clusters:
+
+1. The template sync controller of the framework updates the applicable `ConfigurationPolicy` objects like usual, which
+   triggers reevaluation from the `config-policy-controller`.
+2. Once the `ConfigurationPolicies` have a generation that matches the `lastEvaluatedGeneration` set on the policy
+   status, the status sync controller of the framework will set the rollout status to to `Succeeded` after all of the
```
mprahl, line 83: You should mention that this is set on the replicated policy on the managed cluster and hub.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+   `ConfigurationPolicies` have a compliant status or `Failed` if there is any non-compliant status.
```
mprahl, line 84: You should mention that the propagator will add the rollout status on the root policy after the propagated policy rollout status is updated.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+- **Summary**
+
+  - Add a `RolloutStatus` to the `Policy` status in the CRD.
+
+- **Details**
+
+  - The `RolloutStatus` would be added to reflect: rollout status on the replicated policy, per-cluster on the root
+    policy, and an aggregated status on the root policy. The aggregated status would specify: `Progressing` (this would
+    include `ToApply`), `Succeeded`, `Failed`, or (this would include `TimeOut`).
```
mprahl, line 100: I think `ToApply` means the the policy hasn't been enforced yet on the cluster.

`Progressing` would mean the policy has been enforced on the cluster but the rollout hasn't completed yet.

enhancements/sig-policy/99-policy-placement-strategy/README.md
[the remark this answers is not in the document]
mprahl: @dhaiducek I worry about the rollout strategy being defined on the PlacementBinding instead of the Policy because you can have multiple PlacementBindings per Policy. If multiple set rolloutStrategy, which one wins? Could a subset of clusters rollout `Progressive` and others rollout `ProgressivePerGroup`?

Also, the PlacementBinding can be used for multiple policies and it'd be annoying to have to use a separate PlacementBinding just because your policies have different rollout strategies.

Lastly, I think the UX around PlacementBinding is challenging. I find it much more intuitive to configure how my policy is deployed by setting it on the policy.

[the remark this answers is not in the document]
mprahl: In regards to selective policy enforcement, I personally think it should be ignored or an error returned when a rollout strategy is also defined.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+- The `Placement` library is relatively new and untested outside of its repo and this implementation leans heavily on
+  its logic. While it works in theory, there could be some tweaks/adjustments as development proceeds, lengthening time
+  for implementation. The phased approach intends to address this to make partial implementation feasible.
+
+## Design Details
+
+### Open Questions
+
+1. Should the rollout handler be instantiated with a dynamic watcher/cache instead of the controller client?
+2. Should `RolloutStatus` be reflected per-cluster in the root policy? (The rollout status would also be reflected in
```
mprahl, line 278: Yes to this.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+   each replicated policy.)
+3. Should the `RolloutStrategy` be implemented in the `Policy` CRD instead of `PlacementBinding`? (If that were the
+   case, it would also need to be added to `PolicySet`.)
+4. How should a policy handle when placement decision groups shift? (This implementation leans on the `Placement`
+   library to handle such shifts, only handling the clusters that it returns.)
+5. How is "soak time" (the minimum time for a policy to reach a successful state) and timeout (the maximum amount of
+   time to wait for a successful state), as defined in the placement enhancement, handled for Policies?
```
mprahl, line 285:
```suggestion
5. How is "soak time" (the minimum time for a policy to reach a successful state) and timeout (the maximum amount of
   time to wait for a state), as defined in the placement enhancement, handled for Policies?
```

enhancements/sig-policy/99-policy-placement-strategy/README.md
gparvin: Can I not roll out a policyset that contains inform and enforce policies and the inform policies would not be enforced? I should be able to put any random policies in a policyset with some enforce and some inform and it should be honored appropriately during rollout.  Enforcing some policies could have unintended consequences -- rebooting nodes, upgrading, etc

enhancements/sig-policy/99-policy-placement-strategy/README.md
imiller0: Is the set of "valid" status fields different on the root policy vs the replicated policies? It sounds like ToApply would be valid on a replicated policy but not the root. Similarly it sounds like progressing would be valid on the root but not the replicated set. If so it might help to define the set of valid values for root vs replicated.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+- **Summary**
+
+  - Only continue policy deployment once the previously deployed clusters show as "Compliant".
+
+- **Details**
+
+  - Update the `PlacementBinding` CRD to contain the `RolloutStrategy` struct. (See the
+    [`v1alpha1/RolloutStrategy`](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/types_rolloutstrategy.go))
+    Defaults to `All` if a strategy is not provided or the remediation action is `inform`.
+  - When the `remediationAction` is set to `enforce`, policies not currently being rolled out will be set to `inform` to
```
imiller0, line 211: Once set to enforce do they remain in enforce mode until a subsequent change to the root policy or do they revert to inform once compliant?

enhancements/sig-policy/99-policy-placement-strategy/README.md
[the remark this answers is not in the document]
imiller0: The multiple placement bindings question also seems to apply to the order/timing of when the enforce is done. If two bindings for the same policy have overlapping sets of clusters included both will (as I understand it) calculate decisiongroups independently. Any given cluster may be in group 1 of binding A and group 100 of binding B. Will it get enforced twice? If the max concurrency is 10, do you end up with 11 clusters being enforced due to the different groupings? etc.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Open Questions
+
+1. Should the rollout handler be instantiated with a dynamic watcher/cache instead of the controller client?
+2. Should `RolloutStatus` be reflected per-cluster in the root policy? (The rollout status would also be reflected in
+   each replicated policy.)
+3. Should the `RolloutStrategy` be implemented in the `Policy` CRD instead of `PlacementBinding`? (If that were the
+   case, it would also need to be added to `PolicySet`.)
+4. How should a policy handle when placement decision groups shift? (This implementation leans on the `Placement`
+   library to handle such shifts, only handling the clusters that it returns.)
```
imiller0, line 283: Does this risk clusters being skipped? Or gaps in the decision groups if clusters are unbound/rebound?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+5. How is "soak time" (the minimum time for a policy to reach a successful state) and timeout (the maximum amount of
+   time to wait for a successful state), as defined in the placement enhancement, handled for Policies?
+
```
imiller0, line 286: How does the rolloutStrategy work with policy dependency? Is it possible that if the two policies are bound using separate policybindings and the policies are modified simultaneously the group ordering could be such that rollout of one policy is entirely blocked until completion of all groups of the other policy?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+based on the success or failure of the update.
+
+### Design
+
+For this rollout enhancement, policy propagation by the `governance-policy-propagator` controller on the hub would take
+place in a multiple-pass approach:
+
+1. On the first pass following a root policy creation or update, the `governance-policy-propagator` controller on the
+   hub cluster will replicate all policies to managed cluster namespaces and are created regardless of the rollout
+   strategy specified, with the remediation action set to `inform`. This way, all policies are up-to-date and reporting
```
imiller0, line 68: If a policy is modified/updated by the user while a rollout is ongoing, does the process start over with all policies being set back to inform and rollout beginning with group 1 again?

enhancements/sig-policy/99-policy-placement-strategy/README.md
serngawy: I think it needs more clarification on how the RolloutStatus will move from ToApply->progressing->Succeeded/TimeOut. 
Here is some thoughts;
- **ToApply;** 
     1- when the child policy does not get created yet for the managedCluster
     2- when the child policy is created for the managedCluster **AND** the policy remediation is Inform and the policy in NonCompliant state.
      
- **Progressing;** 
    1- When the child policy is created for the managedCluster **AND** the policy remediation is Enforce and the policy in NonCompliant state.
      
- **Succeeded;** 
   1- When the child policy is created for the managedCluster and it is in Compliant state for either enforce/inform remediation

- **TimeOut;** 
   1- When the child policy is created for the managedCluster  **AND** the policy remediation is Enforce and the policy in NonCompliant state for longer time than it is defined in the rolloutStrategy->TimeOut

enhancements/sig-policy/99-policy-placement-strategy/README.md
serngawy, line 100: I don't think there is a Failed status cause the policy-controller will continue to enforce the policy to the managedCluster as long as it is in NonCompliant state. So status could be  **ToApply, Progressing, Succeeded, TimeOut.**

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+As a system administrator, I want policies to be rolled out only when the policy on previous clusters show as
+"Compliant" and control their concurrency and priority.
+
+- **Summary**
+
+  - Only continue policy deployment once the previously deployed clusters show as "Compliant".
+
+- **Details**
+
+  - Update the `PlacementBinding` CRD to contain the `RolloutStrategy` struct. (See the
```
serngawy, line 208: I think its better to define the RolloutStrategy under the Policy->Spec not under the PlacementBinding cause a policy can have multiple placementBinding so as a user should I repeated RolloutStrategy across all placementBindings? 
I guess defining the RolloutStrategy under PlacementBinding  will be challenging to handle.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+  - `MandatoryDecisionGroups`: groups that should be handled before other groups
+  - `MaxConcurrency`: Concurrency during a progressive rollout
+- Add an aggregated rollout status for the root policy status.
+
+### Non-Goals
+
+Any specialized behaviors outside of those provided by the `Placement` library and, by extension, the `DecisionStrategy`
+enhancement, that might require additional code other than that already provided/required by the `Placement` library.
+This includes the ability to roll back a rollout. A rollback should be implemented by the user by applying the previous
+version of the policy, and GitOps is the recommended method for this.
```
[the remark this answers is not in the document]
dhaiducek: @imiller0 That's true for objects on the managed cluster--this is referring specifically to the previous version of the policy itself on the hub rather than rolling back changes on the managed cluster.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: @serngawy True, but if the enforcement is blocked for some reason (like a missing API on the managed cluster or an invalid manifest), it may be good to reflect that in the root policy since there's no way for the rollout to progress?

dhaiducek: Good point, @mprahl--I wrapped `ToApply` in for simplicity but it would be clearer to use `ToApply`.

@imiller0 The `RolloutStatus` is an iota we're taking from the Placement library: https://github.com/open-cluster-management-io/api/blob/9675ab737d2086dd7094fc962c9617cc71ec7484/cluster/v1alpha1/helpers.go#L21-L38 but having these well defined for each the root and replicated policies would be good.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: They remain in `enforce` until the root policy is updated again.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,302 @@
...
+  `PlacementRule`-centric)
+- Leverage the `Placement` helper library to retrieve placement decisions
+- Reflect rollout status per cluster in the root policy for discoverability (whether a cluster is up-to-date or not)
+- Implement the `RolloutStrategy` struct for policies, including:
+  - `RolloutStrategy`
+    - "All": all clusters at once
+    - "Progressive": one cluster at a time
+    - "ProgressivePerGroup": one group of clusters at a time
+  - `Timeout`: Maximum amount of time to wait for success before continuing the rollout
+  - `MandatoryDecisionGroups`: groups that should be handled before other groups
```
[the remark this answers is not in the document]
dhaiducek: This would solely be priority ordering, but that's an interesting point to have a different timeout setting for the mandatory groups and the subsequent groups--that could be an update for the Placement library in the future.

`MandatoryDecisionGroups` is described in the `DecisionStrategy` enhancement here: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/64-placementStrategy#how-workload-applier-apis-will-benefits-from-using-placementstrategy

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: That's correct

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: I was working on updating this as you were reviewing 🙂 We came to the same conclusion and have updated the enhancement to specify the RolloutStrategy on the Policy rather than the PlacementBinding.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,515 @@
...
+cluster in turn has its policy updated to `enforce`. For each rollout strategy, this looks like:
+
+| Rollout Strategy      | Remediation Inform                      | Remediation Enforce                                                                                                                      |
+| --------------------- | --------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| `All`                 | Policy rolled out to all clusters as-is | Policy rolled out to all clusters as-is                                                                                                  |
+| `Progressive`         | (Same as `All`)                         | Policy applied initially to all clusters as `inform` and <br/>rollout enforces each cluster's policy once compliant (or timed out)       |
+| `ProgressivePerGroup` | (Same as `All`)                         | Policy applied initially to all clusters as `inform` and <br/>rollout enforces each cluster group's policy once compliant (or timed out) |
```
mprahl, line 76:
```suggestion
| `ProgressivePerGroup` | (Same as `All`)                         | Policy applied initially to all clusters as `inform` and <br/>rollout enforces a cluster group's policy and waits for compliant (or timed out) before moving to the next group |
```

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -0,0 +1,543 @@
...
+  apiVersion: policy.open-cluster-management.io/v1
+  kind: Policy
+  metadata:
+    name: policy-namespace.sample-policy
+    namespace: cluster1
+  spec:
+    disabled: false
+    remediationAction: inform # <-- Remediation action updated to prepare for rollout
+    rolloutStrategy:
+      type: ProgressivePerGroup
```
serngawy, line 266: I guess better to include the progressivePerGroup fields as well.
```disabled: false
remediationAction: inform # <-- Remediation action updated to prepare for rollout
rolloutStrategy:
  type: ProgressivePerGroup
  progressivePerGroup:
    mandatoryDecisionGroups:
    - groupName: dev-emea   (Same group name as defined in the placement.)
    timeout: None```

mprahl (approved):
This looks great. We'll work on adding the new requirements discussed today in a separate PR.

PR #101, opened 2023-09-19
Updates to the long-term policy compliance history design

mprahl:
See each commit for details.

### DIFF
--- a/enhancements/sig-policy/98-long-term-compliance-history/README.md
+++ b/enhancements/sig-policy/98-long-term-compliance-history/README.md
@@ -135,8 +135,8 @@ As previously mentioned, the compliance events will be stored in a relational da
 managed by Open Cluster Management and is limited to just Postgresql at the momement. The connection information for
 this database will be in a secret called `governance-policy-database` in the `open-cluster-management` namespace and
 will require either a key of `connectionURL` in the format of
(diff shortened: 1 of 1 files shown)


enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -332,6 +336,11 @@ was evaluated. With this additional information, the Status Sync controller can
 `/api/v1/compliance-events` `POST` request. If the spec couldn't be determined, a value of `unknown` will be recorded.
 The rest of the data can be found in the Kuberenetes `Event` object.
 
+To record a compliance event when a policy is deleted, the Spec Sync controller will record these compliance events
```
gparvin, line 339: Should it basically work the same way if the policy is disabled?

mprahl: Right, because when a policy is disabled on the hub, it deletes the propagated policy which causes the policy templates to be deleted on the cluster too.

mprahl: @gparvin , I think the timestamp and an adjustment to the message would suffice for now.

Since our main focus is currently on OCM policies, I think we are covered. We may need to adjust if we pursue the scraping idea.

gparvin (approved):

PR #103, opened 2023-10-02
[99][sig-policy] PolicySet rollout

dhaiducek:
Followup to:
- #99 

ref: https://issues.redhat.com/browse/ACM-6523

### DIFF
--- a/enhancements/sig-policy/99-policy-placement-strategy/README.md
+++ b/enhancements/sig-policy/99-policy-placement-strategy/README.md
@@ -21,6 +21,7 @@ Currently policies and subsequent updates are pushed out to clusters en masse ba
 been bound. The new `DecisionStrategy` field in the `Placement` API creates cluster groupings that controllers can
 leverage to have segmented, configurable rollouts of resources. This will aid use cases where high availability may be a
 priority or a set of test clusters should receive and verify the updates before the remaining clusters get updated.
+Additionally, adding `PolicySet` rollouts will allow users to slice updates by cluster rather than by `Policy`.
 
 ### Goals
 
@@ -39,6 +40,7 @@ priority or a set of test clusters should receive and verify the updates before 
   - `MandatoryDecisionGroups`: groups that should be handled before other groups
   - `MaxConcurrency`: Concurrency during a progressive rollout
 - Add an aggregated rollout status for the root policy status.
+- Implement rollout for `PolicySet`, allowing rollouts to proceed on a per-cluster basis across all policies in the set
 
 ### Non-Goals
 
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -496,6 +644,55 @@ the updates, and continue enforcement rollout only when the policy on previous c
...
+#### Development Phase 4
+
+- **Summary**
+
+  - Implement rollouts for policy sets.
+
+- **Details**
+
+  - Add an `OrderPolicies` boolean and the `RolloutStrategy` struct to the spec of the `PolicySet`.
+  - `OrderPolicies` signals to the `governance-policy-propagator` controller whether to inject a dependency on the
```
gparvin, line 656: Could an injected dependency between policies conflict with an actual dependency?

dhaiducek: The existing dependency would be added to the list of dependencies on the replicated policy (rather than a conflict per se), so both would block the policy until they're fulfilled. The dependencies are one point of complexity here, but for this it'd become a union of the dependencies.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -496,6 +644,55 @@ the updates, and continue enforcement rollout only when the policy on previous c
...
+    previous policy in the set when propagating policies to the managed cluster namespaces.
+  - Logic would need to be added to the `governance-policy-propagator` controller to detect when a policy is bound to a
+    placement via a policy set to roll out policies in a set per cluster. The `GetClusterRolloutStatus` function should
+    be updated to consider all policies in the policy set for a particular cluster so that the library returns clusters
+    to roll out to based on the entire set of policies (rather than the individual policy). This flow looks like:
+    - Policies in the set across all managed cluster namespaces are set to `remediationAction: inform` upon update, as
+      specified in this enhancement, with the policy set overriding any `rolloutStrategy` specified in the individual
+      policy.
+    - When the policy set rollout is signaled by the `policy.open-cluster-management.io/last-rollout` annotation, the
+      propagator identifies whether a policy was bound to a placement via a policy set. From there:
```
gparvin, line 666: And the policy should only have the one placement defined.  That is the placement on the policyset in this case.  So if the policy is in another policyset too --- that 2nd policyset must use the same placement.  What if that policyset doesn't specify the rollout strategy?

dhaiducek: If the second policy set doesn't specify a rollout, I think the rollout would take precedence if it specifies the same cluster. But I'm not sure why you'd have the same policy in two separate sets with the same placement?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -319,6 +322,149 @@ the updates, and continue enforcement rollout only when the policy on previous c
...
+- **Summary**
+
+  - Add an option in the policy set to specify an `orderPolicies` boolean, signaling to the
+    `governance-policy-propagator` controller to automatically inject policy dependencies down the list of policies in
+    the policy set when replicating policies (i.e. policy 1 is a dependency for policy 2, policy 2 is a dependency for
+    policy 3, and so on).
+  - Add `rolloutStrategy` and `rolloutStatus` to the `PolicySet`. For a policy set, the rollout is successful for a
+    cluster only after all policies in the set become compliant.
+  - The `rolloutStatus` would be an aggregated status across all clusters and policies. The aggregated root policy
+    status would be: `Progressing` (this would include `ToApply`), `Succeeded`, or `Failed`.
```
yiraeChristineKim, line 339: What is "ToApply"

yiraeChristineKim: I just see that policy is waiting to be enforced.. how about "ready", ToApply is ambiguous to me. but if you ok I am good

dhaiducek: "ToApply" is one of the available rollout statuses: https://github.com/open-cluster-management-io/api/blob/c2b66fa4b2df8316b1165899b5eea31d237a4c70/cluster/v1alpha1/helpers.go#L25-L26

How they're used for policy is outlined earlier in this doc:
https://github.com/open-cluster-management-io/enhancements/blob/dffc91e436130c695b6191c4ee485b8e620cd21e/enhancements/sig-policy/99-policy-placement-strategy/README.md?plain=1#L114-L121

dhaiducek: It is defined in the Placement library--it's not something we control directly.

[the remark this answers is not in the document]
dhaiducek: Hmm. Interesting question. I could see it either way but I'm thinking it'd still be marked as Failed since the rollout failed on some clusters. For something like ProgressivePerGroup, setting MaxFailures would give a tolerance to allow the rollout to move on to the next group in the case of flaky cluster(s), so the aggregated status would stay in `Progressing` until then end, when it would wind up `Failed`.

yiraeChristineKim (approved):

PR #104, opened 2023-10-17
Add the proposal of the new feature - MultipleHubs.

xuezhaojun:


qiujian16:
@xuezhaojun I wonder if we should merge this, anything needs to be updated?

xuezhaojun:
> @xuezhaojun I wonder if we should merge this, anything needs to be updated?

Thanks for reminding me, the PR is updated with the actual implementation, we can merge it now.

PR #105, opened 2023-12-20
Update the long-term policy compliance history design

mprahl:
This stems from rethinking assumptions when doing the development.

### DIFF
--- a/enhancements/sig-policy/98-long-term-compliance-history/README.md
+++ b/enhancements/sig-policy/98-long-term-compliance-history/README.md
@@ -67,7 +67,8 @@ confusing if a separate database on the same Postgresql database be used.
 | Column     | Description                                                         |
 | ---------- | ------------------------------------------------------------------- |
 | id         | an internal primary key                                             |
-| name       | a string column for the name of the parent replicated policy        |
+| name       | a string column for the name of the parent root policy              |
+| namespace  | a string column for the namespace of the parent root policy         |
 | categories | an array column of the sorted categories associated with the policy |
 | controls   | an array column of the sorted controls associated with the policy   |
 | standards  | an array column of the sorted standards associated with the policy  |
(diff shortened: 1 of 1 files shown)


enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -81,15 +81,15 @@ string/text column, then those could be used instead of a substring query.
...
-| spec_hash        | a string column of a SHA1 hash of the spec field of the replicated policy                                             |
+| Column    | Description                                                                                                                        |
+| --------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| id        | an internal primary key                                                                                                            |
+| kind      | a string column of the policy Kind (e.g. ConfigurationPolicy)                                                                      |
+| api_group | a string column of the policy group (e.g. policy.open-cluster-management.io)                                                       |
+| name      | a string column for the name of the policy                                                                                         |
+| namespace | an optional string column for the namespace of the policy. This will not be set for native OCM policies to avoid data duplication. |
```
JustinKuli, line 90: Not set on OCM policies because it would just always be the cluster namespace?

enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -67,7 +67,7 @@ confusing if a separate database on the same Postgresql database be used.
 | Column     | Description                                                         |
 | ---------- | ------------------------------------------------------------------- |
 | id         | an internal primary key                                             |
-| name       | a string column for the name of the parent replicated policy        |
+| name       | a string column for the name of the parent root policy              |
```
JustinKuli, line 70: Just the name of the root policy itself, not the `<namespace>.<name>` thing the propagator makes?

Just considering that on the managed cluster, I think I could have two different policies: say `ns-a.important-roles` and `ns-b.important-roles`, which could be different. It probably wouldn't be helpful to lose that namespace information when the event gets into this history?

enhancements/sig-policy/98-long-term-compliance-history/README.md
mprahl: Exactly.

JustinKuli (approved):

PR #107, opened 2024-01-04
Policy Addon: Use JSONB for all JSON columns and update query arguments

mprahl:
This removes the need for spec_hash and this makes it so that the
database IDs can be used to save on network bandwidth.

This also renames the query arguments to be more intuitive.


JustinKuli (commented):
JSONB is another postgres-specific feature, right? Which I'm fine with, I just want to note that we are putting in nice features that rely on a more-specific kind of backing database.

mprahl:
> JSONB is another postgres-specific feature, right? Which I'm fine with, I just want to note that we are putting in nice features that rely on a more-specific kind of backing database.

That's right. We don't have a business need to support more than Postgres so I think it makes sense to leverage Postgres features that do what we want.

JustinKuli (approved):

mprahl:
@JustinKuli could you please reapprove this? I had to rebase the branch.

PR #110, opened 2024-01-17
Minor updates to the long-term compliance history design

mprahl:


enhancements/sig-policy/98-long-term-compliance-history/README.md
```diff
@@ -484,10 +478,6 @@ As a policy user, I require that compliance history events contain additional me
 diff between what is declared in the policy and what is on the cluster to know what the exact violation was. This will
 help in forensic investigations after a breach has occurred of a managed cluster.
 
-#### Story 5
-
-As a policy user, I'd like to import my existing compliance history to the new compliance events database.
```
JustinKuli, line 489: I liked this story. The commit message mentions that this history might be inaccurate, can you explain that? I still feel like it could be better than nothing...

mprahl: The things that would be missing is:
* We don't know the parent policy's info at the time of the event. We could assume the current parent's info, which is a safe assumption.
* We don't know the policy's spec at the time of the event, so we can't assume it hasn't changed.
* If the history goes back a long time, it implies to the user that it's accurate but it'd be missing things like deleted policies, deleted clusters, changes in policies, and etc.

I was thinking that we'd defer this until there is a specific ask and discuss if the user asking for it is okay with these caveats before we implement historical import.

PR #116, opened 2024-04-17
Update OperatorPolicy design to v1beta2

JustinKuli:


JustinKuli:
/hold

We might not bump the version, so details here might change. We don't necessarily need to provide a seamless experience for this change.

JustinKuli:
I'll open a different PR, since a lot of details are changing. I don't want to confuse anyone.

PR #118, opened 2024-04-20
Update the progressive policy rollout design

mprahl:
Note that since there's so much change, it's best to treat this like a new document when reviewing the changes (e.g. look at the markdown file directly instead of the diff).

Part of the proposal applies to all of Open Cluster Management.

### DIFF
--- a/enhancements/sig-policy/99-policy-placement-strategy/README.md
+++ b/enhancements/sig-policy/99-policy-placement-strategy/README.md
@@ -1,4 +1,4 @@
-# Policy placement strategy
+# Progressive Policy Rollout
 
 ## Release Signoff Checklist
 
@@ -11,43 +11,38 @@
 
 ## Summary
 
-Following from the [`DecisionStrategy` enhancement](../../sig-architecture/64-placementStrategy/README.md) in the
-`Placement` API, policies can leverage this new logic to have a configurable and systematic way to roll out policy
-updates to clusters.
+Open Cluster Management policies will support the
+[rollout strategy API](https://open-cluster-management.io/concepts/placement/#rollout-strategy) with the additional
+strategy of `manualPerGroup`. This enhancement allows control over the conditions for when a new policy or update to an
+existing policy is deployed to the Kubernetes fleet in customizable increments.
 
 ## Motivation
 
-Currently policies and subsequent updates are pushed out to clusters en masse based on the placement to which it has
-been bound. The new `DecisionStrategy` field in the `Placement` API creates cluster groupings that controllers can
-leverage to have segmented, configurable rollouts of resources. This will aid use cases where high availability may be a
-priority or a set of test clusters should receive and verify the updates before the remaining clusters get updated.
-Additionally, adding `PolicySet` rollouts will allow users to slice updates by cluster rather than by `Policy`.
+Currently, new policies and updates to existing policies instantly deploy to the clusters selected by the bound
+placement. To do this progressively requires copies of policies or complex templating, and manual management of the
+placements. In complex environments, this can be impractical. The
+[decisionStrategy](https://open-cluster-management.io/concepts/placement/#decision-strategy) placement API allows for
+defining groups of the selected clusters. The
+[rollout strategy API](https://open-cluster-management.io/concepts/placement/#rollout-strategy) defines the conditions
+for how a policy change deploys to the environment.
 
 ### Goals
 
-- Make `Placement` the primary API for placement (currently the governance propagator is somewhat
-  `PlacementRule`-centric)
+- Progressive rollout of new policies and updates to policies based on the
+  [decisionStrategy](https://open-cluster-management.io/concepts/placement/#decision-strategy) field in the `Placement`
+  API
+- Support clusters added to the placement during a rollout
+- Reflect the rollout status in the root and replicated policies for discoverability
+- Allow for manual approval of each decision group
 - Leverage the `Placement` helper library to retrieve placement decisions
-- Reflect rollout status per cluster in the root policy for discoverability (whether a cluster is up-to-date or not)
-- Implement the `RolloutStrategy` struct for policies, including:
-  - `RolloutStrategy`
-    - "All": all clusters at once
-    - "Progressive": one cluster at a time
-    - "ProgressivePerGroup": one group of clusters at a time
-  - `ProgressDeadline`: Maximum amount of time to wait for status before timing out
-  - `MinSuccessTime`: Minimum amount of time to wait before proceeding (i.e. "soak" time)
-  - `MaxFailures`: Tolerance of percentage or number of clusters that can fail and still continue the rollout
-  - `MandatoryDecisionGroups`: groups that should be handled before other groups
-  - `MaxConcurrency`: Concurrency during a progressive rollout
-- Add an aggregated rollout status for the root policy status.
-- Implement rollout for `PolicySet`, allowing rollouts to proceed on a per-cluster basis across all policies in the set
 
 ### Non-Goals
 
-Any specialized behaviors outside of those provided by the `Placement` library and, by extension, the `DecisionStrategy`
-enhancement, that might require additional code other than that already provided/required by the `Placement` library.
-This includes the ability to roll back a rollout. A rollback should be implemented by the user by applying the previous
-version of the policy, and GitOps is the recommended method for this.
+- **Rollbacks** - Reverting a policy does not necessarily rollback the change. It's up to the user to determine how to
+  rollback in the event of a rollout failure.
+- **Group Rollouts** - Rollouts of a group of policies is not a goal at this moment due to technical and user experience
+  complexity. The policy templates in the `policy-templates` array within the policy (e.g. multiple
+  `ConfigurationPolicy`) does rollout at as group.
 
 ## Proposal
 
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -11,43 +11,38 @@
...
 
 ### Non-Goals
 
-Any specialized behaviors outside of those provided by the `Placement` library and, by extension, the `DecisionStrategy`
-enhancement, that might require additional code other than that already provided/required by the `Placement` library.
-This includes the ability to roll back a rollout. A rollback should be implemented by the user by applying the previous
-version of the policy, and GitOps is the recommended method for this.
+- **Rollbacks** - Reverting a policy does not necessarily rollback the change. It's up to the user to determine how to
+  rollback in the event of a rollout failure.
+- **Group Rollouts** - Rollouts of a group of policies is not a goal at this moment due to technical and user experience
```
yiraeChristineKim, line 43: Group of policy means policy set?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+`metadata.ownerReferences` fields set for storing state.
+
+The `spec.decisionGroups` field is set by the user to indicate which groups are approved. These groups directly match
+those in the [decisionStrategy](https://open-cluster-management.io/concepts/placement/#decision-strategy) field in the
+`Placement` API. If the user wants a subset of the groups to be automatically approved, they would add the groups to the
+policy's `spec.rolloutStrategy.manualPerGroup.mandatoryDecisionGroups` array.
+
+If the user wants the approval to be conditional on a policy version, the user must set `spec.approvalsForVersion` to a
+value that matches the `policy.open-cluster-management.io/version` annotation on the root policy. By default, the
+approvals apply to all policy versions.
```
yiraeChristineKim, line 145: apply to lastest?

yiraeChristineKim: all versions little confusing me

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+### Status
+
+#### Replicated Policy Status
+
+A replicated policy is the policy copy in the managed cluster namespace. Each replicated policy will have a rollout
+status defined as follows:
+
+| Status          | Description                                                                                           |
+| --------------- | ----------------------------------------------------------------------------------------------------- |
+| `ToApply`       | The cluster is waiting for the new version of the policy but the old version is still applied.        |
```
yiraeChristineKim, line 159: hmm to me, await , ready  are better` to apply` not clear to me

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+
+```yaml
+apiVersion: v1beta1
+kind: Rollout
+metadata:
+  name: policy-<same as root policy>
+  namespace: <same as root policy>
+spec:
+  approvalsForVersion: "2.0" <-- optional if the user wants to tie the approval to a specific policy "version".
+  decisionGroups:
```
yiraeChristineKim, line 118: So this is the order of groups?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+  namespace: policy-namespace
+spec:
+  disabled: false
+  remediationAction: enforce
+  rolloutStrategy:               # <-- New field
+    type: ProgressivePerGroup
+    progressivePerGroup:
+      minSuccessTime: 5m
+        progressDeadline: 10m
+        maxFailures: 2%
```
yiraeChristineKim, line 82: Do we have some like fail-fast? wherever is fail, stop progressing

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+
+1. The root policy is updated.
+1. The updated policy applies to all clusters in the `dev` group. These clusters have the `rolloutStatus` of
+   `InProgress`. The remaining clusters have the `rolloutStatus` of `ToApply` and keep the previous policy version. The
+   root policy `status.rolloutStatus` value is `InProgress`.
+1. All the clusters in the `dev` group become compliant. Their `rolloutStatus` is set to `Succeeded`.
+1. The updated policy applies to all clusters in the `stage` group. These clusters have the `rolloutStatus` of
+   `InProgress`.
+1. All the clusters in the `stage` group become compliant. Their `rolloutStatus` is set to `Succeeded`.
+1. The updated policy applies to all clusters in the `prod` group. These clusters have the `rolloutStatus` of
```
yiraeChristineKim, line 316: Individual status of group only appears on `kind: rollout?`

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+   `InProgress`. The remaining clusters have the `rolloutStatus` of `ToApply` and keep the previous policy version. The
+   root policy `status.rolloutStatus` value is `InProgress`.
+1. One cluster in the `dev` group stays noncompliant for the duration of the `spec.rolloutStrategy.progressDeadline` of
+   10 minutes. The root policy `status.rolloutStatus` value is `Failed`. The `dev` group still keeps the policy version
+   of the failed rollout. All remaining groups stay with the last successful version.
+
+#### Updated Policy With manualPerGroup in Order
+
+1. The root policy is updated.
+1. All clusters have the `rolloutStatus` of `ToApply` and keep the previous policy version. The root policy
```
yiraeChristineKim, line 334: So if user set both dev and prod to true at the same time. The order would be dev -> prod?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+#### Common Situations
+
+##### A New Cluster Is Added During a Rollout
+
+When a new cluster is added, it receives the same policy as the others in the same group currently have. For example, if
+there is a policy update and a cluster is added to the `stage` group while the `dev` group is still `InProgress`, the
+new cluster would receive the last successful policy version, if any, until the rollout progressed to the `stage` group.
+
+If the cluster is added to an earlier group than the group that is `InProgress`, the rollout switches back to that group
+and waits for the cluster to rollout before resuming progress at the point previous to the new cluster being added.
```
yiraeChristineKim, line 381: My understanding is. a newbie cluster is in progress as soon as it comes. And then the previously progressed cluster restarts its progress.. is this correct?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+
+##### Retrying a Failed Rollout
+
+This is only necessary if `progressDeadline` is used since this is the only way a rollout can fail. Otherwise, the
+rollout indefinitely waits for the clusters to become compliant.
+
+In this situation, the first option is to update the policy spec in some way which would completly restart the rollout.
+
+The second option is to use the `spec.retryRollout.rolloutUID` field on the `Rollout` object. When set to a UID that
+matches `status.rolloutUID`, the Governance Policy Propagator will restart the rollout and create a new rollout UID.
```
yiraeChristineKim, line 402: Are we going to save snaps of policy + rollout?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+kind: Policy
+metadata:
+  name: sample-policy
+  namespace: policy-namespace
+spec:
+  disabled: false
+  remediationAction: enforce
+  rolloutStrategy:
+    ignoreClusterRolloutStatus:
+      matchExpressions:
```
yiraeChristineKim, line 446: Why use this instead of simple group name?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -648,80 +686,29 @@ proceed across the set on a per-cluster basis and, optionally, in the order spec
...
-  for implementation. The phased approach intends to address this to make partial implementation feasible.
+- The `Placement` library does not support the `manualPerGroup` rollout strategy.
 
 ## Design Details
 
 ### Open Questions
 
-1. Should the per-cluster status on the root policy be grouped similar to how they're grouped in the
-   `PlacementDecisions`?
+1. Should the `spec.rolloutStrategy.ignoreClusterRolloutStatus` field be contributed to the rollout strategy API?
```
yiraeChristineKim, line 711: ignoreClusterRolloutStatus -> is this act like ignore cluster group failure ? in other words, it is not calculated in maxfailure?  correct my understand?

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+spec:
+  disabled: false
+  remediationAction: enforce
+  rolloutStrategy:               # <-- New field
+    type: ProgressivePerGroup
+    progressivePerGroup:
+      minSuccessTime: 5m
+        progressDeadline: 10m
+        maxFailures: 2%
+        ...
```
dhaiducek, line 83:
```suggestion
      minSuccessTime: 5m
      progressDeadline: 10m
      maxFailures: 2%
      ...
```

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+
+This is to account for complex environment requirements that may not be conducive to automated rules. For example, a
+group of clusters that handle credit card machines are in a change freeze between the United States Thanksgiving holiday
+and Christmas due to peak shopping, however, other clusters may have low usage during that time and is an ideal period
+for deployments.
+
+To control these groups, the user will interact with a new Kubernetes resource of `Rollout` as shown below.
+
+```yaml
+apiVersion: v1beta1
```
dhaiducek, line 111: Should this have an API group with it?
```suggestion
apiVersion: cluster.open-cluster-management.io/v1beta1
```

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+kind: Rollout
+metadata:
+  name: policy-<same as root policy>
+  namespace: <same as root policy>
+spec:
+  approvalsForVersion: "2.0" <-- optional if the user wants to tie the approval to a specific policy "version".
+  decisionGroups:
+    - groupName: dev
+      rolloutApproved: true <-- user sets this to true when ready
+    - groupName: uat:
```
dhaiducek, line 121:
```suggestion
    - groupName: uat
```

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+status defined as follows:
+
+| Status          | Description                                                                                           |
+| --------------- | ----------------------------------------------------------------------------------------------------- |
+| `ToApply`       | The cluster is waiting for the new version of the policy but the old version is still applied.        |
+| `Progressing`   | The new version of the policy has been deployed but the rollout status is not yet known.              |
+| `Succeeded`     | The policy has confirmed it is compliant.                                                             |
+| `Failed`        | The policy is still noncompliant after `progressDeadline` has elapsed.                                |
+| `TimeOut`       | Time has passed beyond the timeout specified in `spec.rolloutStrategy` without a returned compliance. |
+| `NotApplicable` | The policy was applied to the cluster after the rollout completed.                                    |
```
dhaiducek, line 164: I get that this new status is trying to say it's N/A for the rollout, but that might be confusing to me as a user to see an N/A status and understanding what that means. It seems to me adding a new cluster and then having it fail would also use these other statuses because the rollout (even if the original rollout had ended long ago) has now failed.

Like if I added an entire batch of clusters that were spread across groups, I think I'd want to see whether that "mini" rollout had succeeded.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: This enum has been out there for a little bit, so I suspect there may be opposition to updating this particular one.

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: > Group of policy means policy set?

Correct

enhancements/sig-policy/99-policy-placement-strategy/README.md
dhaiducek: > Do we have some like fail-fast? wherever is fail, stop progressing

The default is to stop rollouts on failures. These tuning points allow it to continue within customizable thresholds.

enhancements/sig-policy/99-policy-placement-strategy/README.md
```diff
@@ -60,424 +55,481 @@ recommended to review it before proceeding.
...
+This control is separate from the policy's `spec.rolloutStrategy` because the `status` field is required for the
+technical implementation as described in another section. Having it separate also allows control of the rollout in a
+console or CLI even if the policy is managed by GitOps. Additionally, if the policy applies to multiple hubs, the groups
+may be different on those hubs.
+
+The name, with a `policy-` prefix, and the namespace of the `Rollout` needs to be the same as the root policy it refers
+to. This object will be automatically created by the Governance Policy Propagator with only the `status` and
+`metadata.ownerReferences` fields set for storing state.
+
+The `spec.decisionGroups` field is set by the user to indicate which groups are approved. These groups directly match
```
dhaiducek, line 138:
```suggestion
The `spec.decisionGroups` field is set by the user to indicate which groups are approved, and the group names are discoverable in the `Rollout` status. These groups directly match
```

dhaiducek (commented):
Looks really good! Thanks for the updates! I have a couple questions/comments.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: You're right. I'll update that.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: The order here doesn't actually matter. I prefer using a map in these cases but most Kubernetes APIs use lists with objects with `name` keys, so this copies that. The order in the placement matters in terms of priority but are not dependencies like in `progressivePerGroup`.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: We don't have statuses per group but only rollout statuses per cluster and a summary rollout status to reflect all clusters.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: Yes, that's correct.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: Not quite. The new cluster always gets a policy. Either the last successful version or the latest version. Which one is dependent on if the cluster is in a group that is being rolled out to or has been rolled out to.

For example, if you have dev, stage, and prod groups and the rollout is on the stage group.

If the new cluster is a dev cluster, the rollout goes back to dev and applies the latest policy to the new cluster before going back to stage.

If the new cluster is a prod cluster, the new cluster gets the last successful policy until the rollout is on the prod group. Then the new cluster gets thte latest policy.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: Could you please rephrase the question?

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: You might have prod clusters that are flaky but you still want to apply the policy to them without ignoring the status of all prod clusters.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: Essentially, the cluster gets the new policy version when its group is being rolled out to, but its status is ignored and not counted towards failure or success.

enhancements/sig-policy/99-policy-placement-strategy/README.md
mprahl: I'm trying to limit the rollout to new policy versions rather than new clusters to make it conceptually simpler for the user and for implementation.

For your scenario, I think the workaround would be to set the policy as inform and then flip it to enforce after the clusters have been added. Then a new rollout would trigger since the generation of the policy changed.

dhaiducek: In that case, maybe the status should be "NewCluster". I think that'd make more sense to me.

yiraeChristineKim (approved):

PR #119, opened 2024-05-28
Update OperatorPolicy design

JustinKuli:
Rename `statusConfig` to `complianceConfig`, and use options `Compliant` and `NonCompliant`, which might be more clear to users.


mprahl (approved):
/hold for other reviews

PR #122, opened 2024-07-03
Add the design for expanding hub templates access

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-12129

### DIFF
--- a/enhancements/sig-policy/122-expanded-hub-templates-access/README.md
+++ b/enhancements/sig-policy/122-expanded-hub-templates-access/README.md
@@ -1,0 +1,310 @@
+# Expand Hub Template Access on Policies
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Open Cluster Management policy hub templates are limited to referencing objects in the same namespace as the root
+policy. The proposal is to add an additional field of `spec.hubTemplateOptions.serviceAccountName` that must reference a
+service account in the same namespace as the root policy. The hub templates would execute using that service account.
+The user is responsible for providing `list` and `watch` permissions on the objects that the hub templates reference for
+that service account.
+
+## Motivation
+
+An Open Cluster Management policy can contain hub templates that get executed as replicated policies are created in each
+managed cluster namespace on the hub. To prevent privilege escalation and data leakage, there is a current restriction
+that only the `ManagedCluster` object and objects in the same namespace as the root policy can be referenced. This is
+limiting in situations such as copying a `Secret` from another namespace to the managed cluster. The current workaround
+is to have a policy applied to `local-cluster` (Hub) that copies the `Secret` to the root policy namespace. Workarounds
+for cluster-scoped objects or objects that have side effects are more complicated.
+
+### Goals
+
+1. A user can specify a service account for hub templates to access objects outside of the root policy namespace.
+1. The current event-driven nature of hub templates is retained with a custom service account.
+1. Tokens for the service account have an expiration of 2 hours.
+
+### Non-Goals
+
+1. Allowing service accounts outside of the root policy namespace.
+
+## Proposal
+
+### Design
+
+Define a new field at `spec.hubTemplateOptions.serviceAccountName` to specify a service account that exists in the root
+policy namespace. When specified, the service account is used instead of the service account of the Governance Policy
+Propagator for hub template execution. Additionally, no namespace or cluster-scoped restrictions will be enforced by the
+Governance Policy Propagator, and instead, it will rely on the Kubernetes API performing authorization checks of the
+service account. Note that due to the event-driven nature of hub templates, the service account must have `list` and
+`watch` permissions for the objects that the hub templates reference.
+
+Note that for backwards compatibility, if the namespace is not specified in a hub template object lookup, the root
+policy namespace will be the default, however, the service account must still have the appropriate permissions in the
+root policy namespace to perform the lookup.
+
+See an example of such a policy below:
+
+```yaml
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: test-config
+  namespace: test-policies
+spec:
+  disabled: false
+  remediationAction: enforce
+  hubTemplateOptions:
+    serviceAccountName: testsa
+  policy-templates:
+    - objectDefinition:
+        apiVersion: policy.open-cluster-management.io/v1
+        kind: ConfigurationPolicy
+        metadata:
+          name: test-config
+        spec:
+          remediationAction: enforce
+          severity: high
+          object-templates:
+            - complianceType: musthave
+              objectDefinition:
+                apiVersion: v1
+                data:
+                  field1: '{{hub fromConfigMap "some-other-namespace" "test-configmap" "test-field1" hub}}'
+                  field2: '{{hub fromConfigMap "some-other-namespace" "test-configmap" "test-field1" hub}}'
+                kind: ConfigMap
+                metadata:
+                  name: my-configmap
+                  namespace: test
+```
+
+### User Stories
+
+#### Story 1
+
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+[kubernetes-dependency-watches](https://github.com/stolostron/kubernetes-dependency-watches) library representing the
+replicated policy. The values are `types.NamespacedName` values of the service account used. Note that for the default
+template resolver, we will use the special name of `<default>` with an empty namespace. In the `Reconcile` method, if
+the service account changes, the `DynamicWatcher` of the old service account is retrieved from `r.TemplateResolvers` and
+`resolver.DynamicWatcher.RemoveWatcher` is called. This should be done regardless of if hub templates are detected since
+the update of the service account name may have removed hub templates as well. Note that this `policyToServiceAccount`
+map can also be used to clean up watches when a replicated policy is deleted since the original replicated policy spec
+is no longer available.
+
+The other complication is the service account must be periodically refreshed before expiration. This can be handled when
```
yiraeChristineKim, line 179: Correct me if I am wrong, Based on my understanding, 100 service account saved, then run 100 goroutine to watch token expire?

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### Story 1
+
+As a policy user, I would like to specify hub templates that reference objects outside of the root policy namespace so
+that I can stop defining additional policies on `local-cluster` to sync the required information to the root policy
+namespace.
+
+### Implementation Details/Notes/Constraints
+
+Although the user experience is rather simple, there's complexity in the implementation. One aspect is that the number
```
yiraeChristineKim, line 99: Are you going to share this in tech share? Hard to understand 100% sorry too complex

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
mprahl: Yes, if there are 100 different `serviceAccountName` values, then there'd be 100 goroutines. It should be quite cheap since it just essentially sleeps until near the expiration, then it refreshes the token, and sleeps again.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
mprahl: Sure, I can also do that during the daily scrum next week. If you have specific questions, I can answer those.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
yiraeChristineKim: Oh so token expiration is 2 hour so during 2 hour, it sleeps

yiraeChristineKim (approved):

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+of service accounts is not known at startup and can constantly change. It is also a bit complex to prevent duplication
+of Kubernetes API watches, underlying TCP connections, and tokens if multiple policies reference the same service
+account.
+
+Currently, the Governance Policy Propagator only uses a single `TemplateResolver` client from the
+[go-template-utils](https://github.com/stolostron/go-template-utils) library and sets restrictions based on the root
+policy namespace. That will still remain the default, however, we'll need exactly one `TemplateResolver` client per
+service account. To do this, I recommend having a new `sync.Map` field named `TemplateResolvers` on the
+`ReplicatedPolicyReconciler` struct and each key would be a `NamespacedName` from `k8s.io/apimachinery/pkg/types`
```
JustinKuli, line 108: I think we have used this pattern of a `sync.Map` to things with individual mutexes before, but it feels a little bit clunky to me. I wonder if any other teams have patterns they use for things like this, just to compare to.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+	break
+}
+
+// TODO: Code to get a token and instantiate a TemplateResolver instance for the service account and
+// set the values on `resolver`.
+```
+
+There should be a goroutine that periodically cleans up entries in `TemplateResolvers` that are for service accounts no
+longer referenced by any root policy. This is simpler than managing a reference count for each service account.
```
JustinKuli, line 164: Maybe if there will be a goroutine to clean things up anyway, there should just be that goroutine fully managing all of the entries, and we use channels to synchronize work. Just a thought - I'm not sure exactly what it would look like.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
mprahl: We do use this pattern and I agree it's a bit clunky but the `sync.Map` is required due to the dynamic nature of the keys since they can change at runtime and can come and go. The alternative is a global lock on a normal map but `sync.Map` is optimized for multiple reads and few writes:
https://pkg.go.dev/sync#Map

One thing we could try is [sync.OnceValue](https://pkg.go.dev/sync#OnceValue) instead of a lock in the struct for `sync.Map` but it uses locks under the hood so we may be better off managing the lock ourselves for efficiency.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
mprahl: It's an interesting idea but I think it'd be more complex than what is suggested in the design since you'd need a channel to request the template resolver for the SA and then a channel to receive it. Each reconcile call would somehow need to pick the correct channel to listen on to get the correct template resolver. I think a `sync.Map` and a `sync.RWMutex` to synchronize the creation but read concurrently is easier to understand and likely more efficient.

Let me know if you had something else in mind than what I pictured!

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+Define a new field at `spec.hubTemplateOptions.serviceAccountName` to specify a service account that exists in the root
+policy namespace. When specified, the service account is used instead of the service account of the Governance Policy
+Propagator for hub template execution. Additionally, no namespace or cluster-scoped restrictions will be enforced by the
+Governance Policy Propagator, and instead, it will rely on the Kubernetes API performing authorization checks of the
+service account. Note that due to the event-driven nature of hub templates, the service account must have `list` and
+`watch` permissions for the objects that the hub templates reference.
+
+Note that for backwards compatibility, if the namespace is not specified in a hub template object lookup, the root
+policy namespace will be the default, however, the service account must still have the appropriate permissions in the
```
gparvin, line 51: So in the example, the following function:
```
                  field3: '{{hub fromConfigMap "" "test-configmap" "test-field1" hub}}'
```
would default to the test-policies namespace and would require list and watch permissions for ConfigMaps in the test-policies namespace to be in the roles for the serviceaccount.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+			// This really does nothing but this satisfies the linter that cancel is called.
+			deadlineCtxCancel()
+		}
+	}()
+
+	return tokenFilePath, nil
+}
+```
+
+### Risks and Mitigation
```
gparvin, line 274: There could be some cases where customers have policies deployed to namespaces that have serviceaccounts already and they could take advantage of these permissions to gain some extra privileges they didn't have before.  The best practice of not mixing policies with workloads would be the mitigation.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+### Version Skew Strategy
+
+## Implementation History
+
+N/A
+
+## Drawbacks
+
+- Complex implementation
```
gparvin, line 300: Requires management of an additional service account by the policy author.  This makes the policy development a little more complex, but trades the benefit of not copying resources to the policy authoring namespace.

enhancements/sig-policy/122-expanded-hub-templates-access/README.md
mprahl: Yes, that's right.

Issue #123, opened 2024-07-08
support multiple addon configs with same GVK


haoqing0110:
Ref https://github.com/open-cluster-management-io/addon-framework/issues/263

PR #124, opened 2024-07-08
Addon Multiple Configs with Same GVK

haoqing0110:
Ref: https://github.com/open-cluster-management-io/enhancements/issues/123

### DIFF
--- a/enhancements/sig-architecture/123-addon-multiple-configs/README.md
+++ b/enhancements/sig-architecture/123-addon-multiple-configs/README.md
@@ -1,0 +1,230 @@
+# Addon Multiple Configs with Same GVK
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal is to propose an enhancement to the addon configurations, enables users to configure multiple configs of the same GVK.
+
+## Motivation
+
+Users have three locations to set [addon configurations](https://open-cluster-management.io/concepts/addon/#add-on-configurations): 
+1. In `ClusterManagementAddon` `spec.supportedConfigs`, user can set the `defaultConfig` where all the addons to have same configurations.
+2. In `ClusterManagementAddOn`, `spec.installStrategy.placements[].configs`, user can set configs for addons on a group of clusters selected by placement.
+3. In `ManagedClusterAddOn` `spec.configs`, user can set configs for addon on a specific cluster.
+
+Currently, configs with the same GVK only support a single config. The config with the same GVK in place `3` will override the config in place `2`, 
+which will override the config in place `1`. The final effective addon configs are listed in the mca.status.
+
+Now we are seeing the requirement to set multiple configs for the same GVK. For example, for `OpenTelemetryCollector`, one CR might be configured to 
+collect traces and another to collect logs on the same cluster. This proposal aims to satisfy that requirement.
+
+### Goals
+
+- Enhance `ClusterManagementAddon` and `ManagementClusterAddon` API to allow user to configure mutiple addon configs with same GVK.
+
+### Non-Goals
+
+- `ClusterManagementAddon` `spec.supportedConfigs` does not allow user to configure mutiple addon configs with same GVK in the `defaultConfig`.
+
+## Proposal
+
+### User Stories
+
+1. The admin can configure the `ClusterManagementAddon` API to set mutilple configs with same gvk per install strategy.
+2. The admin can configure the `ManagementClusterAddon` API to set mutilple configs with same gvk for specific cluster.
+
+## Design Details
+
+### API changes
+
+`cma.spec.installStrategy.placements[].configs` and `mca.spec.configs` already support mutiple configs. 
+
+```go
+	// Configs is the configuration of managedClusterAddon during installation.
+	// User can override the configuration by updating the managedClusterAddon directly.
+	// +optional
+	Configs []AddOnConfig `json:"configs,omitempty"`
+```
+
+### Effective addon configs
+
+As the motivation mentioned, today there are 3 places to set addon configurations. 
+
+Currently, since each GVK supports only one config, the GVK is used as the identifier for configs. When configs with same GVK configured in multiple places, it follows the override sequence and 
+only one config is listed as the effective config in the mca.status.
+
+With the support for multiple configs with the same GVK, `GVK + namespace + name` becomes the identifier. 
+We need to reconsider the logic for setting mca.status when configuring multiple configs in multiple places.
+
+Using the example below (omitting the group and namespace for simplicity), we present three options. 
+We do not expect to support all, but want to involve discussion and finally only choose one.
+
+```yaml
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ClusterManagementAddOn
+metadata:
+  name: example
+spec:
+  supportedConfigs:
+    - resource: r1
+      defaultConfig:
+        name: n1
+    - resource: r2
+  installStrategy:
+    type: Placements
+    placements:
+      - name: example
+        namespace: example
+        configs:
+          - resource: r2
+            name: n1
+          - resource: r2
+            name: n2
+
+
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,230 @@
...
+    name: n1
+    resource: r1
+```
+
+#### Override by GVK
+
+"Override by GVK" is a way try to keep consistent with current behavior. 
+
+The gaps is if the r2n1 and r2n2 resource are actually some global configs, you may not want it to be overriden by r2n3. Compared with "Merge" options, 
+this options can not cover such case and is lack of flexibility.
```
JoaoBraveCoding, line 149: I see this point but at the same time, this approach is more similar to the current behaviour. However, "global configs" use case is a valid one that maybe could be addressed in a follow-up enhancement, I was thinking along the lines of CMA having a dedicated place for default configs like it does for `defaultConfigs` (just an idea).

JoaoBraveCoding (commented):
Thank you @haoqing0110 for putting this together it's a great start!

enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+#### Merge
+
+With `GVK + namespace + name` as the identifier for each config, there is no override concept now actually, all the configured configs will be listed in the mca.status. 
+
+Configs will be ordered, with the last configured on top. For example, configs in mca will be on top of cma configs, and mca configs with higher indices 
+will be on top of those with lower indices.
+
+The addon's own controller will determine how to use the configs (e.g., choose one, read all, etc).
```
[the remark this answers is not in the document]
haoqing0110: In the merge option, the configs in mca and cma will be listed in order. For example, if c1 is an `OpenTelemetryCollector` cr configured to collect traces and c2 is to collect logs, c3 is a new config to collect logs, the controller will see c3, c2, c1, it could always choose the first one to collect logs which is c3, and could also choose the other one for traces which is c1. Does this make sense from addon's view?

enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,230 @@
...
+### User Stories
+
+1. The admin can configure the `ClusterManagementAddon` API to set mutilple configs with same gvk per install strategy.
+2. The admin can configure the `ManagementClusterAddon` API to set mutilple configs with same gvk for specific cluster.
+
+## Design Details
+
+### API changes
+
+`cma.spec.installStrategy.placements[].configs` and `mca.spec.configs` already support mutiple configs.
```
iblancasa, line 47: Multiple configurations are supported but since the GVK is used as the identifier, only one is persisted, right?

enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,230 @@
...
+  name: example
+  namespace: cluster1
+spec:
+  installNamespace: open-cluster-management-agent-addon
+  configs:
+    - resource: r2
+      name: n3
+```
+
+#### Merge
```
iblancasa, line 104: For this strategy: what will happen when we have one resource in the `ManagedClusterAddon` and another one with the same name, GVK and namespace in `ClusterManagemenetAddon`? I would expect an automatic override but I'm not sure if you want to do something to keep both of them in order or something.

enhancements/sig-architecture/123-addon-multiple-configs/README.md
haoqing0110: yes, today user can define multiple configs with the same gvk at the API level, but only one will be listed in the mca.status.

enhancements/sig-architecture/123-addon-multiple-configs/README.md
haoqing0110: if the one in `ClusterManagemenetAddon` has the same gvk, name and namespace as the one in the `ManagedClusterAddon`, they are actually the same config. In this case, since merge uses gvk + name + namespace as the key, so only one will be listed in the `mca.status`.

enhancements/sig-architecture/123-addon-multiple-configs/README.md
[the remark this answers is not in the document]
JoaoBraveCoding: I understand the configs will be shown in order, but this is insufficient to inform the addon if a resource comes from the global config in CMA or a cluster-specific config in MCA. For example, assume a OTELCollector config that would collect logs (c1), but then for a given cluster we are troubleshooting something and we want to send only metrics & traces so we specify (c2 & c3, two separate `OpenTelemetryCollector`). From the addon perspective, it's not clear what config should be used even if they are sorted by (c3, c2, c1). The only alternative is to go through the resources in the Spec of MCA and calculate the intersection with the Status.

After discussing with the team from our POV we prefer option 2. However, if the OCM team favours option 1 then we also do not oppose since we have the workaround to calculate the intersection.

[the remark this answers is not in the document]
haoqing0110: OK, this case makes sense. For this proposal, the option 2 can be selected to be implemented first, since it makes more sense for real use cases. As for the "global configs" issues, to avoid override, the workaround would be to configure the "global configs" at both cma and cma. 

Will update the doc and leave the other 2 options for future needs.

enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,245 @@
...
+As the motivation mentioned, today there are 3 places to set addon configurations. 
+
+Currently, since each GVK supports only one config, the GVK is used as the identifier for configs. When configs with the same GVK configured in 
+multiple places, it follows the override sequence and only one config is listed as the effective config in the `mca.status`. 
+
+With the support for multiple configs with the same GVK, `GVK + namespace + name` becomes the identifier. 
+We need to reconsider the logic for setting `mca.status` when configuring multiple configs in multiple places.
+
+In the initial discussion we listed three options, finally choose the below "Override by GVK" option, as it satisfies more on 
+the real use case than the other two. The other two are listed in the **Alternatives** for future needs.
```
JoaoBraveCoding, line 70:
```suggestion
In the initial discussion we listed three options, finally choose the below "Override by GVK" option, as it better satisfies a real use case. 
The other two are listed in the **Alternatives** for future needs.
```

enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,245 @@
...
+      name: n3
+```
+
+#### Override by GVK
+
+"Override by GVK" is a way try to keep consistent with current behavior. GVK is used as the identifier for a group of configs.
+
+In the above example, config `r2` `n3` in `ManagedClusterAddOn` (place `3`) will override `r2` `n1` and `n2` in `installStrategy` (place `2`). 
+And for default config `r1` `n1`, since no other `r1` resource defined in place `2` and `3`, it will be kept in the effective addon configs.
+The `mca.status` will be like:
```
JoaoBraveCoding, line 115:
```suggestion
In the above example, config `r2` `n3` in `ManagedClusterAddOn` (place `3`) will override `r2` `n1` and `n2` in `installStrategy` (place `2`). The default config `r1` `n1` will remain, since no other config for resource `r1` is defined in either place `2` or `3`.
The `mca.status` will look like the following:
```

enhancements/sig-architecture/123-addon-multiple-configs/README.md
```diff
@@ -0,0 +1,245 @@
...
+    name: n1
+    resource: r1
+```
+
+The limitation is if the `r2n1` and `r2n2` resources are some global configs, the user may not want it to be overridden by `r2n3`. In this case, the user need to define
+both `r2n1`,`r2n2` and `r2n3` in `ManagedClusterAddOn`.
+
+### Effective config values
+
+`WithGetValuesFuncs()` depends on how the funcs handle multiple configs, for example:
```
JoaoBraveCoding, line 143:
```suggestion
`WithGetValuesFuncs()` depends on how some functions handle multiple configs, for example: 
```

JoaoBraveCoding (commented):
Just some suggestions to make the proposal easier to read. But in general LGTM 👍 great job @haoqing0110

haoqing0110:
@qiujian16 plz help do a final review.

qiujian16 (commented):
/approve
/lgtm

PR #125, opened 2024-07-10
Adding enhancement proposal to support AWS IAM for OCM

suvaanshkumar:
In this Proposal, 
We are proposing a way to support IAM based authentication so that Kubernetes environments such as Amazon Elastic Kubernetes Service (EKS) can also be supported by OCM. 
Signed-off-by: Gaurav Jaswal <jaswalkiranavtar@gmail.com>


mikeshng:
CC @jnpacker @berenss @showeimer

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+The role name and policies created on managed cluster IAM will follow a templates to ensure standardization. The role name will contain md5 hash of the values of tags to uniquely identify.
+
+```shell
+arn:aws:iam::<managed-cluster-account-id>:role/ocm-spoke-<md5-hash-unique-identifier>
+```
+
+Another reason to follow a standard name is to ensure that the role can be created before running `clusteradm join` command and the klusterlet/registration agent can assume that role, on startup, with a standardized name.
+
+#### Graceful resource cleanup on both hub and managed cluster when managed cluster triggers unjoin
```
zhujian7, line 159: Currently the `clusteradm unjoin` command only deletes resources on the managed cluster side, the ManagedCluster CR on the hub side will not be deleted.
I am curious if the IAM roles and policies only need to be deleted when deleting the ManagedCluster, if it is, maybe a finalizer is simpler

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+          "<oidc-provider>:sub": "system:serviceaccount:open-cluster-management-hub:registration-controller-sa",
+          "<oidc-provider>:aud": "sts.amazonaws.com"
+        }
+      }
+    }
+  ]
+}
+```
+
+#### Cluster join initiated from managed cluster
```
zhujian7, line 327: We also have some addons that need to access the Hub cluster apiserver, it would be cool if the addon registration workflow was shown here as well.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: We want managedcluster specific resources on the hub to be cleaned up even when managed cluster triggers unjoin and managedcluster doesn't have delete permission on its managedcluster CR on hub.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: @zhujian7 we do not have very good understanding of the addon framework at this point and couldn't think about the finer details of addon registration process. Having said that, my understanding is that if we use the registration driver approach proposed by @qiujian16 the addon registration should be similar to core registration. 

I would request your help on this, if you agree with the core registration approach proposed here. Could you please help extend the same to addon registration?

qiujian16: yes, I think we could add addon registration as non-goal in this proposal, and try to do it in another one. This will limit this for only cluster registration, but should be fine as an alpha feature.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+1. cluster-admin on managed cluster gets a bootstrap kubeconfig to connect to hub,
+   and deploy the agent on managed cluster.
+- it has the identity to create `ManagedCluster`.
+
+2. agent on managed cluster creates `ManagedCluster` on hub if it does not exist.
+- The name of `ManagedCluster` is passed using the `--cluster-name` option while running `clusteradm join` command.
+- The agent on the managed cluster will store it as `agent-name` in `Secret` `hub-kubeconfig-secret`, so restarting agent or redeploying agent will not lose the UID after the cluster is managed successfully.
```
qiujian16, line 334: cluster name is stored as cluster-name, agent-name is an random uid which is unique for each agent.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+    aws eks update-kubeconfig --name <hub-eks-cluster-name> --kubeconfig <path/to/create/kubeconfig> --role-arn <role-on-hub-for-managed>
+    ```
+
+10. agent on managed cluster connects to hub apiserver using the new kubeconfig.
+
+11. agent on managed cluster updates conditions of `ManagedCluster` as `ManagedClusterJoined`.
+
+12. agent on managed cluster appends updates other fields in status of `ManagedCluster`. 
+
+13. agent on managed cluster pulls the`ManifestWork` and creates resource.
```
qiujian16, line 413: the general flow look good to me

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+Registration workflow
+![OCM-using-AWS-IAM.jpg](OCM-using-AWS-IAM.jpg)
+
```
qiujian16, line 417: maybe we should also write an unjoin flow. Or if accept is updated false, will controller remove iam roles also?

qiujian16 (commented):
Thanks! it provides a much clearer picture on what we want to do.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+The EKS kubeconfig will use these temporary IAM creds to generate temporary sts token using `aws eks get-token` command present in kubeconfig with [TTL of 15 mins](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#:~:text=The%20token%20has%20a%20time,each%20time%20the%20token%20expires.). This sts token authenticates every request sent to EKS api-server, and is automatically renewed by kubeconfig on expiry. Snippet of user from an EKS kubeconfig:
+```yaml
+users:
+- name: test-user
+  user:
+    exec:
```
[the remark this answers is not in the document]
jaswalkiranavtar: We want to keep the design for this solution similar to existing csr based auth that uses a hub kubeconfig from a secret called hub-kubeconfig-secret.

Also we are planning to publish and use separate images for registration and work agent on spoke and registration controller on hub if selected `--registration-auth` is `aws-irsa`, that will contain aws cli. This is done to ensure that csr based solution is not unnecessarily bloated with aws cli or sdk. Whether to use aws or non-aws images will depend upon the value of `--registration-auth` and we will have to update clusteradm for this.

Also if my understanding is correct, we will have to make a lot more changes from the exiting csr solution if we decide to use aws sdk approach instead of loading hub kubeconfig file.

Could you please explain what do you mean by tokenfile, if you are talking about the one mentioned [here](https://docs.aws.amazon.com/eks/latest/userguide/pod-configuration.html), that is OIDC token and cannot be used for authentication with EKS, EKS needs sts token which is generated by `aws eks get-token` command.

Please let me know if i am missing anything here and there is a way to avoid using aws-cli without diverging too much from the existing design.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+One managed cluster should not be able to access resources on hub that belong to another managed cluster. 
+
+### Implementation Details/Notes/Constraints
+
+#### Changes required on hub
+The IAM roles and policies for managed cluster will be created when hub-admin accepts the registration request by running `clusteradm accept` command on hub. The role name and policies will be created using templates to ensure standardization. The role name should follow a standard naming convention so that the role on managed cluster can assume the hub role with a standardized name. It will also create an entry aws-auth configmap which will bind the iam role with permissions inside the cluster so that the incoming requests from managed cluster can have the required access inside the hub cluster.
+
+If the feature gate `ManagedClusterAutoApproval` feature gate is enabled, choosing `aws-irsa` instead of `csr` will disable [csr controller](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/csr/controller.go) on hub and enable a new `aws-irsa` controller, which will auto approve all registration requests, and create IAM roles and policies and RBAC permissions.
```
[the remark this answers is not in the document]
jaswalkiranavtar: We thought we will disable it to avoid running a goroutine which will never be used.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+```shell
+arn:aws:iam::<hub-account-id>:role/<hub-cluster-name>_managed-cluster-identity-creator
+```
+
+#### Changes required on the managed cluster
+Choosing `aws-irsa` instead of `csr` will disable [cert_controller](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/clientcert/cert_controller.go) on managed cluster and enable a new `aws-irsa` controller in klusterlet/registration agent, which will create only `ManagedCluster` on hub and not CSR. The `aws-irsa` controller will watch for the updates on `ManagedCluster` , and once accepted on hub and has HubAcceptedManagedCluster condition, it will generate a kubeconfig using following command and save it to a secret called hub-kubeconfig-secret on managed cluster:
```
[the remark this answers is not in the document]
jaswalkiranavtar: Yes you are correct. If we create hub-kubeconfig without authorization on the hub, the work agent will consume the kubeconfig and [start](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/singleton/spoke/agent.go#L48) making requests to hub which will obviously fail because  authorization might not have been granted by that time, hence we decided to keep it waiting.

But please share your thoughts.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+```shell
+arn:aws:iam::<managed-cluster-account-id>:role/ocm-spoke-<md5-hash-unique-identifier>
+```
+
+Another reason to follow a standard name is to ensure that the role can be created before running `clusteradm join` command and the klusterlet/registration agent can assume that role, on startup, with a standardized name.
+
+#### Graceful resource cleanup on both hub and managed cluster when managed cluster triggers unjoin
+
+The `ManagedCluster` will have a new boolean field `clientUnjoinsHub` with the default value of false. The `clusteradm unjoin` command by managed cluster admin will update the `ManagedCluster` on hub and set the `spec.clientUnjoinsHub` to true, which will trigger the new `aws-irsa` controller on hub to clean-up all the managed cluster specific resources, like IAM roles and policies on hub, and ultimately delete the `ManagedCluster`.
```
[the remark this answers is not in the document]
jaswalkiranavtar: Looking [here](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/managedcluster/controller.go#L132), looks like it will delete managed cluster resources like roles and policies and we could have refactored it to include aws iam resources as well. However we want to clean up ManagedCluster CR as well from hub.

Is there a way to do that using any existing feature today? If yes, we are ok to update the proposal.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+         "Action": "sts:AssumeRole",
+         "Resource": "arn:aws:iam::<hub-account-id>:role/ocm-hub-<md5-hash-unique-identifier>"
+      }
+   ]
+}
+```
+
+And following trust relationship policy:
+```json
+{
```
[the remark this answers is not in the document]
jaswalkiranavtar: If my udnerstanding is correct, <klusterlet-name>-registration-sa is the service account on [registration-agent](https://github.com/open-cluster-management-io/ocm/blob/main/manifests/klusterlet/management/klusterlet-registration-deployment.yaml), when deployed in [non-singleton](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_runtime_reconcile.go#L75) mode.

The registration-agent will create hub-kubeconfig-secret in managed cluster when the authoization is granted on hub. The hub-kubeconfig generation needs the iam permissions on hub as mentioned [here](https://github.com/suvaanshkumar/enhancements/blob/OCM-using-AWS-IAM/enhancements/sig-architecture/105-aws-iam-registration/README.md?plain=1#L352), that is why we are trying to add this role.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: We will have to enhance hub controller to remove iam roles, as today it [removes](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/managedcluster/controller.go#L212) k8s rbac resources only. However setting accept to false wont remove ManagedCluster CR, can "setting accept to false" be updated to remove ManagedCluster CR instead of just letting it stay there with [reason denied](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/managedcluster/controller.go#L139)?

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: Thanks for spotting this, will update the proposal.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
qiujian16: What i mean is sts token, and I am thinking you can use AWS go sdk to get the token instead of using the cmd line

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
qiujian16: why do you want to clean up ManagedCluster CR with a field in the managedcluster CR? I think user will just run kubectl delete to remove managedcluster CR.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
qiujian16: my understanding is hub-kubeconfig does not relates to registration-sa. registration-sa is for agent to talk to spoke cluster. It should not be related to the secret that agent talks to the hub.

enhancements/sig-architecture/105-aws-iam-registration/README.md
qiujian16: I think we should remove iam role upon setting accept to false OR manageCluster CR is deleted.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+    exec:
+      apiVersion: client.authentication.k8s.io/v1beta1
+      command: aws
+      args:
+      - --region
+      - my-aws-region
+      - eks
+      - get-token
+      - --cluster-name
+      - my-cluster
```
deads2k, line 118: is this the cluster-name of the hub or the cluster-name of the spoke?

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+% clusteradm init \
+     --wait \
+     --registration-auth=aws-irsa \
+     --context ${CTX_HUB_CLUSTER}
+```
+
+
+#### Story 2 - EKS Hub administrator must initialize a hub using aws-irsa authentication strategy only, using csr authentication strategy will throw an error
+
+`clusteradm init` command must fail if the EKS hub admin tries to initialize the hub, using `csr` authentication or, by leaving it at its default setting.
```
deads2k, line 56: Can the hub be non-EKS but wired to use the AWS IAM OIDC server?

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+```
+% clusteradm init \
+     --wait \
+     --registration-auth=aws-irsa \
+     --context ${CTX_HUB_CLUSTER}
+```
+
+
+#### Story 2 - EKS Hub administrator must initialize a hub using aws-irsa authentication strategy only, using csr authentication strategy will throw an error
```
deads2k, line 54: I agree it's reasonable to specify a list of allowed authentication strategies, but could it be a positive list that allows

1. aws-irsa only
2. csr only
3. aws-irsa and/or csr

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+      - get-token
+      - --cluster-name
+      - my-cluster
+```
+
+We don't have to do anything additional to refresh the tokens, aws/eks infrastructure takes care of that for us.
+
+#### Story 8 - Procedure for hub administrator to follow, to revoke access to a managed cluster
+
+It must be possible for the hub cluster administrator to revoke access to a managed cluster. The hub admin can simply delete the `ManagedCluster` from hub and the new `aws-irsa` controller will delete all managed cluster specific resources on hub.
```
deads2k, line 125: >`aws-irsa` controller

Is this is a controller on the hub or the spoke?  I assume hub, but please be explicit.

enhancements/sig-architecture/105-aws-iam-registration/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+This is to describe the various process of cluster join
+
+Actors:
+1. cluster-admin on managed cluster
+2. AWS IAM admin on managed cluster AWS account
+3. cluster-admin on hub cluster
+4. AWS IAM admin on hub cluster AWS account
+5. hub controller
+6. agent on managed cluster
```
deads2k, line 184: the opt-in on both sides appears to work and either can revoke the connection.  Additionally, with different namespaces for the klusterlet, it appears that the same managedcluster could join multiple hubs (important for migration over time).

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
deads2k: Perhaps this could be modeled as a side-car that persists the token to an emptyDir and avoid the dep?

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: it should be hub, will also update the proposal

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: I am not sure if that would be possible because in addition to AWS OIDC server, we also need IRSA capability of EKS for this solution to work.

Also thinking when would anybody use this kind of setup with non EKS hub and wired with AWS OIDC server.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: Agreed with #1 and 2, but not sure about the third option because our solution is dependent on IRSA which is an EKS feature and both hub and spoke will need to assume AWS IAM role through IRSA. csr solution wont work on EKS and aws-irsa solution won't work outside EKS, so i believe they are exclusive solutions and mix and match might not work.

However, please share your thoughts and let us know if you can think of any scenario where both solutions can work together.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: Just realized that we want to enhance the existing registration-controller on the hub to cleanup all AWS IAM resources. The new `aws-irsa` controller on the hub is meant for auto approval and is similar to [csr controller](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/csr/controller.go).

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: ManagedCluster doesn't have permission to delete ManagedCluster CR, so the only option we have is to update hubAcceptsClient field to false. Today if we set hubAcceptsClient field to false, it deletes all resources on hub but not the managedcluster CR itself. I hope it is ok if we enhance ManagedCluster controller on hub to delete ManagedCluster CR at the end, as we want to cleanup everything managedcluster specific resource on hub, when spoke triggers unjoin.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
jaswalkiranavtar: My understanding is same as well. In EKS IRSA roles are assigned through service accounts, we want to use this service account to assign spoke a role on its IAM, which it will use to assume role on hub.

Spoke agent won't assume a role on hub directly, it first assumes a role on spoke's IAM, and using that role assumes role on hub. Please refer "Example Use chained AssumeRole operations" section on [this](https://docs.aws.amazon.com/eks/latest/userguide/cross-account-access.html) article.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
jaswalkiranavtar: Because we want to ensure that if the unjoin process is triggered by spoke, all (IAM and k8s) spoke specific resources should be cleaned on hub.

Now spoke doesn't have delete permission on hub, it only has [update](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/manifests/rbac/managedcluster-clusterrole.yaml#L19) permission. So only think spoke can do is to set the hubAcceptsClient to false to trigger it and let managedcluster controller on hub to take care of the rest.

If we don't do this, the hub admin will have to manually cleanup all the managedcluster CRs and that is the step we are trying to avoid.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
jaswalkiranavtar: @qiujian16 are you able to share an article that shows how that can be done using aws go sdk and how the result token can be used with client-go library?

I like @deads2k init container approach as well as that seems to be less disruptive to existing design.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
qiujian16: I would suggest the deletion of managedcluster CR is done manually rather than a controller. hubAcceptClient can be turned from false to true also meaning the hub accepts agent again.

enhancements/sig-architecture/105-aws-iam-registration/README.md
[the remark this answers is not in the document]
qiujian16: I think it could work with init container.

enhancements/sig-architecture/105-aws-iam-registration/README.md
deads2k: > Also thinking when would anybody use this kind of setup with non EKS hub and wired with AWS OIDC server.

A hub cluster that allows joining from any cloud provider would be beneficial, rather than foreclosing the possibility early.

enhancements/sig-architecture/105-aws-iam-registration/README.md
deads2k: Can you help describe the IRSA capability that is missing in non-EKS capabilities?  Is the interpretation of AWS tokens somehow not possible?  Is the user mapping not done?  I would expect the ability to interpret AWS tokens to be generally possible so different clusters could inter-operate.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: Yes, we will remove the IAM resources on hub when spoke admin sets spec.hubAcceptsClient to false on ManagedCluster CR on hub.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: Sure for now, we will keep the hub to support all types of auth for now. We will add support of EKS hub using aws iam only for now.

However, the proposal can be enhanced later to use aws iam auth on non EKS hub as well using TokenReview API. The non EKS hub can get the sts token validated by sending TokenReview request to another EKS cluster.

enhancements/sig-architecture/105-aws-iam-registration/README.md
jaswalkiranavtar: Updated the proposal to list all the options under story 2

qiujian16: maybe some rewording here. It is not necessary to aws only and throw error for csr in this case. The user certainly could select to only enable aws-isra

qiujian16 (commented):
In general looks good. I think it just needs some rewording on story 2

enhancements/sig-architecture/105-aws-iam-registration/README.md
suvaanshkumar: Thanks @qiujian16 ,
I just updated the PR. Please let me know if there is anything missing.

qiujian16:
/approve

Thanks, please ensure all the commits are signoff-ed

mikeshng (approved):
/lgtm

PR #129, opened 2024-10-24
Resource selector for `ConfigurationPolicy`

### LINKED ISSUE #128: Resource selector for ConfigurationPolicy
Currently a `namespaceSelector` exists to select relevant namespaces. It would be paired well with a resource selector to also select resources, potentially lowering the bar to entry to allow users to not resort to a `range` using `object-templates-raw`.

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-14828

Closes #128

### DIFF
--- a/enhancements/sig-policy/128-resource-selector/README.md
+++ b/enhancements/sig-policy/128-resource-selector/README.md
@@ -1,0 +1,148 @@
+# Resource selector for `ConfigurationPolicy`
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+Add a resource selector to pair with the existing `namespaceSelector` in order to allow a
+`ConfigurationPolicy` to filter and select objects when an `objectDefinition` doesn't have a name
+provided. The resource selector would also enable the enforcement of `objectDefinitions` without a
+name.
+
+## Motivation
+
+The `ConfigurationPolicy` currently has a `namespaceSelector` to configure the controller to filter
+and range over namespaces relevant to a particular namespaced `objectDefinition`. Furthermore,
+currently an `objectDefinition` without a name is feasible but is not enforceable. A
+`resourceSelector` would both pair well with the `namespaceSelector` for further object filtering as
+well as enable an enforceable unnamed `objectDefinition` by providing the controller with a list of
+names to handle.
+
+Adding a `resourceSelector` would lower the bar to entry for `ConfigurationPolicy`, adding an
+additional configuration available to users without requiring knowledge of Go templating as is
+required with the alternative, `object-templates-raw`.
+
+### Goals
+
+- Add a `resourceSelector` to `ConfigurationPolicy` to allow the controller to handle objects
+  without requiring `object-templates-raw`
+- Consolidate compliance messages to list out objects per namespace (one downside to the alternative
+  `object-templates-raw` is there's a compliance message for each object)
+- Make unnamed `objectDefinition`s enforceable
+
+### Non-Goals
+
+- Establishing complex logic in the `resourceSelector` that could not be handled readily by the
+  Kubernetes API and, by extension, the `kubernetes-dependency-watches` library.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a policy user, I want to be able to create namespaces based on the `namespaceSelector`.
+
+#### Story 2
+
+As a policy user, I want to be able to select objects by label and have the controller iterate over
+them, both to `inform` and `enforce`.
+
+### Implementation Details/Notes/Constraints
+
+Since the `resourceSelector` refers to a specific object, the `resourceSelector` would be embedded
+in each element of the `object-templates` array. The `resourceSelector` would ONLY be composed of a
+`LabelSelector` since that can be handled by the Kubernetes API. The `namespaceSelector` has
+`include` and `exclude` filters for filtering by name, but such a feature requires that it run in
+its own `Namespace` reconciler, which would be unwieldy with a `resourceSelector` where the `Kind`
+is arbitrary.
+
+In a YAML manifest, this new field would appear as follows:
+
+```yaml
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: resource-selector-policy
+spec:
+  remediationAction: inform
+  namespaceSelector:        # <---- Existing namespaceSelector
+    exclude: []             #
+    include: []             #
+    matchLabels: {}         #
+    matchExpressions: []    #
+  object-templates:
+    - complianceType: musthave
+      resourceSelector:         # <---- New resourceSelector
+        matchLabels: {}         #
+        matchExpressions: []    #
+      objectDefinition:
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          labels:
+            selected: true
+    - ...
+```
+
+The `config-policy-controller` would list the provided `Kind` across all namespaces using the
+provided `resourceSelector` and iterate over each matching object in each namespace to determine
+compliance. Namespaces would be determined from either the namespace defined in the
+`objectDefinition` or the `namespaceSelector` results.
+
+The label selectors in `resourceSelector` will be a pointer to determine whether the user has
+provided it to distinguish between fetching all objects vs fetching no objects.
+
+### Risks and Mitigation
+
+- Currently `mustnothave` doesn't consider fields when a name is provided, but _does_ consider
+  fields when a name is not provided.
+  - Caution should be taken (and tests written) to refactor `mustnothave` to consider fields as part
+    of the `resourceSelector`.
+- The implementation could increase load on the Kubernetes API.
+  - Depending on how it's implemented, I wouldn't think the load would be any greater than if
+    `object-templates-raw` were in play. Ideally the initial `List` request would populate the cache
+    and any subsequent `Get` (if any) would fetch from the cache.
+- Adding a `resourceSelector` could enable enforcement where enforcement wasn't previously
+  available, which could have unintended consequences for users.
+  - One possibility here would be to not enable enforcement, but I think there would be more demand
+    for enforcement than not. And a change would need to be made to the policy in order to enable
+    it, which hopefully users would test beforehand.
+
+### Open Questions
+
+1. Should `remediationAction: enforce` be allowed?
+   - Some users may rely on the current behavior that an unnamed `objectDefinition` can only be
+     `inform` and may cause unexpected results when a `resourceSelector` is newly applied to a
+     policy they didn't realize was enforced.
+2. Should the `resourceSelector` be able to determine the namespace if the namespace is not
+   provided?
+   - Currently if the namespace is not provided in the `objectDefinition` and the
+     `namespaceSelector` returns no results, no objects are returned. I think this behavior should
+     be preserved since it would follow the behaviors of tools like `kubectl`, but I could be
+     convinced otherwise.
+3. Should the `resourceSelector` have include/exclude name filtering?
+   - This question was raised previously in standup.
+
+### Test Plan
+
+All code will have tests meeting coverage targets. All code will have e2e tests.
+
+### Graduation Criteria
+
+--
+
+## Implementation History
+
+--
+
+## Alternatives
+
+Users can accomplish this currently using `objects-templates-raw`.
--- a/enhancements/sig-policy/128-resource-selector/metadata.yaml
+++ b/enhancements/sig-policy/128-resource-selector/metadata.yaml
@@ -1,0 +1,10 @@
+title: Resource selector for ConfigurationPolicy
+authors:
+  - "@dhaiducek"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2024-10-23
+last-updated: 2024-10-23
+status: implementable


enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+well as enable an enforceable unnamed `objectDefinition` by providing the controller with a list of
+names to handle.
+
+Adding a `resourceSelector` would lower the bar to entry for `ConfigurationPolicy`, adding an
+additional configuration available to users without requiring knowledge of Go templating as is
+required with the alternative, `object-templates-raw`.
+
+### Goals
+
+- Add a `resourceSelector` to `ConfigurationPolicy` to allow the controller to handle objects
```
mprahl, line 34: Do you think another goal is to consolidate the compliance message for an object-template with a resource selector? One of the main drawbacks with object-templates-raw is you get a separate compliance message per object-template. The use of a namespaceSelector currently consolidates the compliance message but that is limited as you mention in the introduction.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+        metadata:
+          labels:
+            selected: true
+    - ...
+```
+
+The `config-policy-controller` would list the provided `Kind` across all namespaces using the
+provided `resourceSelector` and iterate over each maching object in each namespace to determine
+compliance. Namespaces would be determined from either the namespace defined in the
+`objectDefinition` or the
```
mprahl, line 99:
```suggestion
`objectDefinition` or the `namespaceSelector` results.
```

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+    and any subsequent `Get` (if any) would fetch from the cache.
+- Adding a `resourceSelector` could enable enforcement where enforcement wasn't previously
+  available, which could have unintended consequences for users.
+  - One possibility here would be to not enable enforcement, but I think there would be more demand
+    for enforcement than not. And a change would need to be made to the policy in order to enable
+    it, which hopefully users would test beforehand.
+
+### Open Questions
+
+1. Should `remediationAction: enforce` be allowed?
```
mprahl, line 115: I agree with you making `enforce` a goal.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+    exclude: []             #
+    include: []             #
+    matchLabels: {}         #
+    matchExpressions: []    #
+  object-templates:
+    - complianceType: musthave
+      resourceSelector:         # <---- New resourceSelector
+        matchLabels: {}         #
+        matchExpressions: []    #
+      objectDefinition:
```
mprahl, line 87: Could you touch on how you see templating playing into this? It'd be cool if for every namespace and resource selector result, you could execute the templates with the input namespace and resource name as template variables. You get this with the object-templates-raw alternative, and this could be a simpler way to get the same effect.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+    for enforcement than not. And a change would need to be made to the policy in order to enable
+    it, which hopefully users would test beforehand.
+
+### Open Questions
+
+1. Should `remediationAction: enforce` be allowed?
+   - Some users may rely on the current behavior that an unnamed `objectDefinition` can only be
+     `inform` and may cause unexpected results when a `resourceSelector` is newly applied to a
+     policy they didn't realize was enforced.
+2. Should the `resourceSelector` be able to determine the namespace if the namespace is not
```
mprahl, line 119: Could you please rephrase this question or provide more context? I'm not quite understanding the question.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+   provided?
+   - Currently if the namespace is not provided in the `objectDefinition` and the
+     `namespaceSelector` returns no results, no objects are returned. I think this behavior should
+     be preserved since it would follow the behaviors of tools like `kubectl`, but I could be
+     convinced otherwise.
+3. Should the `resourceSelector` have include/exclude name filtering?
```
mprahl, line 125: I think the way you have it designed with the namespaceSelector feeding into the resourceSelector is good. No need for further filtering.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+The `ConfigurationPolicy` currently has a `namespaceSelector` to configure the controller to filter
+and range over namespaces relevant to a particular namespaced `objectDefinition`. Furthermore,
+currently an `objectDefinition` without a name is feasible but is not enforceable. A
+`resourceSelector` would both pair well with the `namespaceSelector` for further object filtering as
+well as enable an enforceable unnamed `objectDefinition` by providing the controller with a list of
+names to handle.
+
+Adding a `resourceSelector` would lower the bar to entry for `ConfigurationPolicy`, adding an
+additional configuration available to users without requiring knowledge of Go templating as is
+required with the alternative, `object-templates-raw`.
```
yiraeChristineKim, line 30: Love this. No need to learn Go templating

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Open Questions
+
+1. Should `remediationAction: enforce` be allowed?
+   - Some users may rely on the current behavior that an unnamed `objectDefinition` can only be
+     `inform` and may cause unexpected results when a `resourceSelector` is newly applied to a
+     policy they didn't realize was enforced.
+2. Should the `resourceSelector` be able to determine the namespace if the namespace is not
+   provided?
+   - Currently if the namespace is not provided in the `objectDefinition` and the
```
yiraeChristineKim, line 121: What if users wants to apply  All namespace, and name is "*-pod"? Should we need a field for this?

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+`ConfigurationPolicy` to filter and select objects when an `objectDefinition` doesn't have a name
+provided. The resource selector would also enable the enforcement of `objectDefinitions` without a
+name.
+
+## Motivation
+
+The `ConfigurationPolicy` currently has a `namespaceSelector` to configure the controller to filter
+and range over namespaces relevant to a particular namespaced `objectDefinition`. Furthermore,
+currently an `objectDefinition` without a name is feasible but is not enforceable. A
+`resourceSelector` would both pair well with the `namespaceSelector` for further object filtering as
```
yiraeChristineKim, line 24: Does resourceSeletor select a resource by only name? or Are there any other opts?

dhaiducek: The `resourceSelector` fetches objects by label and returns the resulting names, which is feasible with the Kubernetes API. For anything else, users can use `object-templates-raw` for additional filtering.

enhancements/sig-policy/128-resource-selector/README.md
dhaiducek: That's a good point--the message could probably be refactored to list out the objects per namespace rather than have a message for every object--is that what you mean?

mprahl: Either that or consolidate to something like `Pods found as specified: namespace1/name1, namespace2/name2; Pods updated: namespace1/name3 ` per object-template with a resource selector.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+### User Stories
+
+#### Story 1
+
+As a policy user, I want to be able to create namespaces based on the `namespaceSelector`.
+
+#### Story 2
+
+As a policy user, I want to be able to select objects by label and have the controller iterate over
```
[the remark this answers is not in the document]
dhaiducek: Any Kubernetes object can have a label listed in `metadata.labels`. Currently if an `objectDefinition` provides a definition without `metadata.name`, the controller will handle it but will refuse to enforce it. With the `resourceSelector`, it's being fed a list of names, so we can enable enforcement since it's a user-defined filtered list.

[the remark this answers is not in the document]
yiraeChristineKim: "enforce" We usually create new resources. but how `resourceSelector` apply to this?

enhancements/sig-policy/128-resource-selector/README.md
dhaiducek: Haha, oops! Good catch!

enhancements/sig-policy/128-resource-selector/README.md
dhaiducek: Technically when fetching by label across all namespaces, the `resourceSelector` would successfully retrieve a list of objects. The question here is whether the `resourceSelector` should also be able to fill in the namespace if the namespace is not provided in the `objectDefinition` or if we would also require a `namespaceSelector` to be provided as we do today. I lean toward the latter and only allow the `resourceSelector` to fill in the name and not the namespace.

enhancements/sig-policy/128-resource-selector/README.md
dhaiducek: That's one of the cases (i.e. include/exclude name filtering) that can't be handled by the Kubernetes API. During standup the preference was to omit such name filtering since it would require additional controller overhead since it can't be handlded by the Kubernetes server.

enhancements/sig-policy/128-resource-selector/README.md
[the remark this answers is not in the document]
dhaiducek: This wouldn't apply to creating new resources. This would only apply to objects that already exist.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,144 @@
...
+Adding a `resourceSelector` would lower the bar to entry for `ConfigurationPolicy`, adding an
+additional configuration available to users without requiring knowledge of Go templating as is
+required with the alternative, `object-templates-raw`.
+
+### Goals
+
+- Add a `resourceSelector` to `ConfigurationPolicy` to allow the controller to handle objects
+  without requiring `object-templates-raw`
+- Consolidate compliance messages to list out objects per namespace (one downside to the alternative
+  `object-templates-raw` is there's a compliance message for each object)
```
JustinKuli, line 37: This brings up the question: is the content in the compliance message part of our API which we should be very careful about changing? I suppose we *could* only change the compliance messages to a new form if the policy uses a `resourceSelector`, but that seems potentially confusing to users.

Personally, I don't think we need to preserve the existing messages exactly, as long as similar content is still present by default. We could consider exposing a "legacy" message into the CustomMessage template for use, in case a user has some other tooling built on the current format.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,144 @@
...
+- Make unnamed `objectDefinition`s enforceable
+- Make the `namespaceSelector` apply to `Namespace`s when a `resourceSelector` is not provided for a
+  `Namespace` `objectDefinition`, allowing a related `Namespace` to be created alongside related
+  namespaced objects.
+
+### Non-Goals
+
+- Establishing complex logic in the `resourceSelector` that could not be handled readily by the
+  Kubernetes API and, by extension, the `kubernetes-dependency-watches` library.
```
JustinKuli, line 46: This is about the "include" and "exclude" glob filters we have, is that right? I just want to check that there isn't some other complex logic we've considered.

enhancements/sig-policy/128-resource-selector/README.md
JustinKuli: I also agree. I think documenting that adding a resource selector can allow a policy that was previously unenforceable to be enforced would be sufficient warning to users.

enhancements/sig-policy/128-resource-selector/README.md
dhaiducek: Yes, the include/exclude glob filters.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,151 @@
...
+required with the alternative, `object-templates-raw`.
+
+### Goals
+
+- Add a `resourceSelector` to `ConfigurationPolicy` to allow the controller to handle objects
+  without requiring `object-templates-raw`
+- Consolidate compliance messages to list out objects per namespace (one downside to the alternative
+  `object-templates-raw` is there's a compliance message for each object)
+- Make unnamed `objectDefinition`s enforceable
+- Make the `namespaceSelector` apply to `Namespace`s when a `resourceSelector` is not provided for a
```
mprahl, line 39: I think from our discussion, we decided this wouldn't be the case and that the user needs to use resource selector for this.

enhancements/sig-policy/128-resource-selector/README.md
```diff
@@ -0,0 +1,151 @@
...
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          labels:
+            selected: true
+    - ...
+```
+
+The `config-policy-controller` would list the provided `Kind` across all namespaces using the
+provided `resourceSelector` and iterate over each maching object in each namespace to determine
```
mprahl, line 99:
```suggestion
provided `resourceSelector` and iterate over each matching object in each namespace to determine
```

enhancements/sig-policy/128-resource-selector/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+title: Resource selector for ConfigurationPolicy
+authors:
+  - "@dhaiducek"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2024-12-23
+last-updated: 2024-12-23
```
mprahl, line 9:
```suggestion
creation-date: 2024-10-23
last-updated: 2024-10-23
```

enhancements/sig-policy/128-resource-selector/README.md
dhaiducek: Right, or the current namespace template variable if that's implemented. 🙂

mprahl (approved):
/unhold

PR #131, opened 2024-10-28
Additional complianceTypes for ConfigurationPolicy

JustinKuli:
Refs:
 - https://issues.redhat.com/browse/ACM-14825

### DIFF
--- a/enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
+++ b/enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
@@ -1,0 +1,376 @@
+# ConfigurationPolicy ComplianceTypes for different merge behaviors
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+We propose to add new options for the `spec["object-templates][*].complianceType` field in
+ConfigurationPolicies to allow the user to specify new merge strategies: `musthaveapply`, 
+`musthavestrategic` and `musthavemerge`. These new options correspond to existing options in
+`kubectl`, which users may be familiar with. Along with the existing `musthave`, `mustonlyhave`, and
+`mustnothave` options, this should provide more flexibility and control to users, and be able to
+provide a more predictable experience.
+
+## Motivation
+
+The existing `musthave` complianceType has a bespoke algorithm for merging content from the cluster
+state with what is defined in the policy. One specific difference is when the policy includes a list
+in the object template: `musthave` will check the existing list for any items specified in the
+policy, and if any are missing, it will *add* them to the list and apply the change. Tools like
+`kubectl apply` can do something similar on some objects, via a "strategic merge", but not on custom
+resources.
+
+The existing `mustonlyhave` complianceType is simpler to understand: any fields on the object not
+defined in the policy will be removed from the object on the cluster (possibly just reverting them
+to default values). This can be used to avoid other merge behaviors, but requires the policy author
+to include more of the object in the policy, and does not allow the unset fields to be set by other
+controllers. This complianceType has been known to cause issues by repeatedly updating objects on
+the cluster, causing high API server load.
+
+New complianceTypes would give more tools to policy authors to specify the exact behavior they want.
+
+### Goals
+
+1. Policy authors should be able to create a policy to *replace* a specific list in an object on the
+   cluster with what is defined in the policy, *without* that policy reverting other customizations
+   on the object.
+2. Policy authors should be able to create a policy that has the exact same behavior as `kubectl`
+   commands like `apply`, `patch --type=strategic` and `patch --type=merge`.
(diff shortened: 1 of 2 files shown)


enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
```diff
@@ -0,0 +1,376 @@
...
+`mustnothave` options, this should provide more flexibility and control to users, and be able to
+provide a more predictable experience.
+
+## Motivation
+
+The existing `musthave` complianceType has a bespoke algorithm for merging content from the cluster
+state with what is defined in the policy. One specific difference is when the policy includes a list
+in the object template: `musthave` will check the existing list for any items specified in the
+policy, and if any are missing, it will *add* them to the list and apply the change. Tools like
+`kubectl apply` can do something similar on some objects, via a "strategic merge", but not on custom
```
mprahl, line 26: I think strategic merge does work on custom resources:
https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch

mprahl: I take that back. I got confused with this:
https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
```diff
@@ -0,0 +1,376 @@
...
+1. Support strategic merge on custom resources (`kubectl` does not support this).
+
+## Proposal
+
+Add a `musthavemerge` complianceType, corresponding to the `kubectl` command `patch --type=merge`.
+In particular, this allows for replacement of specific fields / objects / lists in an object without
+reverting other "unrelated" changes in the object. This merge strategy works on all resource types,
+including custom ones.
+
+Add a `musthavestrategic` complianceType, corresponding to `kubectl patch --type=strategic`. This
```
mprahl, line 58: What would happen if the user specified this and it was applied on a custom resource? Would it fall back to merge?

enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
```diff
@@ -0,0 +1,376 @@
...
+strategy works only on "built-in" types on the cluster, but allows for more intelligent merges using
+"patchStrategy" information available in those type definitions. For example, it allows for adding
+an additional environment variable to a Deployment, as long as it has a new name.
+
+Add a `musthaveapply` complianceType, corresponding to `kubectl apply`. The patch type is chosen
```
mprahl, line 63: I'm thinking that it may be better to just pick this option and not have the other two options since I don't really see a reason to have merge behavior when strategic merge is available.

What do you think?

enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
JustinKuli: Something like `kubectl patch operatorpolicy --type=strategic ...` fails with an error mentioning that `application/strategic-merge-patch+json is not supported`.

enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
JustinKuli: Oh, good to know that server-side-apply can do strategic merges on custom resources, although it begs the question of why exactly the client-side can't do the same thing - it should have access to the same openapi information from the CRDs...

I'll have to try server-side-apply more, but from a quick read I'd be worried it would have more "conflicts" than we'd want?

enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
JustinKuli: Strategic merge is **very** similar to the bespoke `musthave`. The `name` key that `musthave` uses as a mergeKey is the most common one I found in the API, so it covers a good number of cases. The biggest difference is that `musthave` works on custom resources.

If I had to pick only one new option, I'd actually prefer `musthavemerge`. Its behavior is the most different from what we currently have, and I think story 1 makes a good case for why it would be better than `mustonlyhave`.

mprahl: I see what you mean. Perhaps both does make sense.

enhancements/sig-policy/130-policy-merge-compliancetypes/README.md
```diff
@@ -0,0 +1,376 @@
...
+   on the object.
+2. Policy authors should be able to create a policy that has the exact same behavior as `kubectl`
+   commands like `apply`, `patch --type=strategic` and `patch --type=merge`.
+3. Clarify in documentation the exact behaviors of the complianceType options.
+
+### Non-Goals
+
+1. Support strategic merge on custom resources (`kubectl` does not support this).
+
+## Proposal
```
mprahl, line 51: Could you look into the options that Argo CD provides and why? I believe this is the diff code:
https://github.com/argoproj/gitops-engine/blob/72bcdda3f0a5b80432d9f72e5b30827a530ac349/pkg/diff/diff.go#L75

They seem to allow server side apply if asked for. They seem to try to use strategic merge on native Kubernetes types and fallback to JSON merge (see the `ThreeWayDiff` function, though we wouldn't provide last-applied-configuration and would just provide a copy of the desired like in the `TwoWayDiff` function).

PR #133, opened 2024-11-15
New Update Strategy only when workload changes in ManifestWork

qiujian16:
fixes https://github.com/open-cluster-management-io/enhancements/issues/132

### DIFF
--- a/enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
+++ b/enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
@@ -38,76 +38,77 @@ set.
 
 ## Proposal
 
-We would like to introduce new options in `Update` and `ServerSideApply` update strategy to resolve the above issue.
-So user can set the option in the manifestwork, so when another actor on the spoke cluster update the resource. The
+We would like to introduce new options and `ServerSideApply` update strategy to resolve the above issue.
+The user can set the option in the manifestwork, so when another actor on the spoke cluster update the resource. The
 work-agent will ignore the change and not try to change the resource back. On the other hand, when the resource spec
 defined in the manifestwork is updated by the user, the work-agent will still update the resource accordingly. In
-summary, the option is to ignore the change triggered from spoke cluster only.
+summary, the option is to ignore the change triggered from spoke cluster. 
 
-In addition, in `ServerSideApplly` strategy, we would also introduce a new `ignoreDifferences` options similar as what
+We would also introduce a new `ignoreFields` options similar as what
 is defined in argoCD (https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/). Such that user can choose to not
-let work-agent patch certain resource fields.
+let work-agent patch certain resource fields when it is changed on the managed cluster or in the manifestwork.
 
 ### Design Details
 
 #### API change
 
-The change will be added into the `updateStratey` field. For `Update` strategy, we will add an option
-
-```go
-type UpdateConfig struct {
-    // OnSpokeChange defines whether the resource should be overridden by the manifestwork it is changed
-    // on the spoke by another actor.
-    // +kubebuilder:default=Override
-    // +kubebuilder:validation:Enum=Override;NoOverride
-    // +kubebuilder:validation:Required
-    // +required
-	OnSpokeChange string `json:"onSpokeChange,omitempty"`
-}
-```
-
-For existing `ServerSideApply` strategy, we will add:
+The change will be added into the `updateStratey` field. For `ServerSideApply` strategy, we will add:
 
 ```go
 type ServerSideApplyConfig struct {
 	...
-	// IgnoreFields defines a list of json paths in the resource that will not be updated on the spoke.
-	IgnoreFields []string `json:"ignoreFields,omitempty"`
+    // IgnoreFields defines a list of json paths in the resource that will not be updated on the spoke.
+    // +listType:=map
+    // +listMapKey:=type
+    // +optional
+    IgnoreFields []IgnoreField `json:"ignoreFields,omitempty"`
+}
 
-	// OnSpokeChange defines whether the resource should be overriden by the manifestwork it is changed
-	// on the spoke by another actor.
-	// +kubebuilder:default=Override
-	// +kubebuilder:validation:Enum=Override;NoOverride
-	// +kubebuilder:validation:Required
-	// +required
-	OnSpokeChange string `json:"onSpokeChange,omitempty"`
+type IgnoreField struct {
+    // Condition defines the condition that the fields should be ignored when apply the resource. All the fields
+	// will be ignored if the condition is met, otherwise all fields will be updated.
+    // +kubebuilder:default=OnSpokePresent
+    // +kubebuilder:validation:Enum=OnSpokePresent;OnSpokeChange
+    // +kubebuilder:validation:Required
+    // +required
+    Condition string `json:"condition"`
+
+    // JSONPaths defines the list of json path in the resource to be ignored
+	JSONPaths []string `json:"jsonPaths"`
 }
 ```
 
+The `IgnoreFields` defines the setting that under certain situation, will not update the certain fields on the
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+}
+```
+
+#### agent implemntation
+
+When work-agent identity that the `OnSpokeChange` is set `NoOverride` for a certain resource in the `ManifestWork`, the agent
+when apply the resource to the spoke cluster will add an annotation with the key `open-cluster-management.io/object-hash`.
+The value of the annotation is the computed hash of the resource spec in the `ManifestWork`. Later when another actor
+updates the resource, the work-agent will at first check if the object-hash mismatches with the current resource spec
+in the `ManifestWork`. If it is the same, the resource will not be updated so the change from spoke is ignored. When
```
haoqing0110, line 93: Should we explain the behavior of "Delete" when `OnSpokeChange` is set to `NoOverride` here? It seems when the resource is deleted, object-hash is removed and the resource will be applied again?

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+the resource in the manifestwork is updated, the annotation will not match which then trigger the work-agent to update.
+
+To handle the `IgnoreFields` for `ServerSideApply`, we will remove the fields defined in the `IgnoreFields` and then
+generates the apply patch. The objec-hash will also be computed considering the `IgnoreFields`. 
+
+#### examples
+
+To resolve issue 642, user can set the manifestwork with `OnSpokeChange` set to `NoOverride`
```
mikeshng, line 101:
```suggestion
To resolve [issue 631](https://github.com/open-cluster-management-io/ocm/issues/631), user can set the manifestwork with `OnSpokeChange` set to `NoOverride`
```

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+        resource: secrets
+        namespace: default
+        name: some-secret
+      updateStrategy:
+        type: Update
+        update:
+          onSpokeChange: NoOverride
+```
+
+To resolve issue 670, user also can do the same for argocd application.
```
mikeshng, line 122:
```suggestion
To resolve [issue 670](https://github.com/open-cluster-management-io/ocm/issues/670), user also can do the same for argocd application.
```

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+        resource: application
+        namespace: default
+        name: application1
+      updateStrategy:
+        type: Update
+        update:
+          onSpokeChange: NoOverride
+```
+
+To resolve issue 690, user can set like:
```
mikeshng, line 144:
```suggestion
To resolve [issue 690](https://github.com/open-cluster-management-io/ocm/issues/690), user can set like:
```

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+	// +required
+	OnSpokeChange string `json:"onSpokeChange,omitempty"`
+}
+```
+
+#### agent implemntation
+
+When work-agent identity that the `OnSpokeChange` is set `NoOverride` for a certain resource in the `ManifestWork`, the agent
+when apply the resource to the spoke cluster will add an annotation with the key `open-cluster-management.io/object-hash`.
+The value of the annotation is the computed hash of the resource spec in the `ManifestWork`. Later when another actor
```
mikeshng, line 91: > The value of the annotation is the computed hash of the resource spec in the `ManifestWork`
The `spec` here is not just the resource's spec field but the entire resource object right?
For the case of ArgoCD Application `operation` field. If I need to perform another `operation` what is the recommended method to do that? Is it adding an annotation to the Application resource? Thanks!

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+We would like to introduce new options in `Update` and `ServerSideApply` update strategy to resolve the above issue.
+So user can set the option in the manifestwork, so when another actor on the spoke cluster update the resource. The
+work-agent will ignore the change and not try to change the resource back. On the other hand, when the resource spec
+defined in the manifestwork is updated by the user, the work-agent will still update the resource accordingly. In
+summary, the option is to ignore the change triggered from spoke cluster only.
+
+In addition, in `ServerSideApplly` strategy, we would also introduce a new `ignoreDifferences` options similar as what
+is defined in argoCD (https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/). Such that user can choose to not
+let work-agent patch certain resource fields.
```
skeeey, line 49: This will also be able to resolve the zero creationTime in metadata problem, right?

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,183 @@
...
+}
+```
+
+For existing `ServerSideApply` strategy, we will add:
+
+```go
+type ServerSideApplyConfig struct {
+	...
+	// IgnoreFields defines a list of json paths in the resource that will not be updated on the spoke.
+	IgnoreFields []string `json:"ignoreFields,omitempty"`
```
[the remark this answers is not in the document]
qiujian16: yes, so when something changes in ignoreFields in the mw, the work-agent will not update resource all the time.

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
qiujian16: right, it is the the whole resource payload also include metadata and other fields.

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
qiujian16: the resource itself will be deleted, so does the annotation. The annotation will be kept if the deleteOption is Orphan, and in this case, I do not think we need to remove the annotation.

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Design Details
+
+#### API change
+
+The change will be added into the `updateStratey` field. For `Update` strategy, we will add an option
+
+```go
+type UpdateConfig struct {
+    // OnSpokeChange defines whether the resource should be overriden by the manifestwork it is changed
```
zhujian7, line 59:
```suggestion
    // OnSpokeChange defines whether the resource should be overridden by the manifestwork once it is changed
```

enhancements/sig-architecture/132-work-update-only-when-spec-change/README.md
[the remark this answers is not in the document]
zhujian7: I still don't quite understand, will IgnoreFields ignore OnSpokeChange? No matter OnSpokeChange is Override or NoOverride, changes on the manifestworks will not updated to the spoke? why do we need the `OnSpokeChange` for `ServerSideApply`?

qiujian16:
@zhujian7 

< I still don't quite understand, will IgnoreFields ignore OnSpokeChange? No matter OnSpokeChange is Override or NoOverride, changes on the manifestworks will not updated to the spoke? why do we need the OnSpokeChange for ServerSideApply?

with OnSpokeChange, the work-agent will still update manifests when the resource in the manifestwork is updated.   IgnoreFields makes a finer granularity control on certain fields in the resource, so that even if these fields is changed in manifestwork resource or spoke, the work-agent will not update the resource.

PR #135, opened 2024-12-06
🌱 Update addon template design to support specifying signerCA namespace

zhujian7:
https://github.com/open-cluster-management-io/ocm/issues/688


qiujian16:
also needs to update the metadata.xml with updated lastModified

Issue #136, opened 2024-12-17
Support CEL expressions in placement cluster selector


haoqing0110:
To support requirements like:
- select clusters whose label version > 1.13.1.
- select clusters whose label or claim value matches some regex format. 
- select clusters whose addon score is greater than or less than some value.

PR #137, opened 2024-12-18
select clusters with CEL expressions

haoqing0110:
ref: https://github.com/open-cluster-management-io/enhancements/issues/136

### DIFF
--- a/enhancements/sig-architecture/136-placement-cel-selector/README.md
+++ b/enhancements/sig-architecture/136-placement-cel-selector/README.md
@@ -1,0 +1,321 @@
+# Select clusters with CEL expressions
+
+## Release Signoff Checklist
+
+- [] Enhancement is `provisional`
+- [] Design details are appropriately documented from clear requirements
+- [] Test plan is defined
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal proposes an enhancement to the placement cluster selector, enabling users to select clusters with CEL expressions.
+
+## Motivation
+
+In the OCM placement predicate stage, the user can currently use a label selector or claim selector to selector clusters. For detailed usage see [Label/Claim selection](https://open-cluster-management.io/docs/concepts/placement/#predicates). It can meet most of the use cases. However, there are some more advanced needs recently that the current functionality cannot fulfill. For example, the user wants to filter clusters with label "version" > 1.30.0. It's not smart to list every supported version in the label selector. Another example is bin packing, the user may want to filter clusters with specific resource quantities before prioritizing, the resource quantity will come from another resource, eg `AddonPlacementScore` but not labels or claims. So we have this proposal to support selecting clusters with [CEL expressions](https://cel.dev/).
+
+### Goals
+
+- Enhance `Placement` API to allow users to select clusters with CEL expressions.
+
+### Non-Goals
+
+- Enhance the `AddonPlacementScore` to include the original resource value for selecting clusters that is not in the scope.
+
+## Proposal
+
+In this proposal, we will introduce the `celSelector` in the predicate section. It will be a list of `celExpressions`. The result of the expressions are ANDed. The result of `labelSelector`, `claimSelector` and `celSelector` is also ANDed.
+
+We will also introduce a CEL library, `managedCluster` is declared as a variable and can be used in CEL expressions. Custom functions like comparing versions and getting cluster resources will also be added.
+
+### User Stories
+
+1. The user can select clusters by fields other than labels and claims. For example, select clusters by field `managedCluster.Status.version.kubernetes`.
+2. The user can use CEL [Standard macros](https://github.com/google/cel-spec/blob/master/doc/langdef.md#macros) and [Standard functions](https://github.com/google/cel-spec/blob/master/doc/langdef.md#standard-definitions) on the `managedCluster` fields. For example, select clusters with version 1.30.* and 1.31.* by regex `'^1\\.(30|31)\\.\\d+$'`.
+3. The user can use CEL customized function to select clusters by version comparison. For example, select clusters whose version > 1.13.0.
+4. The user can use CEL customized function to select clusters by `AddonPlacementScore`. For example, select clusters whose cpuAvailable < 2.
+
+## Design Details
+
+### API Changes
+
+Add the new field `celSelector` to the `clusterSelector`. `celSelector` will be list of `celExpressions` with string type.
+
+```go
+type ClusterSelector struct {
+	// LabelSelector represents a selector of ManagedClusters by label
+	// +optional
+	LabelSelector metav1.LabelSelector `json:"labelSelector,omitempty"`
+
+	// ClaimSelector represents a selector of ManagedClusters by clusterClaims in status
+	// +optional
+	ClaimSelector ClusterClaimSelector `json:"claimSelector,omitempty"`
+
+	// CelSelector represents a selector of ManagedClusters by CEL expressions on ManagedCluster fields
+	// +optional
+	CelSelector ClusterCelSelector `json:"celSelector,omitempty"`
+}
+
+// ClusterCelSelector is a list of CEL expressions. The expressions are ANDed.
+type ClusterCelSelector struct {
+	// +optional
+	CelExpressions []string `json:"celExpressions,omitempty"`
+}
+```
+
+### ManagedCluster CEL Library
+
+A ManagedClusterLib will be added.
+- Variable `managedCluster` will be added.
+- Function `scores` will be added. 
+- Function `versionIsGreaterThan`, `versionIsLessThan` will be added. 
+- Function `quantityIsGreaterThan`, `quantityIsLessThan` will be added. 
+
+```go
+// ManagedClusterLib defines the CEL library for ManagedCluster evaluation.
+// It provides functions and variables for evaluating ManagedCluster properties
+// and their associated resources.
+//
+// Variables:
+//
+// managedCluster
+//
+// Provides access to ManagedCluster properties.
+//
+// Functions:
+//
+// scores
+//
+// Returns a list of AddOnPlacementScoreItem for a given cluster and AddOnPlacementScore resource name.
+//
+//	scores(<ManagedCluster>, <string>) <list>
+//
+// The returned list contains maps with the following structure:
+//   - name: string - The name of the score
+//   - value: int - The numeric score value
+//   - quantity: number|string - The quantity value, represented as:
+//     * number: for pure decimal values (e.g., 3)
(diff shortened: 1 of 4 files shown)


enhancements/sig-architecture/136-placement-cel-selector/README.md
```diff
@@ -0,0 +1,235 @@
...
+}
+```
+
+### ManagedCluster CEL Library
+
+A ManagedClusterLib will be added.
+- Variable `managedCluster` will be added.
+The input types may either be instances of `proto.Message` or `ref.Type`. Since OCM API does not implement `proto.Message`, we will add a function called ConvertManagedCluster() to convert ManagedCluster to a Map as an easier way to get started.
+
+- Function `score` will be added.
```
haoqing0110, line 74: Need to enhance `AddonPlacementScore` to add original score value. And aeed a more appropriate name than `score`.

enhancements/sig-architecture/136-placement-cel-selector/README.md
```diff
@@ -0,0 +1,235 @@
...
+  namespace: default
+spec:
+  numberOfClusters: 3
+  clusterSets:
+    - prod
+  predicates:
+    - requiredClusterSelector:
+        celSelector:
+          celExpressions:
+            - managedCluster.score("default, "cpuAvailable") < 2
```
qiujian16, line 218: could we do like managedCluster.score.default.cpuAvailable? or we have to use a function here?

enhancements/sig-architecture/136-placement-cel-selector/README.md
```diff
@@ -0,0 +1,235 @@
...
+  namespace: default
+spec:
+  numberOfClusters: 3
+  clusterSets:
+    - prod
+  predicates:
+    - requiredClusterSelector:
+        celSelector:
+          celExpressions:
+            - managedCluster.metadata.labels["version"].versionIsGreaterThan("1.30.0")
```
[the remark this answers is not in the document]
haoqing0110: Should support both Semantic Versioning and v as a prefix, will update the detail in the proposal.

enhancements/sig-architecture/136-placement-cel-selector/README.md
haoqing0110: managedCluster does not have a score field so if we pass `managedCluster.score.default.cpuAvailable` to the cel for evaluation, it can not recognize it.

Maybe we can use something like managedCluster.score(default).cpuAvailable. function score("crname") will return a map of [scorename]scorevalue, and user can use standard cel expression to select it.

qiujian16: what if we do clusterScore instead of managedCluster instead? will it work?

haoqing0110: Hmm, have considered about this, when evaluate a expression is true or false, the clusterScore may still need a cluster name as input, managedCluster.score("default") equals to score(managedCluster, "default") (but can not write a cel expression like this).

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
```diff
@@ -110,6 +111,10 @@ type AddOnPlacementScoreItem struct {
 	// +kubebuilder:validation:Maximum:=100
 	// +required
 	Value int32 `json:"value"`
+
+	// Quantity defines the original value of the score.
+	// +optional
+	Quantity resource.Quantity `json:"quantity"`
```
qiujian16, line 117: this could be quite noisy, brings frequent update on the API. We would need to consider how this should be rate limited.

enhancements/sig-architecture/136-placement-cel-selector/README.md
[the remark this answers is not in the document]
qiujian16: how about we change this example using managedCluster.Status.version.kubernetes instead?

enhancements/sig-architecture/32-extensiblescheduling/32-extensiblescheduling.md
haoqing0110: Yes, good suggestion. Add more explanation for the new field:  
1. It should be updated together with the value of the score. 
2. For the resources that are updated frequently, suggest to update at a reasonable interval, eg, 60s. 
3. An expired score will be treated as value 0 and quantity 0 in the placement controller.

deads2k (commented):
This is a very neat idea and I think the idea is good.  I think that adding a few things could make this easier to monitor at runtime and mitigate potential issues ahead of time.

1. I didn't see a direct mention of metrics for time spent evaluating CEL at runtime.  Adding those will provide runtime visibility
2. cost limiting on CEL expressions similar to what we do on CRD validation is a good idea to prevent accidental excessive costs.  I think the cost estimator can be configured with expected cardinality of things like `scores`
3. idea we haven't used before (I think because we haven't done controller-y things), is a budget of how much time we'll spend on CEL at runtime in a given unit of time.  Just an interesting idea if it was easy.

enhancements/sig-architecture/136-placement-cel-selector/README.md
```diff
@@ -0,0 +1,343 @@
...
+//
+// Returns true if the first quantity string is less than the second quantity string.
+//
+//	quantityIsLessThan(<string>, <string>) <bool>
+//
+// Examples:
+//
+//	quantityIsLessThan("1Gi", "2Gi") // returns true
+//	quantityIsLessThan("2Gi", "1Gi") // returns false
+//	quantityIsLessThan("1000Mi", "1Gi") // returns true
```
jpbetz, line 158: Can https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/library/quantity.go be leveraged here?  It provides comparisons.

enhancements/sig-architecture/136-placement-cel-selector/README.md
```diff
@@ -0,0 +1,343 @@
...
+// Returns true if the first version string is less than the second version string.
+// The version must follow Semantic Versioning specification (http://semver.org/).
+// It can be with or without 'v' prefix (eg, "1.14.3" or "v1.14.3").
+//
+//	versionIsLessThan(<string>, <string>) <bool>
+//
+// Examples:
+//
+//	versionIsLessThan("1.24.0", "1.25.0") // returns true
+//	versionIsLessThan("1.25.0", "1.24.0") // returns false
```
jpbetz, line 132: Can the k8s semver CEL library be used to handle these use cases? https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/library/semverlib.go

enhancements/sig-architecture/136-placement-cel-selector/README.md
```diff
@@ -0,0 +1,343 @@
...
+	// +optional
+	LabelSelector metav1.LabelSelector `json:"labelSelector,omitempty"`
+
+	// ClaimSelector represents a selector of ManagedClusters by clusterClaims in status
+	// +optional
+	ClaimSelector ClusterClaimSelector `json:"claimSelector,omitempty"`
+
+	// CelSelector represents a selector of ManagedClusters by CEL expressions on ManagedCluster fields
+	// +optional
+	CelSelector ClusterCelSelector `json:"celSelector,omitempty"`
```
jpbetz, line 58: Just yesterday a colleague mentioned to me they were adding a string field to a CRD that is expected to contain a CEL expression but that they had no good way to validate if the contents of the string was a valid CEL expression.

Does the same problem exist here?  It looks like it would.

cc @deads2k who has anticipated this need.

cc @cici37

enhancements/sig-architecture/136-placement-cel-selector/README.md
haoqing0110: yes, I think so. 
The previous concern it seems the expression will be a very long and complex string if use k8s library.  For example: 
```
managedCluster.scores("default").filter(s, s.name == 'memAvailable').all(e, e.quantity.quantityIsGreaterThan("100Mi"))
```
vs
```
managedCluster.scores("default").filter(s, s.name == 'memAvailable').all(e, quantity(e.quantity).isGreaterThan(quantity("100Mi"))
```
But when I'm writing to compare the two expressions today, it seems no big difference, both are fine. We can try to use the k8s library instead.

enhancements/sig-architecture/136-placement-cel-selector/README.md
haoqing0110: Yes, let me try.

enhancements/sig-architecture/136-placement-cel-selector/README.md
haoqing0110: The placement API has [condition](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1beta1/types_placement.go#L418 ) in status, so when the CEL evaluation failed, the error will return and report in the [PlacementMisconfigured](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1beta1/types_placement.go#L427C44-L427C66) condition with error message.

enhancements/sig-architecture/136-placement-cel-selector/README.md
haoqing0110: Hello @jpbetz it seems the Semver version with a v prefix is not supported in the library? See [this case](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/library/semver_test.go#L135).  What does ["tolerant parsing"](https://github.com/kubernetes/kubernetes/blob/v1.32.0/staging/src/k8s.io/apiserver/pkg/cel/library/semverlib.go#L55 ) mean and how to support v as a prefix?

enhancements/sig-architecture/136-placement-cel-selector/README.md
jpbetz: I've opened https://github.com/kubernetes/kubernetes/issues/130639 to track one small improvement that can be made here.

+100 to using the k8s library instead of forking off and starting a nearly identical one.

enhancements/sig-architecture/136-placement-cel-selector/README.md
jpbetz: Looks like the underlying library has a [ParseTolerant](https://github.com/kubernetes/kubernetes/blob/fe7af1c68bc0c772e78d57b5c38ac4ae8ef122ab/vendor/github.com/blang/semver/v4/semver.go#L240) function that: "currently trims spaces, removes a "v" prefix, adds a 0 patch number to versions with only major and minor components specified, and removes leading 0s."

I'm not sure why it is alluded in the semver CEL library since it doesn't call into it.

I've opened https://github.com/kubernetes/kubernetes/pull/130648 to add this support directly to the library.

As a short term mitigation something like`"v1.0.0".find("[^v].*")` should work.

qiujian16 (commented):
/approve
/lgtm

Issue #138, opened 2025-02-12
Support for batch workloads


bhperry:
# Issue
Currently there is not good support for workloads which are expected to complete and be deleted from the spoke cluster where they were assigned. For example, simply creating a ManifestWork that wraps a Job with `ttlSecondsAfterFinished` set will always attempt to re-create the job after it is deleted by the spoke cluster's job controller. As far as I can tell, there is currently no way to express that a ManifestWork should be deleted when its subresources have completed.

~For CronJobs it would be useful to be able to make new placement decisions every time a Job is scheduled. Wrapping a CronJob in ManifestWork will of course only schedule jobs on a single spoke cluster, whereas wrapping it in a ManifestWorkReplicaSet will not have good guarantees of properly following the schedule when migrating between spoke clusters (e.g. if workload is migrated from cluster-A to cluster-B around the time the schedule would have triggered, it may miss that execution entirely and have to wait for the next scheduled time before running).~

# Possible Solutions
To address the issue of completable workloads like Jobs or non-restartable Pods, one possible solution would be to add additional fields to ManifestWork that instruct it to inspect one or more of its workloads for completion (under `spec.manifestConfigs`). Completed workloads would never be restarted, and when they have all reached their completed state the associated ManifestWork would be deleted. (e.g. a ManifestWork with a Job and a ConfigMap could be configured to wait until the Job completes, then be deleted along with the Job and ConfigMap). Just as a rough idea borrowing from the format of feedbackRules, it could look something like:
```
spec:
  manifestConfigs:
    - resourceIdentifier:
        group: batch
        resource: jobs
        namespace: ns
        name: myjob
      completionRules:
        - type: WellKnownCompletions
        - type: JSONPaths
          jsonPaths:
           - path: ".status.conditions[?(@.type=='Complete')].status"
             value: "True"
```
Could also consider adding a type `Deleted` which will consider a workload to be complete when it is deleted by another controller (e.g. the spoke cluster's job controller based on a very short `ttlSecondsAfterFinished`). Properly handling completion rules (especially a `Deleted` rule) would likely require registering a finalizer on the associated objects in order to have better guarantees of reconciliation.

[...]

PR #139, opened 2025-02-24
Workload completion

bhperry:
Closes https://github.com/open-cluster-management-io/enhancements/issues/138

This PR does not address the multicluster CronJob scheduling portion. I will split that out into a different issue as they are separate concerns.

### DIFF
--- a/enhancements/sig-architecture/138-workload-completion/README.md
+++ b/enhancements/sig-architecture/138-workload-completion/README.md
@@ -58,6 +58,12 @@ If a user wishes to complete some of the manifests but not the entire manifestwo
 need to do is set a `Complete` condition rule with the CEL expression `false` for at least
 one other manifest to make the manifestwork incompletable.
 
+Custom condition messages may be provided on condition rules as a static string or as a CEL expression
+that evaluates to a string. If MessageExpression is set on a rule, in addition to the primary `object`
+variable it will also be given a `result` boolean variable with the computed value of the CEL expressions
+on the rule. If both Message and MessageExpression are set, the expression will take precedence as long
+as it returns a non-empty string.
+
 ### Design Details
 
 #### API change
@@ -93,9 +99,19 @@ type ConditionRule struct {
     Type ConditionRuleType `json:"type"`
 
     // CelExpressions defines the CEL expressions to be evaluated for the condition.
-    // Final result is the logical AND of all rules.
+    // Final result is the logical AND of all expressions.
     // +optional
     CelExpressions []CelConditionExpressions `json:"celExpressions"`
+    // Message is set on the condition created for this rule
+    // +optional
+    Message string `json:"message"`
+    // MessageExpression uses a CEL expression to generate a message for the condition
+    // Will override message if both are set and messageExpression returns a non-empty string.
+    // Variables:
+    // - object: The current instance of the manifest
+    // - result: Boolean result of the CEL expressions
+    // +optional
+    MessageExpression string `json:"messageExpression"`
 }
 
 // +kubebuilder:validation:Enum=WellKnownCompletions;CEL
@@ -160,7 +176,8 @@ var jobConditionRule = workapiv1.ConditionRule{
     {
       Expression: "object.status.conditions.filter(c, c.type == 'Complete' || c.type == 'Failed').exists(c, c.status == 'True')",
     },
-  }
+  },
+  MessageExpression: `result ? "Job is finished" : "Job is not finished`
 }
 
 var podConditionRule = workapiv1.ConditionRule{
@@ -170,7 +187,8 @@ var podConditionRule = workapiv1.ConditionRule{
     {
       Expression: "object.status.phase in ['Succeeded', 'Failed']",
     },
-  }
+  },
+  MessageExpression: `result ? ("Pod " + object.status.phase) : "Pod is not finished`
 }
 ```
 
@@ -279,6 +297,8 @@ spec:
               ).exists(
                 c, c.status == 'True'
               )
+          messageExpression: |
+            result ? "Object is complete" : "Object is not complete"
         - condition: MyCondition
           type: CEL
           celExpressions:
@@ -286,6 +306,10 @@ spec:
               object.status.conditions.exists(
                 c, c.type == 'MyCondition' && c.status == 'True'
               )
+          messageExpression: |
+            result
+              ? object.status.conditions.filter(c, c.type == 'MyCondition')[0].message
+              : "Object does not have MyCondition"
 ```
 
 Status conditions
--- a/enhancements/sig-architecture/138-workload-completion/README.md
+++ b/enhancements/sig-architecture/138-workload-completion/README.md
@@ -10,92 +10,118 @@
 
 ## Summary
 
-This proposal introduces completion rules in manifestwork to identify that one or more of its workloads has completed
-and should not be updated or recreated. Automated garbage collection can optionally be enabled with a TTL to delete the entire manifestwork
-once all of its completion rules have been satisfied.
+This proposal introduces condition rules in manifestwork to extract status conditions from
+resources it controls and orchestrate higher level behaviors. The first condition to be managed
+in this way will be the `Complete` condition, which identifies that one or more of its workloads
+has completed and should not be updated or recreated. Automated garbage collection can optionally
+be enabled with a TTL to delete the entire manifestwork once all of its `Complete` condition rules
+have been satisfied.
 
 ## Motivation
 
-OCM currently does not have support for workloads that are expected to complete and be deleted from spoke clusters.
-Often a user wants to run a workload exactly once, and then have the resources they created be automatically garbage collected.
+OCM currently does not have support for workloads that are expected to complete and be deleted
+from spoke clusters. Often a user wants to run a workload exactly once, and then have the
+resources they created be automatically garbage collected.
 
 Examples:
-- Job with `ttlSecondsAfterFinished` set would get cleaned up by the spoke's job controller after completion
-- Pod with `restartPolicy` set to `Never` or `OnFailure` that eventually exits, and is deleted along with its node by the cluster-autoscaler
+- Job with `ttlSecondsAfterFinished` set would get cleaned up by the spoke's job controller after
+  completion
+- Pod with `restartPolicy` set to `Never` or `OnFailure` that eventually exits, and is deleted
+  along with its node by the cluster-autoscaler
 
-In both cases, OCM would recreate the deleted resources so long as the manifestwork still exists, while the intention
-is most likely to only run them once.
+In both cases, OCM would recreate the deleted resources so long as the manifestwork still exists,
+while the intention is most likely to only run them once.
+
+There may be additional conditions in future that the manifestwork controller, or others built
+on top of it, want to use as triggers for various behaviors. With that in mind, this feature is
+designed in a generic way such that any condition can be specified.
 
 ## Proposal
 
-Add new `CompletionRules` to `ManifestConfigs` to identify that a particular workload is completable,
-and how to determine when it is done. `ManifestCondition` would be updated with a new `StatusCompletions`
-containing the status of each completion rule for the manifest. A new field `TTLSecondsAfterFinished` will be
-added to `DeleteOption` which enables garbage collection for the entire manifestwork after all completion rules
-have been satisfied. If a TTL is set with no completion rules, then the manifestwork will never be considered
-finished and therefore never be garbage collected. Likewise if completion rules are set but not TTL, the manifestwork
-will complete but not be automatically deleted.
+Add new `ConditionRules` to `ManifestConfigs` to set certain conditions on the ManifestWork based
+on its resources. Certain conditions may control higher level behaviors of the ManifestWork.
+The first condition rule to be supported will be `Complete`, which identifies that a particular
+workload is completable, and how to determine when it is done. A new field `TTLSecondsAfterFinished`
+will be added to `DeleteOption` which enables garbage collection for the entire manifestwork after
+all `Complete` conditions have been satisfied. If a TTL is set with no `Complete` condition rules,
+then the manifestwork will never be considered finished and therefore never be garbage collected.
+Likewise if `Complete` condition rules are set but not TTL, the manifestwork will complete but
+not be automatically deleted.
 
-Once a particular manifest has been marked as completed, it is no longer elligible to be updated or recreated
-by the spoke controller, regardless of the completion status of other manifests in the manifestwork. When all
-manifests with completion rules have completed, the entire manifestwork is considered to be complete (including
-manifests that did NOT specify completion rules), and no more updates will be made.
+Once a particular manifest has been marked as completed, it is no longer elligible to be updated or
+recreated by the spoke controller, regardless of the completion status of other manifests in the
+manifestwork. When all manifests with `Complete` condition rules have completed, the entire manifestwork is
+considered to be complete (including manifests that did NOT specify `Complete` rules), and no more
+updates will be applied.
+
+If a user wishes to complete some of the manifests but not the entire manifestwork, all they would
+need to do is set a `Complete` condition rule with the CEL expression `false` for at least
+one other manifest to make the manifestwork incompletable.
 
 ### Design Details
 
 #### API change
 
-To configure completion of a manifestwork, `completionRules` will be added to `manifestConfigs`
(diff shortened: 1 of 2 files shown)


mikeshng:
Related to https://github.com/open-cluster-management-io/ocm/issues/842

mikeshng:
@bhperry could you please sign-off on your commit? 

```
git commit -s --amend
git push -f origin workload-completion
```

mikeshng:
CC work maintainers: @elgnay @zhujian7 @zhiweiyin318

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,363 @@
...
+
+```go
+type DeleteOption struct {
+  ...
+
+  // TTLSecondsAfterFinished limits the lifetime of a ManifestWork that has satisfied
+  // all completionRules configured for its manifests. If this field is set, and
+  // the manifestwork has completed, then it is elligible to be automatically deleted.
+  // If this field is unset, the manifestwork won't be automatically deleted even afer completion.
+  // If this field is set to zero, the manfiestwork becomes elligible to be deleted immediately
```
elgnay, line 134: Will the ManifestWork itself be deleted, or will the K8s resources specified in its manifests be deleted from the managed cluster instead?

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,363 @@
...
+add `completionRules` to any of the manifests that are required to complete before the `ManifestWork` should be considered complete.
+
+### Version Skew Strategy
+- The new fields are optional, and if not set, the manifestwork will be handled the same as by previous versions
+- Older versions of the work agent will ignore the newly added fields
+
+## Alternatives
+
+- Instead of adding the `JsonPathsOperator` to support logical OR rules, we could wait for CEL support in order to handle these
+types of expressions. The downside is that WellKnownCompletions for Jobs would not be expressable as a standard rule until the CEL work is complete, but considering that feature is planned for v1.0.0 it could be worth it to keep JsonPaths simple and avoid cluterring the API.
```
qiujian16, line 345: I actually like the alternative approach better.

qiujian16 (commented):
I consider this as two steps:

1. an extensible way to mark the status of the Manifestwork, e.g. CEL. Which means a user can define when to set Completed status as true/false with cel expression. And the API should be extensible enough that we could also allow user to set Available/Degraded status in the future.
2. Agent/server should have some logic to treat competed status. 

I think 2 is very easy to treat, and I am more interested in building an API that could allow user to define Available/Degraded/Complated status conditions.

cc @zhujian7 @elgnay

enhancements/sig-architecture/138-workload-completion/README.md
bhperry: The ManifestWork itself will be deleted, just like a Job with ttlSecondsAfterFinished.

The k8s resources specified in its manifests will respect the configured deletion options as before. So by default they will be deleted as well.

enhancements/sig-architecture/138-workload-completion/README.md
bhperry: The more I think about it I do agree CEL is a much better fit for this. JSONPaths is nice and simple to use, but it's just not flexible enough for the task.

bhperry:
> 1. an extensible way to mark the status of the Manifestwork, e.g. CEL. Which means a user can define when to set Completed status as true/false with cel expression. And the API should be extensible enough that we could also allow user to set Available/Degraded status in the future.

Interesting idea. In my initial design I wanted to focus only on the task of marking completion, and considered things like "Available" vs. "Degraded" you mentioned to be the job of the feedbackRules. But I can potentially see the utility in being able to set arbitrary conditions with this mechanic.

Need to think about this a bit more, but it could look something like:
```yaml
conditionRules:
- type: CEL
  condition: Completed
  celSelectors:
  - expression: ...
```


My main question is, would ManifestWork do anything in response to the Available/Degraded conditions, or is it just another way to provide feedback to the users? Either way, I do like the idea of making this open to be extended for additional functionality in the future.

enhancements/sig-architecture/138-workload-completion/README.md
bhperry: Would you want to see support for JSONPaths as a simple option for users in addition to CEL, or do you think it should only support CEL (and WellKnown... variants)?

bhperry: I guess there's really no reason to have JSONPaths other than that kubernetes users may be more familiar with it. 

But the simple cases like traversing an object path can be written pretty much identically in CEL. And more complicated cases like filtering a list are much easier.

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,316 @@
...
+    WellKnownCompletionsType CompletionType = "WellKnownCompletions"
+
+    CelCompletionExpressionsType CompletionType = "CEL"
+)
+
+type CelCompletionExpressions struct {
+    // Expression represents the CEL expression to be evaluated on the manifest.
+    // The expression must evaluate to a single value  in the type of integer, bool or string.
+    // If the expression evaluates to a structure, map or slice, the manifest will not be considered complete.
+    // Ref to https://cel.dev/ on how to write CEL
```
bhperry, line 91: Not sure if the current CEL selectors work intends to use the kubernetes CEL libraries that extend the standard CEL lib. If so, can link to those docs instead https://kubernetes.io/docs/reference/using-api/cel/

Does look like there are some useful things in there, such as support for k8s Quantity parsing and comparisons.

qiujian16: cc @haoqing0110

qiujian16:
> I do see now that Available and Degraded are existing status conditions used by the controller. So then custom rules would simply override the default behavior for setting those conditions? Probably doesn't have to be worked out now, if I understand correctly you just want the API to be able to support those in future but not necessarily add them as part of this enhancement.

exactly, I want this to be more extensible. The available/degraded condition could be used by other higher level APIs such as ManagedClusterAddon to reflect the addon status or ManifestworkRelicaSet (we do define degraded field in ManifestworkReplicaset but not implemented yet).  And it is not needed to be covered in this proposal.

enhancements/sig-architecture/138-workload-completion/README.md
qiujian16: I think JSONPaths looks simple, but its capability is limited to be use in real use case.

bhperry: Agreed. I was just more familiar with JSONPaths myself, but as I have been experimenting with CEL a bit I can immediately see it is a much better fit. Rremoved JSONPaths from the design.

bhperry:
@qiujian16 Updated proposal to support generic conditions as suggested

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,429 @@
...
+    // +kubebuilder:validation:Required
+    // +required
+    Expression string `json:"expression"`
+    // Value represents the expected value of the expression to set the condition's status to True.
+    // The value will be parsed as an integer, bool, or string depending on the type of the field retrieved from path.
+    // If the value is empty and expression evaluates to bool, it will be compared to boolean `true` by default
+    // If parsing to the appropriate type fails, the condition's status will be False.
+    // +optional
+    Value string `json:"value"`
```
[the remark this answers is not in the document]
bhperry: I did consider adding a message field here, however the issue is it would be static. I proposed an alternative where we accept a [structured object from CEL](https://github.com/open-cluster-management-io/enhancements/pull/139/files#diff-7874cafb0384f9512de4339d4682ccb35aa795abac9cfc7510ec41caf5981ea0R383) so that the message can be changed depending on state.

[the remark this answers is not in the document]
bhperry: Yes, could require the cel expression evaluate to a boolean.

I was thinking having Value here would be easier to update with a patch req if you wanted to change expected result of the condition. But that's a niche case, and considering Value only supports strict equality probably not worth it.

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,429 @@
...
+
+// +kubebuilder:validation:XValidation:rule="self.type != 'CEL' || self.condition != ''",message="Condition is required for CEL rules"
+type ConditionRule struct {
+    // Condition is the type of condition that is set based on this rule.
+    // Any condition is supported, but certain special conditions can be used to
+    // to control higher level behaviors of the manifestwork.
+    // If the condition is Complete, the manifest will no longer be updated once completed.
+    // Required for CEL rules, WellKnownCompletions will default to the "Complete" condition
+    // +optional
+    Condition string `json:"condition"`
```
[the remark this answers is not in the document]
bhperry: I proposed leaving it open to arbitrary conditions because that way users building custom addons or higher level manifestwork controllers can make use of their own conditions with no changes to core OCM.

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,429 @@
...
+```
+
+WellKnownCompletions rules for Job and Pod
+```go
+var jobConditionRule = workapiv1.ConditionRule{
+  Condition: workapiv1.ManifestComplete,
+  Type: workapiv1.CelConditionExpressionsType,
+  CelExpressions: []workapiv1.CelConditionExpressions{
+    {
+      Expression: ".status.conditions.filter(c, c.type == 'Complete' || c.type == 'Failed').exists(c, c.status == 'True')",
```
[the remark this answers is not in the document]
bhperry: Would be nice to leave the root namespace open in case there are other objects in future that need to be referenced here. Like `managecluster.spec....`, allowing cross reference between manifests in the manifestwork.


Ideally, would be good to have the same reference pattern for all uses of CEL.

enhancements/sig-architecture/138-workload-completion/README.md
```diff
@@ -0,0 +1,429 @@
...
+Conditions will be evaluated after applying, however `Complete` is a special case
+that needs to be evaluated by any applier that updates existing manifests
+(currently ServerSideApply and UpdateApply). In these cases, the `Complete` condition will be
+evaluated prior to update on the existing resource, and if it passes then no update will be applied.
+This is required in order to provide guarantees that a manifest will not be updated once
+it has completed. ServerSideApply is the only case where there is a potential for a race condition
+on update, since it does not perform optimistic locking by resourceVersion like UpdateApply does.
+Because UpdateApply uses the openshift `resourceapply` package, it will use a wrapper on the ResourceCache
+interface that overrides `SafeToSkipApply(required runtime.Object, existing runtime.Object) bool`
+in order to evaluate the `Complete` condition before updating.
```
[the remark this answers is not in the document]
bhperry: ### Scenarios:
1. Object is complete (not yet observed by manifestwork)
    1. Update would change complete condition to false
        - Checking before apply would ignore the change and mark manifest complete. Depending on expectations of the user, this may or may not be desirable.
        - Applying first would make the change and continue watching the object
    2. Update is invalid due to object being completed
        - I don't think this exists for native k8s objects. But a validating webhook could be configured that has different rules depending on the state of the object.
        - Checking before apply would skip the invalid change
        - Applying first would fail and result in an error
     3. Update is valid, and would not change condition
         - Checking before apply would skip the change
         - Applying first would make the change, then stop watching the object

2. Object is not complete
    1. Update would set complete condition to true
        - Checking before apply is not necessary
    2.  Update would not change condition
        - Checking before apply is not necessary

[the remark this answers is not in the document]
bhperry: For reference, the [k8s job controller](https://github.com/kubernetes/kubernetes/blob/67e1a1ebbb926e720e7bd1ac7e0cac0436e26f2b/pkg/controller/job/job_controller.go#L838) does check if it has finished before doing any reconciliation. Manages job pods only after checking for existing conditions, and then re-evaluating completed/failed conditions based on existing pods. Although it is slightly different since job forbids updates to its pod template after creation.

enhancements/sig-architecture/138-workload-completion/README.md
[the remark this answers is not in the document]
qiujian16: could consider the mechanism as here https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#message-expression

enhancements/sig-architecture/138-workload-completion/README.md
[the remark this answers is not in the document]
qiujian16: I see, let's start with simplicity.

qiujian16:
I really like the the current proposal, I think we can merge it with some tweaks. Some thing that we also need to consider is CEL metrics, e.g. how much time an expression is evaluate. 

@haoqing0110 would you also give some suggestion on this?

enhancements/sig-architecture/138-workload-completion/README.md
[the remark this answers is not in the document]
bhperry: That's an interesting way to do it. The messageExpression could receive an additional boolean variable with the result of the main expression so you could do things like:
```yaml
condition: Complete
expression: <some complicated expression>
messageExpression: object.metadata.name + (result ? " is complete" : " is not complete")
```

bhperry:
> Some thing that we also need to consider is CEL metrics, e.g. how much time an expression is evaluate.

Agreed. It looks like existing kubernetes resources that use CEL typically bound both individual cost of a single expression, as well as cost of all expressions in the resource. The docs also mention [validating worst case cost](https://kubernetes.io/docs/reference/using-api/cel/#estimated-cost-limits) to ensure no expressions that would be prohibitively expensive are allowed on create/update, so that might be a good thing to add to the validating webhook for ManifestWork.

haoqing0110:
There is [CostLimit and CostTracking](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/compilation.go#L236 ) functions in cel.

enhancements/sig-architecture/138-workload-completion/README.md
haoqing0110: I don't think of a case that needs to compare the resource quantity and mark as complete. Maybe it's just a code level thing to include a specific library in the future when necessary.

bhperry: Probably not resource quantities, but could imagine a CRD using quantity in status and then comparing something like `status.quantity >= spec.quantity`. 

More generally, I do think making CEL in OCM match CEL in standard kubernetes APIs is a good idea for usability and future proofing.

haoqing0110: I'm not sure if a crd/data structure that does not support protobuf can be used directly as a variable in cel (ref [Use variables in a function](https://codelabs.developers.google.com/codelabs/cel-go#4)).
Also, maybe can point to https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiserver/pkg/cel/library for the libraries we use, semverlib.go is a new library not list in the doc.

enhancements/sig-architecture/138-workload-completion/README.md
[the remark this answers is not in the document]
qiujian16: maybe not, reason is more static. We could have some predefined reasons.

enhancements/sig-architecture/138-workload-completion/README.md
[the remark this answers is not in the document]
qiujian16: I think evaluate condition/apply can be in independent threads. When a resource is applied but completed condition is not set, user can use different update strategy to avoid update resource, e.g. createOnly, ignoreFields with server side apply.

enhancements/sig-architecture/138-workload-completion/README.md
bhperry: Manifests are always treated as `unstructured.Unstructured` right? So I think it will work, because object will be `map[string]interface{}` with a cel variable of type `cel.DynType`

bhperry: I was thinking of linking to https://kubernetes.io/docs/reference/using-api/cel/

qiujian16:
/approve

@haoqing0110 would you like to do a final review?

PR #143, opened 2025-05-22
Proposal for placement sort order and score visibility

bhperry:
Enhancement proposal for making PlacementDecision scores easily visible to consumers, and supporting different sort orders in the decisions.

### DIFF
--- a/enhancements/sig-architecture/225-placement-score-visibility/README.md
+++ b/enhancements/sig-architecture/225-placement-score-visibility/README.md
@@ -67,6 +67,8 @@ type PlacementSpec struct {
   // ScoreRateLimit sets maximum rate of updates to recorded scores in placement decisions.
   // Score changes that do not change the selected set of clusters in the placement will
   // not be reflected in decisions more often than the given duration.
+  // +kubebuilder:validation:Pattern="^[0-9]+[h|m|s]$"
+  // +kubebuilder:default:="1m"
   // +optional
   ScoreRateLimit string `json:"scoreRateLimit,omitempty"`
 }


enhancements/sig-architecture/225-placement-score-visibility/README.md
```diff
@@ -0,0 +1,173 @@
...
+`numberOfClusters: N` on the placement, this means a new placement must be created for every value
+of N. Another use-case would be to distribute workloads out to selected clusters randomly weighted
+by their scores. There is currently no way to do this in OCM without recomputing scores for each cluster
+listed in the PlacementDecisions. Displaying scores in an easy to consume manner gives more flexibility
+to the user, and enables a single Placement to be reused for many different scheduling decisions
+(reducing API pressure on the hub).
+
+## Proposal
+
+Add `SortBy` to `PlacementSpec` to configure the sort order for PlacementDecision. This field
```
qiujian16, line 40: one of the issue with sorting by score is that if the score changed to frequently, the decision result will be updated frequently also. We need to think about some approach to avoid that, or at least some document to notice users.

qiujian16, line 40: one of the issue with sorting by score is that if the score changed to frequently, the decision result will be updated frequently also. We need to think about some approach to avoid that, or at least some document to notice users.

enhancements/sig-architecture/225-placement-score-visibility/README.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+Display score to consumer of decisions
+
+```go
+type ClusterDecision struct {
+  ...
+
+  // Score is the computed score for the cluster based on configured prioritizers
+  Score int64 `json:"score"`
```
haoqing0110, line 82: should this type be aligned with addonplacementscore https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/types_addonplacementscore.go#L60 , int32.

enhancements/sig-architecture/225-placement-score-visibility/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+title: workoad-completion
```
haoqing0110, line 1: metadata needs to be updated

enhancements/sig-architecture/225-placement-score-visibility/README.md
bhperry: Including that note in the description of `SortBy: Score` perhaps?

What is the main concern with decision results updating frequently?

enhancements/sig-architecture/225-placement-score-visibility/README.md
bhperry: I used int64 because that is the type it is calculated in https://github.com/open-cluster-management-io/ocm/blob/4cbb12d5a2622ce5a698d5e07530f76166d3a922/pkg/placement/controllers/scheduling/schedule.go#L39-L40

Since final score is not bounded I think it makes sense to use int64

enhancements/sig-architecture/225-placement-score-visibility/README.md
```diff
@@ -0,0 +1,173 @@
...
+Field for sorting decisions
+
+```go
+type PlacementSpec struct {
+  ...
+
+  // SortBy sets the sort order for decisions.
+  // It can be "Name", "Score", or "", where "" is "Name" by default.
+  // If sortBy is "Name", decisions will be orderd alphanumerically by cluster name
+  // If sortBy is "Score", decisions will be ordered numerically in descending order by score,
```
haoqing0110, line 57: As mentioned above, need to notice user that when sortBy is "Score", the placement decision will trigger an update when score changes. Since today we didn't control the frequency of how user update the AddonPlacmentScore, that means if the score changed too frequently, the decision result will be updated frequently also.

haoqing0110: Defining the time (eg. every 1 minutes) to reschedule the placement might be a way to avoid the pressure on API server, but the above discussed, the update of the score is uncontrollable, so perhaps the best way for now is just add some notice to the user.

qiujian16: it might make sense to define a rateLimit, such as MaxUpdatePerMinute

enhancements/sig-architecture/225-placement-score-visibility/README.md
haoqing0110: Just revisited the logic of the placement scheduler to update the event. I think it needs to be refined; in some cased the event should not be recorded.

haoqing0110: @bhperry it's mainly used for recording and troubleshooting.

enhancements/sig-architecture/225-placement-score-visibility/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Summary
+
+This proposal extends existing Placement mechanisms to surface the scores of each cluster
+in PlacementDecision so that users (or future OCM features) can make advanced scheduling decisions
+based on the scored state of clusters.
+
+In addition, it adds support for different sort strategies on the decisions generated by a Placement.
+By default, decisions will continue to be sorted by "Name". An additional sort strategy "Score" will
+enable sorting decisions numerically in descending order by prioritizer score.
```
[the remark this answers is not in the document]
haoqing0110: FYI https://github.com/open-cluster-management-io/ocm/issues/1323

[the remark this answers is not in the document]
haoqing0110: Maybe add a default value in the API as a start, something like: 
```
   // ScoreRateLimit must be defined in [0-9h]|[0-9m]|[0-9s] format examples; 1h , 5m , 30s
  // +kubebuilder:validation:Pattern="^(([0-9])+[h|m|s])$"
  // +kubebuilder:default:="1m"
  // +optional
  ScoreRateLimit string `json:"scoreRateLimit,omitempty"`
```

PR #144, opened 2025-06-02
KEP: clusteradm operator

TylerGillson:
Proposal for wrapping `clusteradm` with a Kubernetes controller that reconciles a `MultiCluster` custom resource, providing declarative lifecycle management for OCM multi-clusters.

Note: the controller already exists and I am proposing that Spectro Cloud donates it to the OCM ecosystem. I spoke to @mikeshng about this briefly in Slack. My team and I wrote and are currently maintaining this controller, as we're using it within a product that we're developing at Spectro Cloud. I would be happy to continue maintaining it if/when it is adopted by OCM.

### DIFF
--- a/enhancements/sig-architecture/225-clusteradm-operator/README.md
+++ b/enhancements/sig-architecture/225-clusteradm-operator/README.md
@@ -10,7 +10,7 @@
 
 ## Summary
 
-This proposal introduces the clusteradm operator, which will provide a declarative interface for managing multi-cluster environments. It wraps the imperative `clusteradm` CLI in a Kubernetes-native way, introducing the `MultiCluster` CRD to enable automated bootstrapping of Kubernetes clusters into a multicluster topology. Ongoing maintenance of the multi-cluster in the form of configuration patches and OCM version upgrades is also supported.
+This proposal introduces the fleet config operator, which will provide a declarative interface for managing multi-cluster environments. It wraps the imperative `clusteradm` CLI in a Kubernetes-native way, introducing the `FleetConfig` CRD to enable automated bootstrapping of Kubernetes clusters into a multicluster topology. Ongoing maintenance of the multi-cluster in the form of configuration patches and OCM version upgrades is also supported.
 
 ## Motivation
 
@@ -31,7 +31,7 @@ OCM's current installation procedure is imperative-only, requiring multiple manu
 
 ## Proposal
 
-Add a new component to the OCM ecosystem: the clusteradm operator. It will reconcile a new OCM `MultiCluster` CRD, which encapsulates all information required to bootstrap and maintain an OCM multi-cluster. The spec of a `MultiCluster` will contain all the same information traditionally provided to `clusteradm init` and `clusteradm join`. The `MultiCluster` API surface will feature 1:1 parity with the `clusteradm` CLI, including all optional flags. In some cases, e.g., kubeconfigs, the spec will contain references to secrets containing auxiliary data to avoid exposing sensitive information.
+Add a new component to the OCM ecosystem: the fleet config operator. It will reconcile a new OCM `FleetConfig` CRD, which encapsulates all information required to bootstrap and maintain an OCM multi-cluster. The spec of a `FleetConfig` will contain all the same information traditionally provided to `clusteradm init` and `clusteradm join`. The `FleetConfig` API surface will feature 1:1 parity with the `clusteradm` CLI, including all optional flags. In some cases, e.g., kubeconfigs, the spec will contain references to secrets containing auxiliary data to avoid exposing sensitive information.
 
 ### User Stories
 
@@ -47,7 +47,7 @@ As an infrastructure or platform engineer, I want to commit the definition of my
 
 1. People may not want to store kubeconfigs for their spoke clusters on the hub cluster. Especially in a non-EKS scenario where the kubeconfigs contain long lived tokens. This can be mitigated by continuing to expand OCM's workload identify capabilities to support native integration with Azure and GCP in alignment with security best practices. Spoke kubeconfigs _could_ also be deleted from the hub cluster following the initial bootstrapping of the multi-cluster, but that introduces many challeges regarding upgrades, reconfiguration, etc.
 
-1. In certain environments, network connectivity from hub to spoke cannot be assumed. The clusteradm operator will **not** be able to bootstrap such spoke clusters automatically. However, additional metadata can be provided when these clusters are joined via `clusteradm join ... --klusterlet-annotation foo=bar`. The clusteradm operator on the hub can subsequently identify spoke clusters that were bootstrapped out-of-band and incorporate the health status of their corresponding `ManagedCluster` objects into the health of the overall `MultiCluster`.
+1. In certain environments, network connectivity from hub to spoke cannot be assumed. The fleet config operator will **not** be able to bootstrap such spoke clusters automatically. However, additional metadata can be provided when these clusters are joined via `clusteradm join ... --klusterlet-annotation foo=bar`. The fleet config operator on the hub can subsequently identify spoke clusters that were bootstrapped out-of-band and incorporate the health status of their corresponding `ManagedCluster` objects into the health of the overall `FleetConfig`.
 
 ## Design Details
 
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+
+## Design Details
+
+### API change
+
+No changes to the core OCM APIs are required.
+
+### MultiCluster API
+
+Current state of the v1alpha1 `MultiCluster` API, which is fully functional and meets all the capabilities outlined elsewhere in this enhancement proposal.
```
qiujian16, line 60: this is very similar as what we did with cluster-api https://github.com/open-cluster-management-io/ocm/tree/main/pkg/registration/hub/importer/providers. And in ACM, we have a way to config klusterlet https://github.com/stolostron/cluster-lifecycle-api/blob/main/klusterletconfig/v1alpha1/types.go on the hub.

enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+	"sort"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MultiClusterSpec defines the desired state of MultiCluster.
+type MultiClusterSpec struct {
+	Hub              Hub               `json:"hub"`
+	Spokes           []Spoke           `json:"spokes"`
```
qiujian16, line 76: what happens if a user wants to add or delete a managedcluster? Can they directly create/delete a managedcluster resource?

enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+  - Dedicated hub + spoke (EKS)
+- Patches/Upgrades
+  - Feature gates (not actually supported at this time, requires feature implementation in `clusteradm` first)
+  - Helm values (singleton)
+  - Upgrade OCM bundle version
+- Uninstallation
+
+### Graduation Criteria
+
+TBD
```
qiujian16, line 548: I think we also iterate on some alternatives. For example:

User can set a global or per cluster configs, so when the user create a managedcluster, the agent will be deployed to the target cluster automatically.

enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+	"fmt"
+	"sort"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MultiClusterSpec defines the desired state of MultiCluster.
+type MultiClusterSpec struct {
+	Hub              Hub               `json:"hub"`
```
qiujian16, line 75: is this necessary? I think user can just apply ClusterManager resource.

enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal introduces the clusteradm operator, which will provide a declarative interface for managing multi-cluster environments. It wraps the imperative `clusteradm` CLI in a Kubernetes-native way, introducing the `MultiCluster` CRD to enable automated bootstrapping of Kubernetes clusters into a multicluster topology. Ongoing maintenance of the multi-cluster in the form of configuration patches and OCM version upgrades is also supported.
+
+## Motivation
+
+OCM's current installation procedure is imperative-only, requiring multiple manual operations, either via `clusteradm` or direct application of OCM custom resources to various Kubernetes clusters. The lack of a declarative interface for managing the lifecycle of an OCM multi-cluster has several limitations:
```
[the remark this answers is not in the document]
mikeshng: Right. I think the proposed API and implementation outlined in this enhancement PR can still land in the clusteradm repo.

enhancements/sig-architecture/225-clusteradm-operator/README.md
TylerGillson: Managed clusters can be added/removed by modifying the `MultiCluster`. The controller will execute the appropriate `clusteradm` commands to achieve the desired result, e.g. `clusteradm join`, `clusteradm unjoin` (against the spoke), `clusteradm accept` (against the hub), etc.

enhancements/sig-architecture/225-clusteradm-operator/README.md
TylerGillson: That assumes that the OCM hub operators are already installed to reconcile the `ClusterManager`. How would the OCM hub components be initialized in the scenario you're describing?

By running `clusteradm init`, the OCM hub components and installed _and_ the `ClusterManager` is created.

enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+
+- Complexity around bootstrapping and maintenance
+- Multi-cluster topology / configuration cannot be managed via GitOps
+- Not aligned with Kubernetes best practices
+
+### Goals
+
+- Declarative lifecycle management of an OCM multi-cluster
+- Simplify and streamline the OCM onboarding and maintenance experience
```
mikeshng, line 26: Users will need to install this clusteradm operator and API as the first step. Would you mind adding a bit more detail about how that installation would work in the KEP? For example, will it be via a Helm chart, a new clusteradm subcommand, or something else?

enhancements/sig-architecture/225-clusteradm-operator/README.md
mikeshng: Qi is referring to the ACM pattern for importing an existing cluster. There's an [import controller](https://github.com/stolostron/managedcluster-import-controller) that watches for the creation of a ManagedCluster, along with its namespace and a way to connect to the spoke cluster (like a kubeconfig Secret). Once those are in place, it can connect to the managed cluster, deploy the klusterlet operator, and create the Klusterlet CR to initiate the registration process.

qiujian16: I mean you will still need to start "clusteradm operator" on the hub/spoke, and still need to handle the version and lifecycle of this operator. It is not clear to me why user cannot just install ocm operators directly. It would need some clarification.

TylerGillson: Everything you said is true. The value add that the proposed operator provides is around consolidation and minimizing manual steps. Users would have a single cluster with network connectivity to all other clusters that acts similarly to a CAPI management cluster. Say you have 20 clusters that you want to link together. With this approach you would only need to install the multi-cluster operator on one cluster and apply a single CR.

enhancements/sig-architecture/225-clusteradm-operator/README.md
```diff
@@ -0,0 +1,557 @@
...
+
+- [] Enhancement is `provisional`
+- [] Design details are appropriately documented from clear requirements
+- [] Test plan is defined
+- [] Graduation criteria for dev preview, tech preview, GA
+- [] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal introduces the clusteradm operator, which will provide a declarative interface for managing multi-cluster environments. It wraps the imperative `clusteradm` CLI in a Kubernetes-native way, introducing the `MultiCluster` CRD to enable automated bootstrapping of Kubernetes clusters into a multicluster topology. Ongoing maintenance of the multi-cluster in the form of configuration patches and OCM version upgrades is also supported.
```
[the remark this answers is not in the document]
TylerGillson: Happy to use whatever name you prefer!

enhancements/sig-architecture/225-clusteradm-operator/README.md
TylerGillson: The operator we've developed is currently installed via Helm. We could certainly extend `clusteradm` to support installing that Helm chart as well. Maybe, `clusteradm fleet-operator-init`? Any command name proposals?

mikeshng: Helm chart sounds good. We probably don't need the clusteradm sub-command for a while.

enhancements/sig-architecture/225-clusteradm-operator/README.md
qiujian16: the goal is very similar as what we did on capi-importer work https://github.com/open-cluster-management-io/ocm/tree/main/pkg/registration/hub/importer. The difference is it happens the join happens when a managedcluster is created on the hub. Would you also take a look at to see if the existing approach is good enough and can be extended to support your goal?

TylerGillson:
@mikeshng should we merge this? Or just close it?

mikeshng (approved):
/lgtm

Issue #146, opened 2025-07-14
KEP: Centralized Namespace Management Across ManagedClusterSets


jnpacker:
This enhancement introduces a global mechanism for defining and managing namespaces on Managed Clusters directly from the hub. The proposal creates a new CRD API for "Global Namespaces" that span all clusters within a ManagedClusterSet, enabling centralized application of Kubernetes access controls through ClusterPermissions. This addresses the current limitation where ManifestWork and ClusterPermission resources must be managed separately and independently, providing a more streamlined approach for multi-cluster governance and security consistency.
The key benefits include automatic and consistent namespace creation across all managed clusters in a set, simplified security management at scale, and improved support for workloads like Virtual Machines that require namespace consistency for migration and visibility. The enhancement is specifically designed to work with existing RBAC tooling while avoiding management of system namespaces. The proposal includes user stories for creating and deleting Global Namespaces, with considerations for handling existing namespaces and deletion strategies. This enhancement fills a gap in the current architecture by providing a unified approach to namespace management that's tightly integrated with ManagedClusterSet boundaries.

mikeshng:
Done in https://github.com/open-cluster-management-io/enhancements/pull/147

PR #147, opened 2025-07-14
KEP: Centralized Namespace Management Across ManagedClusterSets

jnpacker:
* Add support for Globally defined namespace(s) on the ManagedClusterSet
 * Support RBAC via ClusterPermissions for the Globally defined namespace(s)

### DIFF
--- a/enhancements/sig-architecture/146-fleet-namespaces/README.md
+++ b/enhancements/sig-architecture/146-fleet-namespaces/README.md
@@ -1,0 +1,182 @@
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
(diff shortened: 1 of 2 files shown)


jnpacker:
https://github.com/open-cluster-management-io/enhancements/issues/146

enhancements/sig-architecture/146-fleet-namespaces/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Proposal
+
+This is where we get down to the nitty gritty of what the proposal actually is.
+
+### User Stories
+
+#### Story 1
+Create a Global Namespace, this namespace will be created on all Managed Clusters in a Managed Cluster Set, and 
+the appropriate Cluster Permissions will be applied.
```
qiujian16, line 37: if we create a managedclusterbinding that binds a clusterset to a namespace, does it mean we should sync the namespace that creates managedclusterbinding to all clusters? It seems we can add a spec field in managedclusterbinding to achieve that.

PR #151, opened 2025-08-13
Merge cluster-proxy into core

qiujian16:


enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
```diff
@@ -0,0 +1,167 @@
...
+The connection between klusterlet and hub for proxy functionality will remain gRPC-based. This is independent
+of the gRPC registration driver, requiring a separate gRPC configuration file for the proxy. To obtain the gRPC
+configuration for proxy, the klusterlet agent will send a Certificate Signing Request (CSR) to the hub with the
+signer name `open-cluster-management.io/klusterlet-proxy`. The gRPC server on the hub will approve and sign the
+CSR. The gRPC configuration will be saved in the `hub-kubeconfig-secret` with the key `proxy-grpc.yaml`. The
+klusterlet will only start the proxy when it detects the existence of this configuration file. The proxy
+registration process begins after the cluster registration process and will not impact normal cluster
+registration.
+
+#### Proxy API
```
xuezhaojun, line 83: I wonder whether we should use term "proxy", I got same case users get confused about cluster-proxy with other "proxy" settings like addonproxyconfig.proxyconfig.

Could we use another term like "tunnel"?

enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
```diff
@@ -0,0 +1,167 @@
...
+`clusteradm` will be enhanced with a subcommand `clusteradm proxy --cluster <cluster name>`, which will configure
+the kubeconfig with the proxy endpoint in a cluster-specific context and switch the kubeconfig context to the target
+cluster. Users can then use kubectl to connect to the target cluster.
+
+#### Integration with ClusterProfile API
+
+#### Installation
+
+To enable the feature, the user needs to enable the feature gate in `ClusterManager` and `Klusterlet` API.
+In addition, A new proxyConfig field should be added in `ClusterManager` API:
```
skeeey, line 100: how about introduce a config for grpcserver instead of proxyConifg
```go
type GRPCServerConfiguration struct {
    ImagePullSpec string `json:"imagePullSpec,omitempty"`
    FeatureGates []FeatureGate `json:"featureGates,omitempty"`
    EndpointExposure *GRPCEndpointExposure `json:"endpointExposure,omitempty"` 
}
```

skeeey: yes, for registrationConfig, we need  to set  authType with grpc to enable the grpc

enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
xuezhaojun: I believe the complete description of this functionality can be "proxy requests between hub and agent," but it can also be "establish a tunnel between hub and agent that can be used to proxy requests."

For users of this functionality, they indeed only need to know about an endpoint that can proxy requests - they don't need to know about the existence of a "tunnel." However, for developers and maintainers, the term "tunnel" helps distinguish this functionality from other proxy-related features.

If "tunnel" doesn't work, could we adopt other terms such as: gateway, bridge, relay, etc.?

enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+`clusteradm` will be enhanced with a subcommand `clusteradm proxy --cluster <cluster name>`, which will configure
+the kubeconfig with the proxy endpoint in a cluster-specific context and switch the kubeconfig context to the target
+cluster. Users can then use kubectl to connect to the target cluster.
+
+#### Integration with ClusterProfile API
+
+#### Installation
+
+To enable the feature, the user needs to enable the feature gate in `ClusterManager` and `Klusterlet` API.
```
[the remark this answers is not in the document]
xuezhaojun: Don‘t we also need cert and endpoint configuration of this http server also set in cluster-manager?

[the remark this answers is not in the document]
xuezhaojun: What if user want to use their customzied cert as the proxy endpoint maybe exposed in their domain?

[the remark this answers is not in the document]
xuezhaojun: grpc connection is internal used by proxy, but http proxy is facing to user or maybe user's user.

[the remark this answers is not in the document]
qiujian16: I got your point, yes we need at least endpoint of http server. I am not sure we need ca bundle. Do we support setting this in cluster-proxy today?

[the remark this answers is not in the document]
qiujian16: updated

enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
qiujian16: updated

enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
qiujian16: relay seems ok to me, but since cluster-proxy has been known to many, I am not quite sure how strong the desire to rename. 

@jnpacker thought?

enhancements/sig-architecture/149-moving-cluster-proxy-to-core/README.md
```diff
@@ -0,0 +1,269 @@
...
+        potocol: HTTP
+        httpEndpointExposure:
+          endpoint: https://<external http server>
+      - type: agentToServer
+        protocol: GRPC
+        grpcEndpointExposure:
+          endpoint: grpc://<external grpc address>
+```
+
+Example of grpc registraion with proxy enabled:
```
skeeey, line 182:
```suggestion
Example of grpc registraion with proxy disabled:
```

skeeey:
> @skeeey @xuezhaojun are your fine with the API change on the clusterManager, if so we will need to update clustermanager API sooner.

I'm fine with it

xuezhaojun:
> @skeeey @xuezhaojun are your fine with the API change on the clusterManager, if so we will need to update clustermanager API sooner.

Yes, the PR looks good to me.

PR #161, opened 2025-10-28
Enhancement for feedback scrape type

annelaucg:


enhancements/sig-architecture/179-work-feedback-scrape-type/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+What are the caveats to the implementation? What are some important details that
+didn't come across above. Go in to as much detail as necessary here. This might
+be a good place to talk about core concepts and how they relate.
+
+### Risks and Mitigation
+
+What are the risks of this proposal and how do we mitigate. Think broadly. For
```
qiujian16: I think one of the risks we have to consider is there might be too many watch event, and there could be too many updates call to the manifestwork.

enhancements/sig-architecture/179-work-feedback-scrape-type/README.md
```diff
@@ -0,0 +1,150 @@
...
+`spec.manifestWorkTemplate.manifestConfigs`. 
+
+#### Spoke Level ManifestWork Controller Reconciler 
+
+Default: resources continue to be polled on the existing interval.
+If a resource is marked WATCH, the controller registers it for watch-based updates:
+Immediate status/feedback recomputation on events.
+Poll remains as a safety net.
+
+Create a FeedbackScrape interface that polls or watches resources.
```
qiujian16: will this only impact resource status feedback, or it also impacts resource apply logic?

annelaucg: Which apply logic?

qiujian16: oh I mean this controller https://github.com/open-cluster-management-io/ocm/blob/main/pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go

enhancements/sig-architecture/179-work-feedback-scrape-type/README.md
annelaucg: Added some more details to this section on how we can limit the updates to the ManifestWork object. 

Need to still think through a few more implementation details (eg. how can MWRS verify if the resource is Ready to stop the informer from watching).

enhancements/sig-architecture/179-work-feedback-scrape-type/README.md
```diff
@@ -0,0 +1,150 @@
...
+If a resource is marked WATCH, the controller registers it for watch-based updates:
+Immediate status/feedback recomputation on events.
+Poll remains as a safety net.
+
+Create a FeedbackScrape interface that polls or watches resources. 
+
+In the ManifestWork Controller (pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go)
+introduce a watch-based path alongside the existing poll loop. When syncing the ManifestWork, register a informer for the resource if `feedbackScrapeType`  WATCH. 
+
+When there is a change seen on the WATCH type, patch the status conditions for that resource.
```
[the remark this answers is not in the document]
qiujian16: I do not think there is, I can build some demo code on that. I think dynamic informer per resource/name/namespace might be too many. I am thinking dynamic informer per  resource/namespace is good enough. Just something need to be considered when we implement it.

enhancements/sig-architecture/179-work-feedback-scrape-type/README.md
```diff
@@ -0,0 +1,188 @@
...
+If a resource is marked WATCH, the controller registers it for watch-based updates:
+Immediate status/feedback recomputation on events.
+Poll remains as a safety net.
+
+Create a FeedbackScrape interface that polls or watches resources. 
+
+In the ManifestWork Controller (pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go)
+introduce a watch-based path alongside the existing poll loop. 
+
+When syncing the ManifestWork, register a informer for the resource if `feedbackScrapeType`  WATCH. If the `feedbackScrapeType` is no longer WATCH, unregister the informer. If the ManifestWork is in Ready state (Progressing == False and Degraded == False / NotSet), unregister the informer. This will prevent long running WATCH from triggering too many ManifestWork changes. If there is new rollout, and the status changes to Progressing, then register the informer again.
```
zhujian7: Could you explain more details about the register/unregister informer part? I have a concern that in some cases, even with this watch informer, we might still not observe the whole status change process of the resource, since it may take several seconds to wait for the informer cache to sync when starting the informer. For example, if the replica number changes from 1->2->3 in 5 seconds, and the waiting for the deployment informer cache sync process takes 6 seconds, the result we see will still be 1->3.

zhujian7: I am thinking of the situation every time the register starts up. Let us first leave the manifestworkreplicaset alone and focus on the manifestwork. And let us say there is a manifestwork:
```
apiVersion: work.open-cluster-management.io/v1
kind: ManifestWork
metadata:
  namespace: <target managed cluster>
  name: hello-work-demo
spec:
  manifestConfigs:
    - resourceIdentifier:
        group: apps
        resource: deployments
        namespace: default
        name: hello
      feedbackScrapeType: WATCH
  workload:
    manifests:
      - apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: hello
          namespace: default
        spec:
          replicas: 1
          selector:
            matchLabels:
              app: hello
          template:
            metadata:
              labels:
                app: hello
            spec:
              containers:
                - name: hello
                  image: quay.io/asmacdo/busybox
                  command:
                    ["sh", "-c", 'echo "Hello, Kubernetes!" && sleep 3600']
```

After this is applied on the hub,  will we register a deployment informer for this? And after the manifestwork becomes `Applied` true, `Available` true, will we unregister the informer?

And then if we change the replica of the deployment in the manifestwork from 1 to 3, will we re-register a deployment informer?

qiujian16: >  If the ManifestWork is in Ready state (Progressing == False and Degraded == False / NotSet), unregister the informer.

I think this should be updated.

zhujian7:
@annelaucg please signoff on every commit, you have 6 commits in the PRs, you can first use `git rebase --interactive HEAD~6`, then `squash` them to one commit; after that `git commit --amend --signoff` and `git push -f`

PR #164, opened 2025-11-12
Support rollback for ManifestWorkReplicaSet

youngbupark:
* Support rollback for ManifestWorkReplicaSet

### DIFF
--- a/enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
+++ b/enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
@@ -1,0 +1,469 @@
+# Support Rollback for ManifestWorkReplicaSet
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This enhancement proposes adding rollback capabilities to [ManifestWorkReplicaSet](https://open-cluster-management.io/docs/concepts/work-distribution/manifestworkreplicaset/) (MWRS) to enable safe recovery from failed multi-cluster deployments. The solution leverages Kubernetes ControllerRevision resources to maintain a historical record of ManifestWork template changes in MWRS, similar to how StatefulSet and DaemonSet controllers track revision history. 
+
+The enhancement introduces two key rollback mechanisms: `abort` (via cancelling rollout) and `undo` (via spec updates). When a progressive rollout fails, the controller can automatically revert to the previous stable revision without updating MWRS spec across all updated clusters, minimizing downtime and impact. For `undo` scenarios, users can update the MWRS spec by referencing a previous revision history (e.g. using CLI), and the controller will restore that revision with a regular rollout strategy. `undo` may need additional tool to simplify the manual rollback process, such as a CLI similar to [kubectl rollout undo](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_rollout/kubectl_rollout_undo/).
+
+New API fields include `revisionHistoryLimit` to control history retention, `failureStrategy` to enable automatic abort, and status fields (`currentRevision`, `updateRevision`, `abortedTime`, `collisionCount`) to track rollout state. The enhancement also introduces a new `Progressing` condition type and `RolloutDegraded` reason to provide better visibility into rollout and rollback operations.
+
+## Motivation
+
+The ManifestWorkReplicaSet (MWRS) simplifies multi-cluster workload distribution by supporting several rollout strategies that enable gradual deployment of workloads across clusters. This allows operators to safely deploy applications to multiple managed clusters with controlled rollout patterns (using rolloutStrategy).
+
+However, the current implementation has limitation: when a progressive rollout fails, the Work controller simply stops the rollout, leaving clusters in an inconsistent state. Some clusters may have the new version deployed while others remain on the old version, and the failed clusters are stuck in a failed state. This creates several problems:
+
+1. **Inconsistent Cluster State**: Partial rollouts leave the system in an undefined state where different clusters run different versions
+2. **Extended Downtime**: Critical applications may experience extended downtime while waiting for manual intervention
+
+For critical applications, especially those requiring high availability, the ability to automatically roll back to a previous stable revision on failure is essential to minimize impact and maintain service continuity. This enhancement addresses these limitations by providing both automatic and manual rollback capabilities with proper revision history management.
+
+### Goals
+
+- Define ManifestWork template revision management.
+- Enable abort/undo rollback features with the revision history.
+- Enable automatic abort on the failure of rollout.
+
+### Non-Goals
+
+- Support rollback command in clusteradm CLI or kubectl plugin.
+- Support manual abortion.
+- Rollback across different API versions of ManifestWorkReplicaSet.
+- Use the same rolloutStrategy to roll back to the previous revision.
+
+## Proposal
+
+### User Stories
+
+#### Story 1 — Automatic rollback on failure
+
+As a user, I want MWRS to automatically abort workloads in the event of a failed rollout. When a deployment fails, MWRS should identify the previous revision and restore it consistently to the already rolled out clusters. The abort should be able to perform any additional operations needed to safely revert to the stable version.
+
+#### Story 2 — Manual rollback capability
+
+As a user, I want to manually roll back to a previous stable revision when needed. I should be able to update the ManifestWorkReplicaSet spec to reference a previous revision, and the controller should restore that revision across all clusters.
+
+### Implementation Details/Notes/Constraints
+
+The rollback mechanism leverages Kubernetes `ControllerRevision` resources to maintain a historical record of `ManifestWorkReplicaSet` changes. This approach is similar to how StatefulSet and DaemonSet controllers track revision history.
+
+Key implementation constraints:
+- Revision history is limited by `revisionHistoryLimit` to prevent unbounded storage growth
+- Automatic abort (`failureStrategy`) is opt-in and should be used with caution when CRDs are involved
+- Rollback operations require the previous revision to be available in ControllerRevision history
+- The controller uses hash-based collision detection to ensure unique revision naming
+
+### Risks and Mitigation
+
+1. **CRD Rollback** - If the manifestwork template includes a CRD version update, rolling back the CRD resource can cause significant impact on the system and application. 
+   - **Mitigation**: Using the automatic abort feature (`failureStrategy`) should be opt-in and used with caution. Users should be aware of the implications when rolling back CRD changes.
+
+2. **Storage Growth** - Maintaining revision history could lead to unbounded storage growth.
+   - **Mitigation**: The `revisionHistoryLimit` field allows users to control the maximum number of revisions retained. The controller will automatically prune old revisions.
+
+3. **Automatic abort failure** - If a abort operation itself fails, the system could be left in an inconsistent state.
+   - **Mitigation**: The controller tracks rollback status through conditions and status fields, allowing for monitoring and manual intervention if needed.
+
+## Design Details
+
+This design describes the changes to ManifestWorkReplicaSet and ManifestWork API objects and proposes enhancements to the current reconciler logic to support revision history with rollback.
+
+To enable rollback features, we propose two different modes:
+* The `abort` operation cancels the progressing rollout and rolls it back to the stable revision without updating `.spec.manifestWorkTemplate` to the older revision.
+* The `undo` operation requires explicit manual action by updating `.spec.manifestWorkTemplate` to an older revision.
+
+### ManifestWorkReplicaSet API Object
+
+To support abort and rollback feature, we will have the following API changes:
+
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWorkReplicaSet
+metadata:
+  name: <MWRS_Name>
+spec: 
+  # the number of revision history limit
+  revisionHistoryLimit: 3
+  placementRefs:
+    - name: placement-rollout-progressive
+      rolloutStrategy:
+        type: Progressive
+        progressive:
+          ...
+        # (NEW) failureStrategy: if AbortAll, automatically aborts on failure.
+        failureStrategy: AbortAll | nil
+  manifestWorkTemplate:
+    manifestConfigs:
+    - feedbackRules:
+        - jsonPaths:
+          - name: isHealthy
+            path: '.status.conditions[?(@.type=="Healthy")].status'
+          - name: updatedReplicas
+            path: '.status.updatedReplicas'
+          - name: replicas
+            path: '.status.replicas'
+          - name: replicas
+            path: '.status.updatePodHash'
+          - name: replicas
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+  # the number of revision history limit
+  revisionHistoryLimit: 3
+  placementRefs:
+    - name: placement-rollout-progressive
+      rolloutStrategy:
+        type: Progressive
+        progressive:
+          ...
+        # (NEW) abortOnFailure: if true, automatically aborts and rolls back on failure; otherwise, rollout ends on failure
+        abortOnFailure: true
```
haoqing0110: Will abort follow the same rolloutStrategy? Like auto rollback all at once, Progressive.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+    work.open-cluster-management.io/placementname: placement1
+spec: 
+  ...
+```
+
+### `Abort` vs `Rollback`
+
+`Abort` is used to **cancel** the current rollout. When aborting the current rollout, spec is not changed, but `status.abort` is set to `true`, which will set `status.abortedTime`. then loading the older revision and then apply `.spec.manifestWorkTemplate` of the old revision to all clusters at once.
+
+On the other hand, `rollback` is the explicit action. If user wants to roll back to the older revision, the user (or cli) can find the revision from the list of ControllerRevision resources and apply the older manifestTemplate to update the `.spec.manifestWorkTemplate`.
```
haoqing0110: Does this mean the user can not use Revision to manual rollout? but need to copy and update the `.spec.manifestWorkTemplate`.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+### Goals
+
+- Define manifestwork template revision management.
+- Enable rollback feature with the revision.
+- Enable automatic rollout on the failure of rollout.
+
+### Non-Goals
+
+- Manual rollback via kubectl plugin commands (rollback will be performed by updating `.spec.manifestWorkTemplate`)
+- Rollback of CRD versions (automatic rollback should be used with caution when CRDs are involved)
```
annelaucg: Does this mean CRDs deployed by MWRS will not be included in the rollback?

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+- Rollback operations require the previous revision to be available in ControllerRevision history
+- The controller uses hash-based collision detection to ensure unique revision naming
+
+### Risks and Mitigation
+
+1. **CRD Rollback Risk** - If the manifestwork template includes a CRD version update, rolling back the CRD resource can cause significant impact on the system and application. 
+   - **Mitigation**: Using the automatic rollback feature (`abortOnFailure`) should be opt-in and used with caution. Users should be aware of the implications when rolling back CRD changes.
+
+2. **Storage Growth Risk** - Maintaining revision history could lead to unbounded storage growth.
+   - **Mitigation**: The `revisionHistoryLimit` field allows users to control the maximum number of revisions retained. The controller will automatically prune old revisions.
```
annelaucg: Is there any maximum that the user can define?

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+    message: "New revision mwrs-XXXXXXX is successfully progressed"
+    reason: NewManifestWorkAvailable
+    status: "True"
+    type: Progressing
+
+  placementSummary:
+  - availableDecisionGroups: 1 (2 / 2 clusters applied)
+    name: placement1
+    summary:
+      # (NEW) updated: count of clusters with the update revision applied
```
annelaucg: Do we want to add the new `desiredTotal` field here too?

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+    # existing labels.
+    work.open-cluster-management.io/manifestworkreplicaset: default.mwrset-1
+    work.open-cluster-management.io/placementname: placement1
+spec: 
+  ...
+```
+
+### `Abort` vs `Rollback`
+
+`Abort` is used to **cancel** the current rollout. When aborting the current rollout, spec is not changed, but `status.abort` is set to `true`, which will set `status.abortedTime`. then loading the older revision and then apply `.spec.manifestWorkTemplate` of the old revision to all clusters at once.
```
annelaucg: nit: MWRS.spec is not changed

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
youngbupark: It will support any resources, however, we will need to add the caution in the document in the future. I would delete this part.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+```yaml
+apiVersion: work.open-cluster-management.io/v1
+kind: ManifestWork
+metadata:
+  name: <MWRS_Name>
+  namespace: <cluster name>
+  labels:
+    # (NEW) controller-hash is to distinguish between old and new
+    # Manifestworks during Manifestwork template updates.
+    work.open-cluster-management.io/controller-hash: <hash>
```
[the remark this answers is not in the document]
youngbupark: observedGeneration is monotonically increasing, which means whenever you update the spec, generation will increase. We cannot use this as a revision.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,425 @@
...
+### Enhanced Rollout with revision history
+
+1. `(NEW)` Reconstruct ManifestWorkReplicaSet history:
+    1. List existing `ControllerRevision`s controlled by this `ManifestWorkReplicaSet`.
+    1. Find the history of the current `ManifestWorkReplicaSet`'s `.spec.ManifestWorkTemplate` desired state, and create one if not found:
+    1. The `.name` of the history will be unique, generated from `.spec.ManifestWorkTemplate` hash with hash collision resolution (`.status.collisionCount`) 
+    1. If history creation failed:
+        1. If it is because of name collision:
+            1. If the collided history is same as `ManifestWorkReplicaSet`'s `.spec.ManifestWorkTemplate` desired state, there is the already created the history
+            1. Otherwise, bump ManifestWorkReplicaSet `.status.collisionCount` by 1,
```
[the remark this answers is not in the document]
youngbupark: Yes right.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
youngbupark: In the first iteration, we will auto abort to all clusters for the simplicity.  Please let me know if you think we still need to follow rolloutStrategy for abort.

qiujian16: I would define it as failureStrategy, with abortAll or rollbackAll as the strategy name. Try to avoid boolean type.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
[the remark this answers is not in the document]
qiujian16: please add some explanation on how the hash is computed. Will it only include spec? or also labels/annotations?

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,434 @@
...
+            namespace: default
+status:
+  # (NEW) collisionCount: hash collision count for revision naming
+  collisionCount: 0
+  # (NEW) currentRevision: the current revision hash
+  currentRevision: YYYYYYY
+  # (NEW) updateRevision: the desired update revision hash
+  updateRevision: XXXXXXX
+  # (NEW) abort: true if abort operation has started
+  abort: true | false
```
qiujian16: this seems to overlap with abortedTime. I think we can just use abortedTime

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,434 @@
...
+| --- | --- | --- | --- |
+| AsExpected (True) | | NewRevisionCreated (True) | |
+| AsExpected (True)| Progressing (False) | NewRevisionCreated (True) | Progressing (False) | 
+| AsExpected (True)| Progressing (False) | NewRevisionCreated (True) | Progressing (False) | 
+| AsExpected (True)| NotAsExpected (False) | ProgressDeadlineExceeded (False) | RolloutDegraded (False) |
+| AsExpected (False)| NotAsExpected (False) | RolloutAborted (False) | RolloutDegraded (False) |
+
+### Open Questions
+
+1. Should we have the delay before starting aborting a rollout ?
```
qiujian16: sounds reasonable, but we could consider this as an additional abort strategy in beta phase.

qiujian16 (commented):
thanks, I'd like to have two examples on how mwrs status looks like during rollback:
1. upgrade and abort, how will the status changes (conditions/reasons/summary etc) among each state. And how user can know whether a rollback is done in which revision?
2. manual rollback, how user set rollback and check if rollback is finished.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,436 @@
...
+2. `(NEW)` Update `.spec.updateRevision` to the newly calculated revision.
+3. Iterate each placement references (`.spec.placementRefs`)
+    1. Get all manifestworks associated with the current placement
+    1. `(NEW)` Find the progressing manifestwork hash
+        1. Use `.spec.updateRevision` if `.status.abort` is true, otherwise, use `.status.currentRevision`.
+    1. Find manifestworks which has the same hash of the progressing manifestwork hash.
+        1. [Determine cluster rollout status (Failed, Success)](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go#L96)
+    1. Create Rollout handler
+    1. Calculate RolloutResult (rollout/timeout/removed candidate clusters)
+    1. `(NEW)` If rolloutResult includes timeout clusters and `.spec.placementRefs[*].rolloutStrategy.abortOnFailure` is true
```
haoqing0110: What's the condition to trigger automatic abort? How about the failed clusters?

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -215,9 +214,9 @@ spec:
 
 ### `Abort` vs `Rollback`
 
-`Abort` is used to **cancel** the current rollout. When aborting the current rollout, spec is not changed, but `status.abort` is set to `true`, which will set `status.abortedTime`. then loading the older revision and then apply `.spec.manifestWorkTemplate` of the old revision to all clusters at once.
+`Abort` is used to **cancel** the current rollout. When aborting the current rollout, spec is not changed, but `status.abort` is set to `true`, which will set `status.abortedTime`. Then, it looks up ControllerRevision which has `.status.currentRevision` (older revision) and then apply `.spec.manifestWorkTemplate` to ManifestWork resources of all cluster namespaces at once.
```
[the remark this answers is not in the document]
youngbupark: The reason why I define two fields is because we may want to allow users to manual abort feature in the future via CLI similar to argo rollout. So if user wants to abort (stop and revert) the current rollout, they can just set abort to true. But I agree that we can introduce this field later. I will use .status.abortedTime only at this time.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
youngbupark: When minSuccessTime is exceeded for the current rollout group, it will start the abort process.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
youngbupark: updated.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
youngbupark: @haoqing0110 I added new step `Evaluate abort condition` Please review it to see if it makes sense.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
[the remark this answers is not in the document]
youngbupark: I have added the section to explain the detail of hash compute.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,469 @@
...
+    work.open-cluster-management.io/controller-hash: <hash>
+
+    # existing labels.
+    work.open-cluster-management.io/manifestworkreplicaset: default.mwrset-1
+    work.open-cluster-management.io/placementname: placement1
+spec: 
+  ...
+```
+
+#### Calculate the hash
```
youngbupark: @qiujian16 I added the hash calculation details. PTAL.

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,469 @@
...
+
+### Status transition
+
+This section describes the detail status transition for each rollout condition.
+
+#### New manifestwork update transition
+
+| Transition order | PlacementVerified (Status) | ManifestworkApplied (Status) | Progressing (Status) | PlacementRollOut (Status) | status.currentRevision | status.updateRevision | status.abortedTime |
+| --- | --- | --- | --- | --- | --- | --- | --- |
+| 1 | AsExpected (True) | | NewRevisionCreated (True) | |  | 6b6dcbd695 | |
```
qiujian16: could you explain the meaning of each row and column? e.g. ManifestworkApplied is a condition type, but NewRevisionCreated is a reason or another condition type?

enhancements/sig-architecture/229-manifestworkreplicaset-rollback/README.md
```diff
@@ -0,0 +1,469 @@
...
+| 2 | AsExpected (True)| Progressing (False) | NewRevisionCreated (True) | Progressing (False) |  | 6b6dcbd695 | |
+| 3 | AsExpected (True)| Progressing (False) | NewRevisionCreated (True) | Progressing (False) |  |  6b6dcbd695 | |
+| 4 | AsExpected (True)| AsExpected (True) | NewRevisionAvailable (True) | Complete (True) | 6b6dcbd695 | 6b6dcbd695 | |
```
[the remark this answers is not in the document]
qiujian16: yeah, I think it would finally be false meaning no further step is running.

Issue #182, opened 2026-06-26
Feature request: Support Hosted Mode in AddOnTemplate


kahirokunn:
## Describe the enhancement

Please provide Hosted Mode support for the `AddOnTemplate` API.

Some addons need to run their agent from a hosting cluster while still accessing the managed cluster with least-privilege credentials. However, `AddOnTemplate` currently does not appear to provide a way to express this Hosted Mode deployment pattern.

## Expected behavior

* `AddOnTemplate` can declare or support Hosted Mode for an addon agent.
* The addon agent can be installed into a hosting cluster.
* The hosted agent can access the managed cluster using a provisioned, least-privilege kubeconfig or equivalent credential.

## Additional context

This request comes from the discussion here:

https://github.com/open-cluster-management-io/managed-serviceaccount/pull/325#discussion_r3467926162

mikeshng:
As discussed, we might not be ready to do this. @kahirokunn maybe we can close this for now.

PR #183, opened 2026-06-26
Add hosted mode support for addon template (#182)

mikeshng:
Add hosted mode support for addon template enhancement proposal

For https://github.com/open-cluster-management-io/enhancements/issues/182

## Summary by CodeRabbit

* **Documentation**
  * Added a new design proposal for enabling Hosted Mode in addon templates.
  * Documented how install mode-aware behavior, manifest placement, and hosted/managed cluster deployment flow should work.
  * Included expected status handling, cleanup behavior, examples, and rollout guidance.
  * Added enhancement metadata with ownership, status, and related references.

### DIFF
--- a/enhancements/sig-architecture/182-addontemplate-hosted-mode/README.md
+++ b/enhancements/sig-architecture/182-addontemplate-hosted-mode/README.md
@@ -1,0 +1,361 @@
+# Hosted Mode Support for AddOnTemplate
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [website](https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/)
+
+## Summary
+
+This proposal adds Hosted Mode support to `AddOnTemplate`-based addons. Today, addons built with the
+addon-framework Go library can deploy their agents on a hosting cluster (outside the managed cluster) using the
+Hosted Mode mechanism introduced in [enhancement #63](../63-hosted-addon/README.md). Template-based addons cannot
+do this because the template agent controller does not enable Hosted Mode and does not expose the install mode to
+template rendering. This proposal makes template-based addons work with Hosted Mode the same way code-based addons
+already do.
+
+## Motivation
+
+The [AddOnTemplate enhancement (#82)](../82-addon-template/README.md) explicitly listed Hosted Mode as a non-goal
+at the time of its initial design. Since then, Hosted Mode has become widely used in OCM. It allows addon agents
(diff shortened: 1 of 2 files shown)


enhancements/sig-architecture/182-addontemplate-hosted-mode/README.md
```diff
@@ -0,0 +1,361 @@
...
+on the hosting cluster. If the addon agent needs to access the managed cluster, the addon deployer must provide a
+`{addon-name}-managed-kubeconfig` secret in the install namespace on the hosting cluster, following the same
+pattern described in [enhancement #63](../63-hosted-addon/README.md).
+
+### Example
+
+#### AddOnTemplate with Hosted Mode Annotations
+
+```yaml
+apiVersion: addon.open-cluster-management.io/v1beta1
```
mikeshng, line 204: This enhancement assumes AddOnTemplate is already promoted to v1beta1. That work should be quite trivial compare to this enhancement.

kahirokunn:
@mikeshng Thank you for the great proposal!

After reviewing it, I think there are still some challenges when trying to express hosted addons like managed-serviceaccount or cluster-proxy using AddOnTemplate.

First, it's unresolved how to provide the kubeconfig that the agent running on the hosting cluster uses to access the managed cluster. Currently, managed-serviceaccount has its own provisioner on the addon side and generates/updates a managed kubeconfig using a least-privilege ServiceAccount token, but this seems less like something specific to managed-serviceaccount and more like a challenge common to hosted addons in general.

For this reason, how about giving this responsibility to AddOnTemplate or the addon-framework side? For example, a mechanism that provisions, for the agent on the hosting cluster, a kubeconfig for accessing the managed cluster with least privilege. If it could handle generating the kubeconfig using a ServiceAccount token on the managed cluster side, placing it into a Secret on the hosting cluster side, refresh, and cleanup, then each addon wouldn't need to have its own provisioner.

On top of that, how to reflect the prepared managed kubeconfig into the AddOnTemplate manifests is also a challenge. The proposal adds `INSTALL_MODE`, but since AddOnTemplate is basically simple value substitution, it's difficult to selectively output parts of the YAML or entire resources depending on the install mode. For example, adding kubeconfig volume/args/RBAC only in the Hosted case, creating additional ServiceAccount/Role/Deployment, or not creating hosted-only resources in the Default case. Since `hosted-manifest-location` determines the placement target in the Hosted case, handling both Default/Hosted with a single AddOnTemplate would also seem to require this kind of install-mode-dependent manifest selection.

With these two things in place, hosted addons like managed-serviceaccount and cluster-proxy could likely be expressed with AddOnTemplate as well.

mikeshng:
@kahirokunn what you said is all true. It's a huge amount of work to get it fully functioning.

kahirokunn:
Exactly... I think it will take a significant amount of time for both design and implementation to get it fully functioning. 😂

mikeshng:
This is much harder than initially discussed. Closing for now.

